PR #310, opened 2026-06-09
Fix STING Export Centre PDF pipeline (failure, watermark, combine UX, naming)

beckykyomugisha:
## Summary

End-to-end review and fix of the **STING Export Centre** PDF pipeline. Four defects were found from a real export session where all sheets reported as failed despite PDFs being written, and the configured watermark never appeared.

### 1. Every PDF reported "PDF export returned false" (critical)
`ExportSinglePdf` set `PDFExportOptions.Combine = false`. In Revit 2022+ the `FileName` property is **ignored** when `Combine = false` — Revit names files with its own rule (`Sheet-<name>.pdf`). The success check `File.Exists(<custom stem>.pdf)` therefore always failed, so valid exports were reported as failures *and* came out with the wrong filenames (`Sheet-SITE PLAN….pdf` instead of `…-A002-S2-P01.pdf`).
**Fix:** export each sheet with `Combine = true` (a single-view combine is just a 1-page PDF), so the requested ISO 19650 filename is produced exactly.

### 2. Diagonal watermark / bookmarks never appeared in OnePerSheet mode
Watermark and bookmark injection only ran in the combined-PDF path. In per-sheet mode it was never invoked, so a configured `DRAFT` watermark was silently dropped.
**Fix:** per-sheet PDFs now get the same watermark/bookmark post-processing as combined sets.

### 3. "Combine all sheets into one PDF" was undiscoverable
The Output-mode dropdown listed raw enum names (`OnePerSheet`, `OnePerSet`, …). `OnePerSet` is the combine-everything option but the label said nothing.
**Fix:** descriptive labels — `OnePerSet` now reads **"Single combined PDF (all selected sheets)"**.

### 4. Combined-PDF filenames embedded the first sheet's identity + `_All`
Combined files were named `…-A002-S2-P01_All.pdf`, borrowing an arbitrary sheet's number/level.
**Fix:** combined PDFs now carry set-level ISO identity via a `ResolveNaming` token-override overload — sheet-number slot becomes a set label (`ALL` / group name), level becomes `XX`, discipline/role reflect the group. A single combined set reads `…-ZZ-XX-DR-Z-ALL-S2-P01.pdf`.

## Files changed
- `StingTools/Docs/ExportCenterEngine.cs` — single-PDF `Combine` fix, per-sheet post-processing, `ResolveNaming` override overload, set-level combined naming.
- `StingTools/UI/StingExportCenterDialog.cs` — descriptive combine-mode labels.

## Testing / caveat
Per repo convention, this was written against documented Revit 2025+ API signatures but **not** compiled (`dotnet build`) — no .NET/Revit toolchain in the Linux sandbox. Verify in Revit before merge:
1. OnePerSheet export with watermark on → correct filenames + diagonal `DRAFT` on every page.
2. "Single combined PDF (all selected sheets)" → one merged PDF named with set-level identity.
3. Per-discipline / custom-group combined exports → clean distinct filenames.

The `Combine = true` per-sheet pattern is the standard Revit 2022+ workaround for custom PDF filenames.

https://claude.ai/code/session_01YVjBERxRYLDAJpweR9zk5a

---
_Generated by [Claude Code](https://claude.ai/code/session_01YVjBERxRYLDAJpweR9zk5a)_

### DIFF
--- a/StingTools/Core/StingToolsApp.cs
+++ b/StingTools/Core/StingToolsApp.cs
@@ -2198,6 +2198,7 @@ namespace StingTools.Core
                 ("Tag3D",                "3D Tag",        "T3", DrawingColor.Crimson,      typeof(HubTag3DCommand).FullName),
                 ("CreateTagFamilies",    "Tag Families",  "TF", DrawingColor.DarkCyan,     typeof(HubCreateTagFamiliesCommand).FullName),
                 ("AutoTag",              "Auto Tag",      "AT", DrawingColor.DarkGreen,    typeof(HubAutoTagCommand).FullName),
+                ("ExportCenter",         "Export Center", "EX", DrawingColor.RoyalBlue,    typeof(HubExportCenterCommand).FullName),
             };
 
             var buttons = new List<PushButtonData>(12);
@@ -2611,6 +2612,14 @@ namespace StingTools.Core
     {
         public Result Execute(ExternalCommandData data, ref string message, ElementSet elements)
             => HubDispatcher.Run(data, "AutoTag", ref message);
+    }
+
+    [Transaction(TransactionMode.ReadOnly)]
+    [Regeneration(RegenerationOption.Manual)]
+    public class HubExportCenterCommand : IExternalCommand
+    {
+        public Result Execute(ExternalCommandData data, ref string message, ElementSet elements)
+            => HubDispatcher.Run(data, "ExportCenter", ref message);
     }
 
     [Transaction(TransactionMode.ReadOnly)]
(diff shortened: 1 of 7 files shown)


StingD85:
Superseded by #318.

Same situation as #313→#316: main already has an Export Center **more advanced** than this branch (DwgMerger/NwcExporter/OdaConverter/XmlWriter + ISO 19650 naming + R2025 fixes), and both evolved the shared `ExportCenter*` files from the #305 base. The PDF-pipeline items in this PR's title — **watermark on all PDFs, combine UX, ISO naming/no-_All** — main already does (`InjectWatermark` stamps every page; `OutputSettings.NamingTemplate` defaults to the full ISO 19650-2 template).

#318 salvages only the genuinely-missing parts, re-implemented against main's current engine:
- **ScheduledExportRunner** (main had the `ScheduledExport` model but no execution layer)
- **"Changed Since Last Export" delta automation** (absent on main)

Build clean (0 errors), purely additive. Closing in favour of #318 — the branch can be deleted on merge.

PR #313, opened 2026-06-16
Fix tag-family CSV match rate, path resolution, and _BOOL parameter type conflict

beckykyomugisha:
## Summary

Three bugs fixed in the Create Tag Families / Load Shared Params pipeline:

- **Fix TagConfigCsvReader banner-drop bug** (`38cf679`) — the `⚠ WARNING PARAMETERS` sentinel handler reset `currentFamily`/`currentPlan` to null _without_ first committing the parsed tier rows. Because virtually every family block in the ARCH/GEN/MEP/STR CSVs ends with this banner, essentially all T4–T10 plans were silently discarded. The per-category CSV match rate went from ~8% to ~100% after the fix.

- **Fix CSV path resolution** (`fa9d955`) — `TagConfigPlanResolver.ResolveCsvPath()` now tries 7 additional fallback locations (relative to the assembly directory) so the tag-config CSVs can be found whether the plugin runs from `CompiledPlugin/`, a Revit add-ins folder, or directly from the source tree.

- **Fix 174 `_BOOL` shared parameters typed as TEXT instead of YESNO** (`b3a0e40`) — `MR_PARAMETERS.txt` and `MR_PARAMETERS.csv` declared every parameter whose name ends in `_BOOL` (visibility, style, cost-data, box/leader params) with `DATATYPE = TEXT`. When `AddSharedParameters()` in `TagFamilyCreatorCommand` detected a StorageType mismatch (project already had e.g. `CST_LOCAL_MAT_BOOL` registered as Yes/No from a prior load), it tried to remove and re-add it as Text — Revit rejected this with an "Error - cannot be ignored" modal. All 174 entries corrected to `YESNO` in both files.

## Test plan

- [ ] Open Revit with the updated plugin
- [ ] Run **Create Tag Families**: verify the "CSV plan match" rate in the dialog is ≥ 90% (was ~8%)
- [ ] Confirm no "cannot be added… conflicts with existing… type" error dialog appears during tag family creation
- [ ] Run **Load Shared Params**: confirm `CST_LOCAL_MAT_BOOL` and other `_BOOL` params bind as Yes/No without errors
- [ ] Spot-check that T1–T3 labels on existing families are unaffected (T1–T3 are outside the tier-variation plan and are never touched by `TagConfigCsvReader`)

https://claude.ai/code/session_01S8L4KoxvKHoJ4osBMcinvk


beckykyomugisha:
## ⚠️ Review before merge — semantic conflict with main's Phase 187i (do not blind-merge)

`git merge-tree` reports **zero textual conflicts**, but that is misleading here. This branch and `main` independently worked the **same** tag-`_BOOL` datatype problem and reached **opposite** conclusions:

- **This PR** (`b3a0e40`) flips **174 `_BOOL` params → YESNO**.
- **main / Phase 187i** (+ ~7 sibling commits) retyped the overlapping label-formula `_BOOL` params **→ TEXT** (`TAG_PARA_STATE` etc.), because Revit tag-label formulas can't concatenate Yes/No params. main still ships a runtime guard (`"label params have non-TEXT types"`).

A clean file-by-file merge can therefore produce a **globally inconsistent** parameter set — reintroducing the very `StorageType` conflict both branches were fixing. (This branch also churned both directions across its 21 commits, so its own net state needs untangling first.)

### Recommended path (build-capable agent, not a UI merge)
1. **Rebase/compare onto the *current* main** — base here is the pre-KUT tip, ~3,901 commits stale.
2. **Split** the standalone bug-fixes — CSV banner-drop parser (`38cf679`), `ResolveCsvPath` fallbacks (`fa9d955`), `GetExpectedStorageType` Revit-2025 fix (`16aa22c`), SaveAs path sanitisation (`1a4c2e4`) — from the `_BOOL` datatype churn. The standalone fixes are likely safe to take.
3. **Reconcile datatypes once** against Phase 187i (keep main as the datatype baseline; apply only genuinely-missing deltas; rule of thumb: label-formula `_BOOL` → TEXT, non-label logic `_BOOL` → YESNO). Much of the datatype work here may already be **superseded**.
4. **Build clean + run the registry↔CSV drift validators** (`DataPipelineEnhancementCommands`) and, in Revit, **Create Tag Families** (match rate ≥ 90%, no type-conflict modal) + **Load Shared Params**.

Leaving as **draft** pending this reconciliation.

---
_Generated by [Claude Code](https://claude.ai/code)_

StingD85:
Superseded by #316, which reconciles this work onto **current** main.

This branch's base was stale and its net `_BOOL` retyping is YESNO-ward — the **opposite** of main's Phase 187i resolution (label-formula `_BOOL` → TEXT). Merging it directly would re-break tag-family creation despite `merge-tree` reporting clean.

#316 instead:
- **takes** the 3 standalone datatype-neutral fixes main lacked (`38cf679` CSV banner parser, `fa9d955` match-rate 8%→100%, `1a4c2e4` SaveAs path sanitise);
- **carries over the one genuine datatype delta** main missed — `CLN_HOIST_TRACK_BOOL` YESNO→TEXT (a label-formula param), aligned across all sources;
- **drops** `16aa22c` (moot — main has none of the GetExpectedStorageType machinery and builds clean on Revit 2025) and the 145-param YESNO churn (superseded by Phase 187i).

Verified: build 0/0, validate-data CI 0 drift, `_BOOL` TEXT=174/YESNO=109. Closing in favour of #316.

StingD85:
Recording why this was closed, and why the branch is being kept rather than merged — plus something the check turned up that is still open.

## The problem this PR found was real

`MR_PARAMETERS.txt` declared every `_BOOL` parameter as `TEXT`, and `AddSharedParameters()` hit an unignorable Revit modal when a project already had one registered as Yes/No. That was a genuine defect, and the first two fixes here (the `⚠ WARNING PARAMETERS` banner dropping every parsed tier row, and the CSV path fallbacks) landed on their own merits.

## Why the branch's later direction was not taken

The branch's last commits go the other way — `TAG_7_SECTION_VISIBLE_A..F_BOOL` from `YESNO` to `TEXT`, then *"make ALL label-referenced `_BOOL` params TEXT at source"*. Phase 188 chose the opposite: keep the native type and add a **`_TXT` display mirror** beside it.

Making a flag `TEXT` at source fixes the label and breaks three other things:

- **Scheduling and sorting.** A `TEXT` "Yes"/"No" sorts as text and cannot be filtered as a boolean in a Revit schedule.
- **Revit's own formatting.** A native `YESNO` renders per the user's locale; a text literal does not.
- **Round-tripping.** Anything writing the parameter must now agree on the exact spelling of "Yes"/"No" forever.

The mirror keeps `YESNO` for the model and gives the label a `TEXT` companion to read. Both consumers get the type they need, and neither constrains the other. That is what PR #795 shipped for fourteen flags (`ASS_TIEIN_CONNECTED_BOOL`, `CST_LOCAL_MAT_BOOL`, `RGL_KCCA_APPROVAL_REQ_BOOL` and eleven more), and why merging this branch now would *revert* Phase 188 for the parameters it touches.

The branch is deliberately not deleted — it is the record of the alternative.

## What checking this turned up: 16 flags are still exposed

`LabelParamTypeValidator` collects label parameters by reading the `"parameter"` / `"param"` **keys** in each category's `tier_2` / `tier_3` / `warnings` arrays. It never sees a parameter named **inside a `"formula"` string**.

`LABEL_DEFINITIONS.json` contains formulas of the form:

```json
"formula": "if(TAG_7_SECTION_VISIBLE_A_BOOL, ASS_TAG_7A_TXT, \"\")"
```

**Sixteen `_BOOL` parameters appear only that way — never as a `parameter` key.** All sixteen are still `YESNO`, none has a `_TXT` mirror, and the validator cannot flag any of them:

`TAG_7_SECTION_VISIBLE_A..F_BOOL` (6) and `TAG_PARA_STATE_2..10_BOOL` (10).

By the rule the validator itself states — *"Revit label formulas cannot use YESNO parameters as the condition of `if(...)`"* — these are exactly the broken case. This PR's *"12 more"* commit was reaching for the same set; it had the diagnosis right and the remedy wrong.

The fix is a mirror for each, as #795 did for the other fourteen, plus teaching the validator to parse parameter names out of `formula` strings so the gap cannot reopen. Logging separately.

StingD85:
**Correction to my previous comment — the "16 exposed flags" finding is wrong. There is no defect. Please disregard that section.**

I claimed `TAG_7_SECTION_VISIBLE_A..F_BOOL` and `TAG_PARA_STATE_2..10_BOOL` were unmirrored, still `YESNO`, and invisible to validation. The first two facts are true; the conclusion drawn from them is not.

## What I missed

Those sixteen are **gates**, not label values, and they have their own validator: `LabelParamTypeValidator.ValidateGateConditionForms`, called from the Create Tag Families pre-flight. It scans `calculated_value_templates`, the per-category `warnings` formulas, and the CSV `Formula` column, and applies the rule I had backwards:

| gate storage | form in the formula | verdict |
|---|---|---|
| `YESNO` | bare — `if(GATE, …)` | **correct** |
| `YESNO` | `GATE = "Yes"` | flagged |
| `TEXT` | bare | flagged — expected `GATE = "Yes"` |
| `TEXT` | `GATE = "Yes"` | correct |

`IsTagGateName` names `TAG_PARA_STATE_*_BOOL` and `TAG_7_SECTION_VISIBLE_*_BOOL` explicitly, and `GateRefRegex` matches them. So these sixteen are in scope of a validator built for exactly this question, and `if(TAG_7_SECTION_VISIBLE_A_BOOL, ASS_TAG_7A_TXT, "")` against a `YESNO` gate is **the expected form, not a broken one**.

`LabelParamTypeValidator.ValidateSourceFile` ignores them correctly: it governs parameters a label *displays*, which must be `TEXT`. A gate is a condition, not a displayed value, and stays `YESNO`. Two rules, two validators, two disjoint sets — not one rule with a blind spot.

## So the design is more complete than I described

PR #795 mirrored fourteen flags because labels were **displaying** them (`"param": "CST_LOCAL_MAT_TXT"`). It stopped at fourteen because fourteen was the correct answer, not because the gate could only see fourteen.

The rest of the previous comment stands: the mirror approach preserves the native type for scheduling, locale-aware rendering and round-tripping while giving labels the `TEXT` they need, and this branch's "TEXT at source" direction would revert that. This PR's *"12 more"* commit was reaching for the sixteen gates — which needed no change at all.

## How the error happened

I derived a blind spot from a grep rather than from reading the consumer. `ValidateSourceFile` genuinely does not parse names out of `formula` strings — I checked that correctly — and I concluded nothing did, without searching for another consumer. `ValidateGateConditionForms` is forty lines below it in the same file.

No code or data was changed on the strength of the wrong claim; the branch I had opened to make the fix is discarded.

PR #319, opened 2026-06-18
Export Centre — complete pipeline overhaul (rebased on main, supersedes #318)

beckykyomugisha:
## Why this exists

PR #310 was closed in favour of #318 on the basis that `main` already covers its PDF-pipeline items. I verified `main` directly and that's **only partly true** — `main` still carries the original bugs this work fixes. This branch is rebased cleanly onto current `main` and is now a **complete superset of #318**: it includes the bug fixes #318 dropped *and* the scheduler + delta automation #318 adds. **#318 can be closed in favour of this.**

## The bugs on `main` that #318 does not fix

- `ExportCenterEngine.ExportSinglePdf` still sets **`Combine = false`**. In Revit 2022+ that makes Revit ignore `PDFExportOptions.FileName` and name files by its own rule (`Sheet-<name>.pdf`), so `File.Exists(<custom stem>.pdf)` is always false → **every per-sheet export is reported "PDF export returned false"** even though a (mis-named) PDF was written. This is exactly the "0 successful / 3 failed" field symptom.
- Watermark only runs in the *combined* path — **OnePerSheet mode never stamps the watermark**.
- No output verification (0-byte / corrupt), no atomic write, no locked-file handling.

## What this branch contains (4 commits, cherry-picked onto `main` with no conflicts)

**Bug fixes (the part #318 drops)**
- **Export-failure fix** — per-sheet export uses `Combine = true`, so the ISO 19650 filename is produced exactly and the success check passes.
- **Robustness pack** — `RobustPdfExport`: render to a sibling temp folder, verify (non-zero bytes + opens + page count via PDFsharp), post-process the verified temp, then atomically rename into place. Locked-target detection ("close it in Bluebeam/Acrobat") + retry-with-backoff; Revit failure text captured via `Application.FailuresProcessing`.
- **Watermark on per-sheet PDFs**; **combined-naming cleanup** (`…-XX-DR-Z-ALL-S2-P01.pdf`, no `_All`); **combine-mode UX** labels; **Hub** ribbon button next to Auto Tag.

**Scheduler + delta + polish (parity with #318, plus more)**
- **Delta automation** — built-in sets "Changed Since Last Export" / "Resume Last Failed"; per-sheet last-export records persisted after each run.
- **ScheduledExportRunner** — headless driver for `ScheduledExports`, manual command + opt-in DocumentSaved trigger.
- **Watermark & combine polish** — per-page text (token + per-suitability auto-phrase), tiled diagonal, auto drawing-register cover page, combine-order selector. Post-processing unified through one `PostProcessExport` orchestrator (bookmarks → watermark → cover) on the verified temp.
- **SHA-256 checksum manifest** + **post-export shell hook**.

`main`'s existing advances (DwgMerger / OdaConverter / NwcExporter / XmlWriter / ISO-naming defaults / R2025 fixes) are untouched.

## Testing / caveat
Written against documented Revit 2025+ / PDFsharp 6 / .NET 8 APIs but **not compiled** in the Linux sandbox. Verify in Revit: a OnePerSheet export with watermark on (correct filenames + diagonal stamp on every page), a "Single combined PDF" export, and the cover-sheet/scheduler paths. The `Combine = true` per-sheet pattern is the standard Revit 2022+ workaround for custom PDF filenames.

### DIFF
--- a/StingTools/Core/StingToolsApp.cs
+++ b/StingTools/Core/StingToolsApp.cs
@@ -1413,6 +1413,18 @@ namespace StingTools.Core
                 {
                     StingLog.Warn($"DocumentSaved geometry sync trigger: {geoEx.Message}");
                 }
+
+                // Unattended export — run any due scheduled-export jobs. No-op unless the
+                // user opted in (ExportCenterState.EnableSaveTriggeredSchedules) and a job
+                // is actually due; never throws out of the save handler.
+                try
+                {
+                    StingTools.Docs.ScheduledExportRunner.RunDue(doc, fromSave: true);
+                }
+                catch (Exception schEx)
+                {
+                    StingLog.Warn($"DocumentSaved scheduled export: {schEx.Message}");
+                }
             }
             catch (Exception ex)
             {
--- a/StingTools/Core/StingToolsApp.cs
+++ b/StingTools/Core/StingToolsApp.cs
@@ -2217,6 +2217,7 @@ namespace StingTools.Core
                 ("Tag3D",                "3D Tag",        "T3", DrawingColor.Crimson,      typeof(HubTag3DCommand).FullName),
                 ("CreateTagFamilies",    "Tag Families",  "TF", DrawingColor.DarkCyan,     typeof(HubCreateTagFamiliesCommand).FullName),
                 ("AutoTag",              "Auto Tag",      "AT", DrawingColor.DarkGreen,    typeof(HubAutoTagCommand).FullName),
+                ("ExportCenter",         "Export Center", "EX", DrawingColor.RoyalBlue,    typeof(HubExportCenterCommand).FullName),
             };
 
             var buttons = new List<PushButtonData>(12);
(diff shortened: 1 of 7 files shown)


StingD85:
Superseded by #322. #319 was branched before #318 merged, so it re-added `ScheduledExportRunner.cs` + delta automation already on `main` (merge-tree conflicts) and was never compiled.

#322 rebases this onto current `main`, taking **only** the export bug fixes (Combine=true honouring the ISO filename, RobustPdfExport with temp-render+verify+atomic-rename+retry, per-sheet watermark, no `_All` naming) and keeping main's #318 scheduler/delta. It builds clean (0 errors). #322 is a DRAFT pending the Revit smoke test that both this PR and #318 skipped — it will be marked ready + merged once a human confirms 'N of N succeeded' in Revit.

Closing in favour of #322.

StingD85:
Superseded by #322, which rebases the real per-sheet PDF fix onto current main and adds the STING-Hub Export Center button + SELECT-panel bulk-selection/grouping controls. Closing as superseded.

PR #448, opened 2026-07-20
fix(server): a cache outage must not turn a 404 into a 500; Redis in CI

StingD85:
Phase 0 of the ISO 19650 IM runner: turn main's CI green again.

## What was actually wrong

The runner's brief attributed the failures to `AuthController.Login` throwing on
Redis. It does not — it fails open, and the CI log says so ("proceeding
(fail-open)"). That frame appeared in the log only because the gate script
printed a 30-line tail on abort, and the tail happened to land there.

The real path is `ProjectAccessAttribute` line 73. Its project-visibility cache
(Phase 175 P1-13) is an **optimisation** over the authoritative
`ProjectVisibility.CanSeeProjectAsync` query, but `GetStringAsync` /
`SetStringAsync` were unguarded. An unreachable Redis propagated out of the
filter, so **every gated request 500'd** — including the cross-tenant ones the
filter exists to answer with 404.

That is a production robustness bug, not just a CI gap: a Redis blip currently
takes out every project-scoped endpoint. It is not a security hole (it errors,
it does not leak), but it is an availability one.

## Three fixes

1. **`ProjectAccessAttribute` fails soft.** Read failure → treated as a miss;
   write failure → "not cached". The decision comes from the DB either way, so
   correctness is unchanged; only the hot-path saving is lost while Redis is down.
2. **`redis:7` service in CI**, beside the `postgres:16` that #444 added. Not to
   make the suite pass — the code now fails soft — but for prod parity and
   **speed**: ~900 Redis calls were each burning a full 5 s timeout, which is most
   of why the run took 30 minutes.
3. **`check-new-failures.sh` can now report what it finds.** Its summary
   precondition matched only the VSTest console format (`Passed!`/`Failed!`). A
   solution-level `dotnet test` goes through MSBuild and, when `VSTestTask`
   returns false, emits only `Test Run Failed.` / `Total tests: N`. So the gate
   aborted with *"no test summary in output — the run did not complete"* on
   precisely the runs whose diff was worth reading. **Both CI runs cited in the
   brief failed this way — neither ever printed a NEW-failures list.** Replayed
   against the real CI output, the fixed script correctly lists the five.

## Red-then-green

- Reproduced locally by pointing `Redis:Connection` at a dead port: the five
  named tests fail `Expected: NotFound / Actual: InternalServerError`, identical
  to CI. With the guard, they pass **with Redis still unreachable**.
- Adding Redis to CI would have quietly stopped exercising the outage path — the
  five only caught it because Redis was absent. So `ProjectAccessCacheOutageTests`
  pins it: an `IDistributedCache` that throws on every operation, asserting both
  directions (foreign project still 404s; visible project still allowed — a
  fail-soft that denied everything would satisfy the first assertion alone).
  Verified red without the fix, green with it. It drives the filter directly
  rather than standing up a second `WebApplicationFactory`, which is unreliable
  here (DEP-7) — same approach as Phase 222.
  - My first draft of that test passed **vacuously**: `PlanscapeDbContext`'s
    global tenant filter falls back to `Guid.Empty` without an `ITenantContext`,
    making every project invisible. Fixed by wiring a stub; the mirror assertion
    is what caught it.

## Two corrections to the brief

- **No baseline lines to delete.** The brief said the last run showed 31 baseline
  entries passing. It did not — no `now PASS` notice appears in either run,
  because the gate aborted before computing one. Against the real CI output the
  fixed script reports **zero** baseline entries passing. Locally many *do*
  appear to pass, but that is the Redis-on-6379 artefact that produced the stale
  baseline in the first place. **CI is the authority — I deleted nothing.**
  Once this lands, CI resembles that Redis-having environment, so the next run may
  legitimately flip a batch to passing; those should be deleted in a follow-up
  using that run's `now PASS` list.
- **DEP-6b (stretch) skipped, and not because it fought me.** ROADMAP already
  records that the Redis-backed limiter fails to trip *with* a reachable Redis, so
  adding Redis to CI does not unblock it. ROADMAP annotated accordingly.

## Verification

| Check | Result |
|---|---|
| `dotnet build Planscape.sln` (`-t:Rebuild`) | 0 errors, **14 warnings — unchanged** |
| `dotnet build StingTools.csproj` (Revit 2025) | 0 errors, 0 warnings |
| Full suite gate | `No new failures. (66 failing, all known.)` |
| Collateral check | 67 unique failures both with and without the change; the only delta is a per-run random DEP-7 victim (`PersonalAccessToken…` on stock, `AuditCategories…` with the fix) |

DEP-7 recurrence observed twice, as a different random victim each run — consistent
with the documented process-global `JobStorage.Current` teardown. Not baselined.

## Also

`pull_request` now triggers on `.github/workflows/planscape-server.yml`, matching the
existing `push` trigger — a workflow change should be exercised on its own PR.

## Deploy

None needed. No Render action taken or required.

## Runtime-unverified

This PR touches server + CI only; no Revit plugin code changed (plugin built solely as
a ground-rule gate). **No in-Revit checklist applies.**


StingD85:
## CI result — green ✅

| | before (#446 on main) | after |
|---|---|---|
| Workflow | ❌ failure | ✅ **pass** |
| Gate 2 output | *"no test summary in output — the run did not complete"* | **`No new failures. (66 failing, all known.)`** |
| Total / passed / failed | 440 / 362 / 78 | **442 / 369 / 73** |
| Gate 1 (Postgres) | 6/6 | 6/6 |
| Wall clock | **30m 34s** | **2m 25s** |

The five named tenant-isolation tests pass. The +2 total is
`ProjectAccessCacheOutageTests`. The 12× speedup is the ~900 five-second Redis
timeouts going away.

**DEP-7 recurrence, as the ground rules anticipated.** The first attempt failed on
one NEW entry — `AuditCategoriesConfiguredTests.Get_WithConfig_ETagDiffersFromBuiltInOnly`,
which builds a second `WebApplicationFactory` at line 224 and is therefore a textbook
DEP-7 victim. It also appeared as a random victim in local runs (a *different* test,
`PersonalAccessTokenTests.ActiveTokenCap_IsEnforcedAtTwenty`, was the victim on stock
main). Re-ran the job per protocol; it did not recur. **Not baselined.**

**Correcting my own speculation in the PR body:** I suggested that once CI had Redis,
a batch of baseline entries might legitimately flip to passing and want deleting in a
follow-up. That did **not** happen — the fixed gate reports **zero** `now PASS` entries
in CI even with Redis running. The 66-entry baseline is accurate for CI as-is, and no
follow-up deletion is warranted. The "many pass locally" effect is something other than
Redis availability alone.

PR #449, opened 2026-07-20
fix(drawings): P0 remediation from drawings-production review

StingD85:
Implements the **P0** tier of the drawings-production deep review (`docs/DRAWINGS_PRODUCTION_REVIEW.md`, PR #447 — ~85 findings, 5 Critical). One numbered fix per commit.

Every commit builds clean against Revit 2025 (`0 Warning(s), 0 Error(s)`, matching the repo baseline). **Nothing here has run inside Revit** — smoke-test list at the bottom.

## Findings fixed

| # | Finding | Fix |
|---|---|---|
| P0-1 | **C-1** most style-pack payload runtime-dead | Setter-only JSON aliases on `ViewStylePack` / `StyleVgOverride` / `StyleFilterRule` |
| P0-2 | **C-2 + E-1** both `ResolveExtends` folds strip fields | New `ExtendsMerge` generic default-aware overlay |
| P0-3 | **W-1** two penetration workflow presets inert | Step-key rename + 3 new `ResolveCommand` cases |
| P0-4 | **A-11** `LABEL_DEFINITIONS.json` mojibake | Text-level repair of 530 strings / 1,171 chars |
| P0-5 | **V-3, V-7, V-8** filter construction + data | `LogicalOrFilter`; 12 dead filter definitions repaired |
| P0-6 | **C-3** producer sheet identity ignores context | `STING_SHEET_CONTEXT_TXT` added to the sheet key |

## Measured results

Where a claim could be tested outside Revit, it was. `ViewStylePack.cs` has no Revit dependency, so it compiles standalone; `ResolveExtends` and `SheetKey` take only POCOs/strings, so they were invoked on the built assembly via reflection.

**C-1** — deserialising the real corporate JSON, before → after:

| | main | fixed |
|---|---|---|
| filter rules bound | 19 | **97** |
| `corp-healthcare-pressure` rules | 0 | **6** |
| vgOverride `projColor` / `projWeight` | 107 / 107 | **295 / 316** |
| vgOverride `cutColor` / `cutWeight` | 45 / 45 | **54 / 59** |
| rule `surfFgColor` / `projLinePattern` | 0 / 0 | **15 / 5** |

Each figure equals that key's exact count in the JSON, and a serialize/deserialize round-trip preserves all 97 rules.

**C-2 / E-1** — 14 packs declare `templateMode: "managed"`; **0 now lose it** through the fold (was: all 14). 3 more correctly *inherit* managed from a managed parent. A child retains a parent's `titleBlockParams`, `isoNaming`, `packageId`, `system`, `materialPack`, `tagTextSizeMm`, `titleBlockSymbolType`, while its own values and overrides still win.

**W-1** — all 13 step tags across both files resolve, checked against the 495 `ResolveCommand` cases parsed out of `WorkflowEngine.cs`.

**A-11** — JSON re-parses, key count unchanged at 62,745, zero residual mojibake markers, every remaining non-ASCII run is a single clean codepoint.

**V-3/V-7/V-8** — across all 298 filters: 0 leaf rules with an op outside the factory's 14-op grammar, 0 nodes that are neither leaf nor compound, 0 remaining `Family Name` params.

**C-3** — per-level contexts produce distinct, stable sheet keys that don't collide with each other, with the empty-context key, or across packages.

## Findings the review missed, fixed here

- **`surfFgColor` (15×) and `projLinePattern` (5×) were also unbound.** The Phase 139 alias pass covered the four line keys and stopped short of these. Found by diffing *every* key in the JSON against every bound `JsonProperty`, rather than only checking the keys the review named.
- **`plumb-high-pressure-zone` and `plumb-backflow-risk` use a third rule schema** — `{"kind":"compound","op":"or","operands":[…]}` instead of `{"logic":"or","rules":[…]}`. Those nodes satisfy neither `IsLeaf` nor `IsCompound`, so `BuildFilter` returned null silently. The review found 2 invalid leaf ops; these are 2 more dead filters.
- **`DrawingType.extends` could never inherit a field with a non-null POCO default** (`PaperSize` "A1", `Discipline`/`Phase` "*", `Purpose` "Plan", `Orientation` "Landscape", `Scale` 100, `DetailLevel` "Medium", both sheet patterns). The old `!IsNullOrEmpty` guard treats a child's *default* as a value, so an A0 parent always resolved back to A1. Caught by a test assertion, not by reading. Safe to change: 0 of 93 shipped types use `extends`.
- **`LABEL_DEFINITIONS.json` damage is much wider than "em-dashes and mÂ²"** — 15 patterns including `₂` (413×, chemical formulae in medical-gas labels), `⚠` (65×, the leading glyph of every warning label) and `Δ` (132×). Also, **5 characters in the file were already correct**, so the review's suggested whole-string re-encode would have corrupted them; repaired the 15 known-bad runs instead.

## Deliberate scope calls

- **`ViewStylePack`'s six already-working scalars keep their original guards.** That fold runs on every `Get()`, and changing `lineWeightScale` merge semantics would alter rendering. The default-clobbers-parent issue is therefore fixed for `DrawingType` (dormant path) but still open for style packs — recorded in ROADMAP.
- **C-3 falls back rather than duplicating.** If `STING_SHEET_CONTEXT_TXT` is not bound, `ReadSheetContext` returns `null` (not `""`) and the producer reuses the old (type, package) match **with a warning naming the parameter**. Reproducing the old stacking is bad; silently minting a duplicate sheet on every run would be worse.
- **`DrawingTypes_FromScopeBoxes` resolves to `GenerateFromScopeBoxesCommand`.** The dock-panel button for that tag runs a separate inline reimplementation in the handler — that's review finding **W-5**, left for a later pass and noted in a comment at the new case.
- **Nothing skipped as wrong or already-fixed.** Every P0 finding reproduced. Two were slightly mis-described — C-1's short-form keys are on `StyleVgOverride` (the `StyleFilterRule` ones were already aliased in Phase 139), and V-7's 8 filters carry *no* `kind` field rather than `kind: builtin` — but the defects and the fixes stand in both cases.

## Must be smoke-tested in Revit before merge

| Area | Check |
|---|---|
| Style packs (C-1) | Apply a `corp-healthcare-*` pack — filter colour coding appears where it previously did not |
| Extends fold (C-2) | A pack with `templateMode: "managed"` engages the managed-template path |
| Workflows (W-1) | Run Penetration Sweep and Penetration Register end to end |
| Filters (V-3/V-7/V-8) | The 12 repaired filters mint; a multi-material class filter actually selects elements |
| Producer (C-3) | Run `LoadSharedParams` **first**, then produce per level over ≥2 levels: one sheet per level, and a re-run reuses them rather than adding more |
| Labels (A-11) | A medical-gas tag renders CO₂ / N₂O with a real subscript, and warnings with ⚠ |

`LoadSharedParams` must run once per project before C-3 behaves correctly — that is what binds the new `STING_SHEET_CONTEXT_TXT`.

P1 was not started; the full P1/P2 tail is listed as open in `docs/ROADMAP.md`.

### DIFF
--- a/StingTools/Commands/Drawing/BatchProduceCommands.cs
+++ b/StingTools/Commands/Drawing/BatchProduceCommands.cs
@@ -344,20 +344,31 @@ namespace StingTools.Commands.Drawing
                 {
                     contextsToProduce = grids.Select(g =>
                     {
-                        var bb = new BoundingBoxXYZ();
                         try
                         {
                             var c = g.Curve as Line;
                             if (c == null) return null;
                             var origin = (c.GetEndPoint(0) + c.GetEndPoint(1)) * 0.5;
-                            double depthFt = sec.DepthMm / 304.8;
-                            var d = c.Direction;
-                            // perpendicular = (-d.Y, d.X)
-                            var perp = new XYZ(-d.Y, d.X, 0).Normalize();
                             var len = (c.GetEndPoint(1) - c.GetEndPoint(0)).GetLength();
-                            var widthFt = len * 0.5 + 5.0 / 0.3048;
-                            bb.Min = origin - perp * widthFt + new XYZ(0, 0, -3.0 / 0.3048);
-                            bb.Max = origin + perp * widthFt + new XYZ(0, 0, 30.0 / 0.3048);
+
+                            // P-5: one shared frame builder with the producer.
+                            // This used to offset Min/Max by a perpendicular
+                            // vector in MODEL space with an implicit identity
+                            // transform — so height landed on the wrong axis
+                            // (the producer put it on Y, this on Z), and for a
+                            // north-south grid the perpendicular's negative
+                            // components made Min > Max, which CreateSection
+                            // rejects outright. The section now cuts ALONG the
+                            // grid and looks perpendicular to it, which is what
+                            // "section along grid line" means.
+                            var bb = DrawingProducer.BuildSectionBox(
+                                origin:       origin,
+                                cutDirection: c.Direction,
+                                halfWidthFt:  len * 0.5 + 5.0 / 0.3048,
+                                bottomZ:      origin.Z - 3.0 / 0.3048,
+                                topZ:         origin.Z + 30.0 / 0.3048,
+                                depthFt:      sec.DepthMm / 304.8);
+
                             return new DrawingContext { CustomBounds = bb, Tag = "Grid-" + g.Name, PackageId = preset.PackageId };
                         }
                         catch (Exception ex) { StingLog.Warn($"Suppressed: {ex.Message}"); return null; }
--- a/StingTools/Commands/Drawing/DrawingHealTitleBlocksCommand.cs
+++ b/StingTools/Commands/Drawing/DrawingHealTitleBlocksCommand.cs
@@ -66,7 +66,8 @@ namespace StingTools.Commands.Drawing
                     MainContent =
                         "This rewrites every title-block parameter declared by each sheet's profile " +
                         "(titleBlockParams + per-symbol overlays). Scale / detail level / view template / " +
-                        "view-style pack / annotation are NOT touched. Locked sheets are skipped.",
+                        "view-style pack / annotation are NOT touched. Style-locked sheets are skipped, and " +
+                        "title blocks frozen with PRJ_TB_LOCK_BOOL are left untouched.",
                     CommonButtons = TaskDialogCommonButtons.Ok | TaskDialogCommonButtons.Cancel,
                     DefaultButton = TaskDialogResult.Ok,
                 };
@@ -75,6 +76,7 @@ namespace StingTools.Commands.Drawing
                 var auditRows = new List<HealRow>();
                 int totalParams = 0;
                 int healed = 0;
+                int lockedSkipped = 0;
                 using (TitleBlockParamApplier.Batch())
                 using (var tx = new Transaction(doc, "STING — Heal Title Blocks"))
                 {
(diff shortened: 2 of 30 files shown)


StingD85:
## Managed-template hardening added (E-3, E-4, E-6)

Per review feedback: the C-2 fix takes the managed-template path from unreachable to live for 14 packs, so `ManagedTemplateSyncer`'s known bugs are cleared in the same series rather than being activated. Commit `75366653a`.

**E-3 — minting produced junk views, not templates.** `ResolveSeed` falls back to a non-template live view when there is no `STING - ` seed, i.e. on every fresh project. `CopyElement` on a live view yields another live view: `SetViewTemplateId` rejects it, the `IsTemplate`-filtered lookup can never find it again, so the next run mints another — that's the `_(2)` / `_(3)` accumulation. Non-template seeds now use `View.CreateViewTemplate()`; template seeds keep `CopyElement`, which was already right. The result is `IsTemplate`-checked, and a failed mint is deleted rather than left behind.

The rename fallback was the *other* half of the accumulation: `templateName + "_(2)"` by construction never matches the lookup, so every run added one more. It now fails loudly, names the likely conflict, cleans up, and creates nothing.

**E-4 — the review has this backwards, in both directions.** Checking `ViewDiscipline` in `RevitAPI.dll`:

| | shipped | actual enum |
|---|---|---|
| Architectural | 1 | 1 ✅ |
| Structural | 2 | 2 ✅ |
| Mechanical | 4096 | **4** ❌ |
| Electrical | 4097 | **8** ❌ |
| Plumbing | 4098 | **16** ❌ |
| Coordination | 4095 | 4095 ✅ |

`VIEW_DISCIPLINE` is a bit-flag parameter and `Coordination = 4095` is the all-bits union — so the value the review flagged as invalid was already correct, and the one it called "the only correct one" was not. Real defect is the three MEP disciplines (`corp-standard-hvac` / `-elec` / `-plumb`). Now reads the enum members directly so it can't drift again; verified by invoking `ResolveViewDiscipline` against the enum for all six values the corporate packs use.

**E-6 — the built list was discarded.** Now computes the complement the API actually wants: `SetNonControlledTemplateParameterIds(GetTemplateParameterIds() minus managed ids)`, so a pack controls exactly what it declares. That is what stops a seed controlling `VIEW_SCALE` from overriding the per-profile `DrawingType.Scale` — `scale` is deliberately excluded from `DefaultManagedFields` for precisely this reason. Wrapped so a failure degrades to previous behaviour with a warning instead of aborting the sync.

### Smoke-test list updated

Added the requested check plus two covering the other two fixes:

| Area | Check |
|---|---|
| **Managed minting (E-3)** | **Apply a `templateMode:"managed"` pack on a fresh project with no `STING - ` seed templates: a real view template is created, and repeated runs add no `_(2)` / `_(3)` junk views** |
| Discipline (E-4) | A managed `corp-standard-hvac` / `-elec` / `-plumb` template shows the right discipline in Revit |
| Template control (E-6) | With a managed pack applied, per-profile `DrawingType.Scale` still wins over the seed template's scale |

Builds clean (0 warnings, 0 errors). Proceeding to P1-1 (token substitution unification).

PR #452, opened 2026-07-20
fix(iso19650): consent before moving files; one coord-log contract; a gate that works

StingD85:
Phase 1 remediation for the ISO IM runner. **Stacked on `claude/iso19650-consolidation`,
not `main`** — it repairs that branch's own work, so it targets it directly. That branch
has no PR yet; this should merge into it before it goes to main.

## Why this exists

The runner's Phase 1 (ProjectStore + migration + path discipline) looked done on the
consolidation branch. Audited properly, the **paths** were done and little else, and the
enforcement was illusory:

> `tools/path_discipline_baseline.txt`: *"ALL sibling writers migrated onto StingPaths
> (WP6 complete). Baseline is now ZERO."*

That zero was an artifact of the regex. Proof — the same injected residue, both gates:

```
NEW gate: FAILED -- legacy bucket name(s) outside the resolvers (1 occurrence)
          FAILED -- new hand-rolled _BIM_COORD path(s): 3, baseline 1     exit 1
OLD gate: OK -- no new sibling _BIM_COORD paths (0 remaining across 0 files)  exit 0
```

The old gate had three holes (no `STING_BIM_MANAGER` / `_bim_manager` in the pattern;
`[^)]*` cannot cross a close-paren; a five-name variable allowlist that missed the
dominant two-line idiom) — **and it was wired into no workflow at all**, so it had never
run in CI regardless.

## The six items

| # | Item | State before |
|---|---|---|
| 5 | Migration behind consent | **Moved files on every document open, unprompted**, deleted drained sources, no record |
| 2 | Coord log one contract | Three-way split; BCC timeline silently empty |
| 3 | Open-issue miscount | BCC reported **0 open issues** on any store current code writes |
| 4 | Meetings tab | Bound to two collections that were **never assigned** |
| 6 | Stranded artefacts | Saved searches **lost on a loop**; distribution groups never found |
| 1 | Path gate | Wired into nothing; baseline fictitious |

### 5 — the blocker
`MigrateFromLegacy` ran inside `DocumentOpened` on *every* open while
`FolderConsolidateCommand`'s header claimed *"It NEVER runs automatically."* The engine
now enforces the claim: `consented: true` required, at-most-once via
`.sting_consolidation.json`, **no source ever deleted** (drained folders renamed
`*.migrated_yyyyMMdd`), every relocation recorded source-to-destination with an undo hint.
Document-open only detects and logs. The folder-setup dialog and `FolderMigrate` tag —
which both migrated the instant they were clicked, with no preview — now route through one
shared `RunWithConsent`.

### 2 — coord log
`WarningsManager` wrote JSONL to `.jsonl`; `MaterialAuditLogger` wrote JSONL *into* `.json`
under a different field schema; all four readers opened `.json` and whole-file
deserialised, throwing on entry 2 into a swallowing catch. Now `CoordLog` (paths, legacy
read fallbacks) + `CoordLogFormat` (codec). **9 unit tests** in `StingTools.Tags.Tests`,
including the multi-entry case that was the real failure — a single-entry log parsed fine
under the old reader, which is why this survived.

### 3 — the miscount
`IndexOf("\"status\":\"OPEN\"")` matches only *compact* JSON; `CoordStores.WriteArray`
writes `Indented`. The branch's own change broke the count. Now parsed and routed through
`IssueStatusNormalizer`.

### 4 / 6
Meetings + action items now populated in `BuildCoordData` from `CoordStores.Meetings(doc)`;
**516 lines** of dead builder deleted. `SearchQueryBuilder` was the one copy of
`ResolveProjectRoot` (of ten) missing the canonical branch — saved searches were written to
the sibling, moved away by consolidation, recreated empty, repeatedly.
`distribution_groups.json` reader now asks the owning store for its path.

### 1 — the gate
Two tiers: legacy bucket names are a **hard zero**; hand-rolled `_BIM_COORD` is
**ratcheted**. Wired into `stingtools-plugin.yml` (with `tools/**` added to the triggers).
Nine legacy sibling sites fixed, writers first — including `LpsAutoIssueRaiser`, which
wrote LPS issues to the pre-consolidation folder while every reader used
`CoordStores.Issues`, so raised LPS issues were invisible everywhere. Three genuine legacy
fallbacks carry an explicit `path-discipline: legacy-fallback` marker rather than a hidden
allowlist. **Tier-2 baseline is now honest at 139 sites across 118 files.**

## Item 6's condition — verified before deciding

You made keeping `_data/` conditional on the template-engine artefacts still working. They
do: every Template Engine v1.1 artefact (`deliverables.json`, `transmittals.json`,
`workflow_state.json`, `audit_log_*.jsonl`, `doc_sequences.json`, templates, workflows,
search index) resolves through one shared per-file `ResolveProjectRoot` used by **both** its
writer and its readers, so they cannot disagree. The two where that did not hold are fixed
here. Deviation documented in the CHANGELOG.

## Verification

| Check | Result |
|---|---|
| Plugin build (`-t:Rebuild`, Revit 2025) | **0 errors, 0 warnings** |
| `StingTools.Tags.Tests` | 143 passed / 2 failed — the 2 are pre-existing `CsiMasterFormat` failures, unchanged (136 total & same 2 before my change) |
| Path-discipline gate | exit 0 — Tier 1 = 0, Tier 2 = 139 within baseline |
| Gate actually catches regressions | Proven by injection, above |

Not run: `Planscape.sln` — this PR touches no server code.

## In-Revit verification checklist — all RUNTIME-UNVERIFIED

1. Open a project with a legacy `STING_BIM_MANAGER` / `_bim_manager` folder. Confirm
   **nothing moves** and the log says *"Run the Folders_Consolidate command…"*.
2. Run `Folders_ConsolidateAll`. Confirm dry-run CSV, the confirm dialog, and that on Yes
   files move, sources become `*.migrated_<date>`, and `.sting_consolidation.json` lists
   every move.
3. Run it again — expect *"already been consolidated — it runs once only."*
4. Click Migrate in the folder-setup dialog and the `FolderMigrate` button: both should now
   show the preview/confirm, not migrate silently.
5. Trigger coordination actions, then open BCC → coordination timeline: entries appear
   (previously always empty). Confirm a material edit also appears, with a real timestamp.
6. BCC → Issues KPI: open/critical counts non-zero on a project with open issues.
7. BCC → Meetings: meetings and action items render; overdue flags only on unresolved.
8. BCC → transmittals panel: distribution groups found rather than "No
   distribution_groups.json found".
9. Save a search, reopen the project, confirm it survives.
10. Raise an LPS issue; confirm it appears in the BCC issue list.
11. Export the coordination log to CSV/XLSX and clear it; confirm both see the same file.

## Known residual (ROADMAP IM-1..IM-6)

Notably **IM-1**: `ExtractIfMissing` still runs before consolidation, so a project whose
customised templates sit in a legacy folder gets stock templates seeded first and its own
renamed aside — nothing lost on disk, but the customisation stops taking effect. Left out
to keep to the agreed scope. Also filed: 139 remaining hand-rolled paths (IM-2), the
`ConfigPathForModel` string overload the BCC still uses (IM-3), the `issue_id`/`id` schema
fork and the four forked escalation paths (IM-4/5, both Phase 2 work), and
`StingTools.Clash.Tests` not building on this branch (IM-6, pre-existing — 14 `CS0246`,
which is why the coord-log tests live in `Tags.Tests`).

## Deploy

None. No deploy action taken or required.


StingD85:
## CI note — the new gate step did not execute on this PR

`stingtools-plugin.yml` triggers on `pull_request: branches: [main]`. This PR targets
`claude/iso19650-consolidation`, so only the labeler ran; **the "Check path discipline"
step I added has not yet run in CI.** It will first execute when this stack reaches a PR
into `main`.

What that leaves unproven, stated plainly:

- The gate's **logic** is verified — locally it exits 0 on the current tree and exits 1 on
  injected residue that the old gate passed (both transcripts in the PR body).
- The gate's **portability** is not. I ran it under Windows PowerShell 5.1; the CI step
  runs `shell: pwsh` (PowerShell Core) on `ubuntu-latest`. `pwsh` is not installed on this
  machine, so I could not execute it there. `Select-String -AllMatches -Context 3,0`,
  `Get-ChildItem -Filter`, and the `\`→`/` path normalisation are all cross-platform as
  written, but that is reasoning, not a run.

If it does misbehave under pwsh, the failure mode is loud (the step fails), not silent —
it cannot regress to the old "reports OK while seeing nothing" behaviour, because a zero
Tier-1 count is now printed explicitly alongside the Tier-2 total rather than being the
default state.

PR #459, opened 2026-07-20
feat(pm3): dayworks capture + build-up + final-account pricing; ROADMAP truth-up

StingD85:
## Scope change from the original brief

The PM-3 brief listed four items. Verification against `origin/main` found **three
already landed**, so this PR implements only the one real gap and corrects the stale
ROADMAP. Not rebuilt (anchors verified on main):

| Item | Where it already lives |
|---|---|
| Schedule-driven cash-flow S-curve | `Core/Schedule/CashFlowSCurve.cs`; consumed by EVM at `VariationAndEvmCommands.cs:636`, manual-% fallback intact at `:643-664` |
| Contractor CVR (+ L&E, line CTC, commitments) | `Core/Cost/CvrEngine.cs`, `Commands/Cost/Pm3LifecycleCommands.cs`, dispatched `Cvr_Report` |
| MSP/XER parser convergence + TASKPRED | `Core/Schedule/ScheduleImporter.cs` — `PredecessorLink` / `Relationship` / `TASKPRED` |

## The gap this fills

The tender annexure ships a **DAYWORKS SCHEDULE**, but it is a *rates framework* whose
own wording defers quantities to be "priced at final account". Nothing captured the
instructed sheets in between — no path from a CA instruction under Clause 5.7 to a
value in the Final Account. `VariationItem.RateSource` already reserved `"Daywork"` as
distinct from `"StarRate"`: the seam was cut and left unfilled.

- **`Core/Variation/DayworkModels.cs`** — `DayworkRecord` / `DayworkBuildUp` /
  `DayworkRegister`, pure and headlessly tested. Mirrors `StarRate` (labour / plant /
  materials lines reusing `StarRateLine` and its basis-by-`Unit` pricing) but rolls up
  via NRM2 percentage additions per section instead of overhead-then-profit.
- **Four ReadOnly commands** in `Commands/Cost/DayworkCommands.cs` — `Daywork_Capture`
  (headless `DayworkJson` ExtraParam **or** the new `UI/DayworkCaptureDialog.cs`),
  `Daywork_Register`, `Daywork_Price`, `Daywork_Attach`. All four dispatch in
  `StingCommandHandler` and have buttons in `BOQCostManagerPanel`.
- **Pricing uses the sheet's own build-up percentages**, not today's project config,
  so re-running `Daywork_Price` cannot silently reprice historic sheets.
- **No double-count by construction** — an attached sheet's value rides its VO into
  the waterfalls, so only *unattached* priced sheets are added standalone. Final
  Account and AFC both add it on that basis.

## Two judgement calls worth your eye

**1. The server's daywork percentages look like placeholders.** Defaults carry over
from `BoqDocument.Daywork*Pct` (115/110/112) so plugin and server agree numerically.
But the convention — pinned to RICS "Definition of Prime Cost of Daywork" and to the
annexure's own wording ("DW.10 Materials — **percentage addition** for OH&P") — is
`gross = net × (1 + pct/100)`. Under that reading a 110% materials addition prices
materials at **2.1× net**, which is not a sane tendered figure (RICS-typical
materials/plant is 10-20%; labour 115% *is* typical). Rather than price silently at
them, `DayworkBuildUp.Warnings()` flags any materials/plant addition over 50% and the
capture + pricing surfaces show it. **Worth confirming 110/112 against a real tender —
if they were meant as factors (×1.10), the server defaults are the thing to change.**

**2. Register location deviates from the brief.** The brief said
`_BIM_COORD/dayworks.json`; this uses `STING_BIM_MANAGER/dayworks.json` — the sibling
of `variations\` and `star_rates\`, which its callers read in the same breath.
`VariationEngine.SaveStarRate` (the pattern the brief also told me to follow) writes
there, and "sidecar-root unification" is still an open PM-7 item, so a third root
would make that worse. Easy to flip if you'd rather match the brief literally.

## Also in here

- `VariationItem.DayworkId` added as a **sibling** of `StarRateId`, not a reuse —
  `RateSource` already distinguishes the two build-up documents, and punning one id
  field would make the link unresolvable without also reading `RateSource`.
  (`StarRateId` is currently declared-but-unconsumed anywhere in the repo.)
- **Bug fixed in passing:** the AFC XLSX export printed neither fluctuations nor
  dayworks, so its printed lines did not sum to the AFC figure beneath them whenever
  fluctuations were non-zero.
- **Stretch deferred deliberately.** The three CVR fusion deltas (retention not broken
  out, PS movement not tracked, Value not read off the S-curve) are recorded as a new
  WP-Q ROADMAP bullet rather than implemented — they change a landed, tested feature
  and belong in their own PR.

## Gates

- `dotnet build` — **0 errors, 0 warnings** (matches the repo 0/0 baseline).
- `StingTools.Cost.Tests` — **105 passed** (17 new: build-up arithmetic, the addition
  convention incl. an explicit guard against the `×pct` misreading, zero/negative
  clamping, basis-by-unit on lines carrying both Hours and Quantity, register totals,
  attach transfer, warning behaviour). `Scheduling` 38 and `Boq` 196 also green.
- `docs/CHANGELOG.md` Phase 223 added; `docs/ROADMAP.md` PM-3 + WP-Q trued up.

**Caveat:** not exercised inside Revit — the four command tags, the capture dialog and
the panel buttons are unverified at runtime.


StingTools/Commands/Cost/DayworkCommands.cs
```diff
@@ -0,0 +1,469 @@
...
+                    .SetSubtitle($"{sheet.Description} → {targetVo.Number}")
+                    .AddSection("ATTACHED")
+                    .Metric("Sheet value", $"{sheet.Currency} {sheet.GrossTotal:N2}")
+                    .Metric("Variation", targetVo.Number)
+                    .Metric("Variation total (now)", $"{targetVo.Currency} {targetVo.TotalValue:N2}")
+                    .AddSection("FINAL ACCOUNT")
+                    .Metric("Still unattached (added separately)",
+                        $"{reg.Currency} {reg.UnattachedPricedGrossTotal:N2}")
+                    .Text("This sheet now reaches the final account through its variation and is no "
+                        + "longer added standalone — the waterfall counts it once.")
```
beckykyomugisha, line 456: **Edge case: attaching to a not-yet-agreed VO silently drops the daywork from the Final Account — and this panel note asserts the opposite.**

The double-count guard is correct *when the target VO is agreed*, but `Daywork_Attach` lets you attach to any variation regardless of status. Trace an attach to a `Draft`/`Submitted`/`Reviewed` (or later `Rejected`) VO:

- The sheet gets `VariationNumber` set → `IsAttached == true` → excluded from `DayworkRegister.UnattachedPricedGrossTotal`, so `DayworkEngine.UnattachedPricedTotal` no longer adds it.
- The Final Account only counts **agreed** VOs — `FinalAccountCommands.cs:212-213` filters `Status == Approved || Status == Incorporated`. A non-agreed VO isn't in `agreed`, so its daywork value isn't there either.

Net: the priced, instructed daywork **falls through both totals** and vanishes from the Final Account until the VO is approved. The AFC still catches it (it includes `pendingVo`), but this panel's `.Text(...)` on line 455-456 tells the user the opposite — *"now reaches the final account through its variation … the waterfall counts it once."* That's only true for an agreed VO.

It's arguably consistent with how every pending VO behaves, so this may be by-design — but a QS won't expect a signed, priced daywork to disappear from the Final Account with a message actively reassuring them it won't. Cheapest fix is to make the note honest about the VO's status (no behavioural change):

```csharp
// before building the panel:
bool voAgreed = targetVo.Status == VariationStatus.Approved
             || targetVo.Status == VariationStatus.Incorporated;
string faNote = voAgreed
    ? "This sheet now reaches the final account through its variation and is no "
      + "longer added standalone — the waterfall counts it once."
    : $"Carried by variation {targetVo.Number} (status {targetVo.Status}), which is not yet "
      + "agreed: it reaches the AFC via pending variations but will NOT appear in the Final "
      + "Account until that variation is Approved or Incorporated. It has already been removed "
      + "from the standalone daywork total.";
```

then replace the `.Text(...)` call with `.Text(faNote)`. `VariationStatus` is already in scope (`StingTools.Core.Variation`). If you'd rather prevent the situation than annotate it, gate the attach-target picker to agreed VOs only — but flexibility to attach early seems intended, so the honest note is probably the better call.

Related, lower priority (no change needed here, backlog note): once attached there's no un-attach path, and `Daywork_Price` gates on `Status != Priced`, so if the VO is later rejected or deleted the sheet stays `IsAttached` pointing at a dead VO — excluded from the daywork total permanently while its value leaves with the VO. Low-frequency orphan; worth a follow-up ticket rather than blocking this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

beckykyomugisha (commented):
Reviewed the full diff. Solid work — the double-count guard is correct by construction (attach writes `Quantity=1, UnitRate=GrossTotal` so the VO's `TotalValue` carries the sheet, and `IsAttached` removes it from the standalone total), the percentage-addition semantics are right and well-guarded by `DayworkBuildUp.Warnings()`, the `STING_BIM_MANAGER` register location is the more consistent choice, and the AFC fluctuations-not-printed fix is a real pre-existing bug. Tests exercise real behaviour, including an explicit guard against the `×pct` misreading.

One inline finding worth a decision (not a blocker): attaching a daywork to a **not-yet-agreed** variation drops it out of the Final Account while this command's own result panel tells the user it won't — details + a no-behaviour-change fix on `DayworkCommands.cs`. Plus a lower-priority orphan-on-VO-rejection note in the same thread.

Two things outside this thread:
- **The 110/112 server defaults** (your judgement call #1) are almost certainly placeholders under the addition convention (110% materials = 2.1× net). The `Warnings()` mitigation is the right move for *this* PR; the clean fix is to correct the source defaults at `Planscape.Server/.../BoqController.cs:81-83` to RICS-typical values in a separate server ticket, so both sides agree on sane numbers rather than agreeing on placeholders. Don't reopen this PR for it.
- **`mergeable_state` is `dirty`** — GitHub reports a merge conflict with `main`. Worth a rebase before this comes out of draft.

Not verified here: the 0-error/0-warning build and 105-green claims (no .NET/Revit in the review sandbox) — confirm in a real build, same standing caveat as the repo norm.

---
_Generated by [Claude Code](https://claude.ai/code)_

beckykyomugisha:
Follow-up on the review: the server daywork-percentage-defaults question (judgement call #1 in the PR description — the 110/112 additions that price materials/plant at ~2.1× net) is now tracked separately in #476, so it doesn't gate this PR.

This PR stays as-is; its `DayworkBuildUp.Warnings()` (>50% guard) is the correct interim mitigation. #476 carries the source-of-truth fix, deliberately separated because it touches the server (outside this PR's scope). One thing #476 records that's relevant here: when it lands, the server default (`BoqController.cs:81-83`) and the plugin defaults (`DayworkModels.cs:66-68`, `DayworkCommands.cs:55-57`) must change **in lockstep** — they were matched on purpose, so fixing one side alone reintroduces the disagreement.

---
_Generated by [Claude Code](https://claude.ai/code)_

StingD85:
Conflict resolved and pushed. `docs/CHANGELOG.md` — both sides appended to the head of a newest-first file, and these **are** parallel additions: `main`'s Phase 253 is the material-schedule session, this branch's Phase 223 is PM-3 dayworks. Different work, no overlapping claim, and `main` carries no Phase 223 of its own. Both kept, `main`'s first — the file is ordered by recency, not phase number (224, 226, 225 appear interleaved further down).

The resolution asserts that neither side claims a phase the other also claims, which is the thing that would make "keep both" record one phase twice.

Net against `main`: 13 files, +1473/−16, nothing deleted. Build 0/0, 1083 Boq tests pass, KUT gate 80 assertions. **All checks green.**

Left as a **draft** — marking it ready is your call, not mine.

PR #396, opened 2026-07-13
feat(placement): Family Converter — change a family's host / placement type

StingD85:
## Family Converter

A durable Placement-Centre tool that changes a Revit family's **host / placement type** (e.g. Ceiling-hosted → Face-based) and keeps it working across the project family library. Built to the self-contained spec `docs/superpowers/specs/2026-07-13-family-converter-prompt.md`.

`Family.FamilyPlacementType` is read-only in the API, so the engine picks one of two mechanisms per `(source → target)` pair:

- **P1 — checkbox toggle (lossless):** Unhosted level-based → Face-based via `BuiltInParameter.FAMILY_WORK_PLANE_BASED` on the OwnerFamily, then `LoadFamily` overwrite. Geometry/params/connectors untouched; existing instances survive.
- **P2 — template rebuild (lossy, gated):** `NewFamilyDocument` from the target host `.rft`, `CopyElements` geometry + reference planes, recreate params (shared by GUID / family by name-type) + formulas + types via `FamilyManager`, re-apply the original category, `SaveAs` under `_BIM_COORD/Families/Converted`, `LoadFamily`, best-effort instance rehost. Connectors don't survive `CopyElements` and are honestly flagged as a manual step.

### Deliverables (§11)
**New:** `StingTools/Core/Placement/FamilyHostConverter.cs` (engine + per-doc registry with corporate baseline + `<project>/_BIM_COORD/family_host_templates.json` override), `StingTools/Data/Placement/STING_FAMILY_HOST_TEMPLATES.json` (6-target matrix + path rules).
**Modified:** `StingPlacementCenter.xaml[.cs]` (new "Family Converter" tab after "Library"), `docs/CHANGELOG.md`, `CLAUDE.md` (one-line pointer).

The tab renders entirely into the shared inline Report panel via `RunInlineAction` — no pop-ups except the OS file/folder pickers. Per-row `DataGridTemplateColumn` combo (per-row `AllowedTargets`, not column-wide) with a live P1/P2 Path hint; batch runs isolate per-family failures. Reuses `StingFamilyLoadOptions`, `InstanceRehostSnapshot`, `FamilyQuickEditHelpers`, `FamilyCategoryCompatibility`, `SymbolLibraryCreator.ResolveTemplateFolder`.

### ✅ Build-verified (§12-A — I ran these)
```
dotnet build StingTools/StingTools.csproj -p:RevitApiPath="C:\Program Files\Autodesk\Revit 2025" -t:Rebuild
[...]

PR #480, opened 2026-07-21
fix(server): sane daywork default additions (115/15/15), not 110/112

### LINKED ISSUE #476: fix(server): daywork percentage-addition defaults (110/112) misprice at ~2.1× net — correct or require
## Summary
The BOQ document's daywork percentage-addition defaults are commercially implausible placeholders. Under the convention the field itself declares — NRM2/RICS **percentage additions** on net prime cost, i.e. `gross = net × (1 + pct/100)` — the materials and plant defaults price at **~2.10× / ~2.12× net**, which no real tender would carry (RICS-typical materials/plant additions are ~10–20%).

Surfaced during review of #459 (plugin dayworks capture/pricing), which mirrors these server defaults "so both sides agree numerically." The plugin's math is correct; the **values** are the defect. Right now server-default + correct-convention combine to over-price every daywork sheet on any project that never overrides the percentages.

## Evidence (all on `main`)
- Defaults set on create — `Planscape.Server/src/Planscape.API/Controllers/BoqController.cs:81-83`:
  ```csharp
  DayworkLabourPct    = req.DayworkLabourPct    ?? 115m,
  DayworkMaterialsPct = req.DayworkMaterialsPct ?? 110m,   // → 2.10× net
  DayworkPlantPct     = req.DayworkPlantPct     ?? 112m,   // → 2.12× net
  ```
- Intent is unambiguous — the entity doc-comment says so: `Planscape.Server/src/Planscape.Core/Entities/BoqDocument.cs:47`
  > `/// Daywork percentage additions per NRM2 (Labour / Materials / Plant).`

  So this is the **addition** convention; the plugin's `gross = net × (1+pct/100)` is right, and 110/112 are wrong numbers — not a convention mismatch.
- Labour 115% is defensible (absorbs statutory on-costs, supervision, profit). Materials/plant at >100% are the implausible ones.

## Blast radius (verified — small, and there is NO silent-reprice risk)
Both sides **freeze** the value, so correcting a default cannot reprice existing work:
- **Server** — the three fields are storage-only: written/echoed by `BoqController` (create `:81-83`, update passthrough `:111-113`, response echo `:49`, request record `:800-802`) and persisted by EF (`PlanscapeDbContext.cs:1791-1793`, `numeric(5,2)`). **No server calculation service consumes them** — the only Migrations hits are column definitions, not seeded data. Changing the C# default therefore affects only **new** documents (null request); existing rows keep their stored value.
- **Plugin (#459)** — each sheet stores its own percentages onto `DayworkRecord.BuildUp` at capture. Per that PR: *"Pricing uses the sheet's own build-up percentages, not today's project config, so re-running `Daywork_Price` cannot silently reprice historic sheets."* Changing the `TagConfig` fallback only seeds **newly captured** sheets.

⚠️ The real consequence is the **inverse** of a reprice: a default fix will **not** reach existing rows/sheets that already hold 110/112 — those stay mispriced until an **explicit backfill**. So the open decision is *whether to backfill existing placeholder data*, not *beware of repricing*.

## The decision this needs
NRM2/RICS daywork additions are **tender-specific** — there is no universal "correct" percentage. Two ways to stop shipping a misleading default:

- **Option A (recommended): plausible defaults.** Materials/plant → RICS-typical (~15%), labour → keep ~115% (confirm against a real tender). Keeps one-click pricing; #459's `DayworkBuildUp.Warnings()` (>50% guard) remains the backstop.
- **Option B: no magic default.** Make the three percentages required at BOQ creation (or default to 0 and block `Daywork_Price` until set), forcing the tendered figure. Safest against silent mispricing, but a larger behavioural change touching the plugin's pricing flow.

Whichever is chosen, **change server and plugin in lockstep** — they were deliberately matched, so fixing one side alone reintroduces the disagreement.

## Coupled plugin sites (land with this or immediately after #459 merges)
- `StingTools/Core/Variation/DayworkModels.cs:66-68` (property initializers)
- `StingTools/Commands/Cost/DayworkCommands.cs:55-57` (`ProjectBuildUp()` `TagConfig` fallbacks: `COST_DAYWORK_*_PCT`)

## Acceptance criteria
- [ ] Server create-defaults at `BoqController.cs:81-83` reflect the chosen option (sane additions, or required-no-default).
- [ ] Plugin defaults (`DayworkModels.cs:66-68`, `DayworkCommands.cs:55-57`) match the server in the same change or a paired PR.
- [ ] Decision recorded on whether to backfill existing rows/sheets holding 110/112.
- [ ] Real tendered figures confirmed against an actual project daywork schedule before closing (not assumed).
- [ ] `docs/ROADMAP.md` WP-Q bullet updated / closed as appropriate.

## Not this ticket
#459 stays as-is — its `Warnings()` guard is the correct interim mitigation. This is the source-of-truth fix, deliberately separated because it touches the server (outside #459's scope) and needs its own review.

_Open questions above (exact percentages, backfill scope) are for the owner to resolve — they do not block triage._

beckykyomugisha:
Closes #476.

## What
Corrects the `BoqDocument` daywork percentage-addition **create-defaults** in `BoqController.cs`:

| Field | Before | After |
|---|---|---|
| `DayworkLabourPct` | `?? 115m` | `?? 115m` (unchanged) |
| `DayworkMaterialsPct` | `?? 110m` → ~2.10× net | `?? 15m` |
| `DayworkPlantPct` | `?? 112m` → ~2.12× net | `?? 15m` |

## Why
The field's own doc-comment (`BoqDocument.cs:47`, *"Daywork percentage additions per NRM2"*) declares the RICS **addition** convention: `gross = net × (1 + pct/100)`. Under that convention the old 110/112 defaults priced materials and plant at ~2.1× net — not a figure any real tender carries (RICS-typical materials/plant additions are 10–20%). So the convention was right and the **values** were the defect. Labour 115% is genuinely typical (it absorbs statutory on-costs, supervision and profit), so it stays.

These are **fallback defaults only** — the tendered percentage is set per project via the request/override path (`:111-113`), which is untouched.

## Blast radius / migration
- The three fields are **storage-only**: written/echoed by `BoqController` and persisted by EF (`PlanscapeDbContext.cs:1791-1793`). **No server calculation consumes them**, and the migrations only *define* the columns — they never seed 110/112. So **no data migration is required**; new documents get the corrected fallback, existing rows are untouched.
- ⚠️ If a live database already holds rows written at the old 110/112 default, they stay at those values until an explicit one-line `UPDATE` — flagged for the owner (couldn't be checked from the build sandbox).

## Lockstep with PR #459
The plugin mirrors these defaults on purpose (so both sides agree numerically). The plugin-side correction to the same **115/15/15** lives in the files #459 introduces (`DayworkModels.cs:66-68`, `DayworkCommands.cs:55-57`) and rides that PR — it can't be made on `main` because those files don't exist here yet. Both PRs land on the same numbers.

## Not verified
Built without `dotnet` (no .NET SDK in the sandbox) — the change is a two-token literal edit plus a comment, but the server CI gate (`planscape-server.yml`) should confirm on this PR. Same standing caveat as the repo norm.

---
_Generated by [Claude Code](https://claude.ai/code/session_0119mCvUHrqv6u78ki1WATqb)_


StingD85:
Updated onto current `main` (was 672 commits behind). **All 13 checks now pass** — the two failures dated 2026-07-21 were a statement about the branch's age, not its correctness, the same pattern as #396.

I verified the change itself rather than only the CI result. `BoqDocument` documents these as *"Daywork percentage **additions** per NRM2"*, so the previous `110`/`112` meant gross = 2.10× and 2.12× net prime cost for materials and plant. The new `15`/`15` is RICS-typical (10–20%), and labour staying at `115` is a normal daywork labour addition absorbing statutory on-costs, supervision and profit. The values are carried onto an issued tender document and are not consumed by any server-side calculation, so the defect was a wrong number on paper rather than a wrong total.

Net effect on `main`: **one file, +7/−2, nothing deleted** (measured with `merge-tree`, not `git diff main..branch`).

Left as a **draft** — marking it ready is the author's call, not mine.

PR #504, opened 2026-07-30
Deploy-readiness: Render Postgres URL parsing, health-check 403 fix, server hardening (448 tests green)

StingD85:
**Deploy-readiness for Planscape Server on Render.** 19 commits: the two fixes without which a first Render deploy fails silently, a batch of server security/correctness fixes surfaced while making the code actually run against real Postgres, and a test suite brought to green (448 passing, 0 skips, real-PostgreSQL harness).

## The two deploy-blocking fixes

- **`23c15c37c` — Postgres provider-URL parsing + pool caps.** Render (like Heroku/Fly/Railway/Neon) injects the connection string as `postgresql://user:pass@host/db`. Npgsql 8 only accepts the `Host=…;Port=…` keyword form, so **the API could not connect to Render Postgres at all** without this. Also caps pools (Npgsql defaults to 100/process; Render starter Postgres tops out ~97) — api 20+10, worker 15+15.
- **`b341d2acc` — health check pointed at an endpoint that 403s in production.** `render.yaml` had `healthCheckPath: /health`, the full diagnostic, which S11 hardening gates behind a private-network caller + `X-Health-Token`. Render's prober can send neither, so every probe → 403 → Render marks a healthy service unhealthy → **deploy never goes live.** Repointed to `/health/live` (AllowAnonymous, `{status:"alive"}`).

## Server correctness/security (found while making it run for real)

- `d39e065cc` — stop remapping JWT claims; **25 role checks were dead**
- `0d9a525ab` — three anonymous endpoints (incl. signup) blocked by the tenant filter → duplicate-email registration and licence activation unblocked
- `d9532fc6a` / `003821c27` / `838d44360` / `b6b9f6ce7` — tenant-context wiring: fast-path answered before the tenant check; keyword overrides unreachable; unset Plan 402'd every write
- `c7a414b60` — admin project-visibility bypass never fired
- `a1f9ab55b` — JSON cycles 500'd responses
- `e951a3b02` — rate limiter ran before auth (degraded to per-IP); `457c82e2c` — rate-limit licence activation
- `1129ac29b` — planscape-web access-token screen, so a subscriber never needs curl
- `9afce5834` — corrected Render tier guidance + cost (~$54/mo all-starter across 7 services, not the retired £12/2-service figure)

## Diff shape

233 files, +5,356 / −19,322 — the large deletion count is mostly test-fixture and generated-artefact cleanup; the substantive change is concentrated in `Planscape.Server/`.

## Do not merge blind

This is the hard prerequisite for the Render deploy, but it is 19 commits of server change. Verify `dotnet build Planscape.sln -c Release` is clean and the test suite runs green before merging. Once merged, `render.yaml` on `main` is deploy-correct.


StingD85:
## Branch updated from `main` + re-verified (deploy-readiness gate)

The branch was **139 commits behind `main`**, which is why the diff appeared to remove ~19k lines — a stale-merge-base artifact, not an actual revert. Merged `origin/main` in; **no conflicts**.

Post-merge state: **0 behind / 19 ahead**, diff vs main **+2,487 / −141 across 38 files**, all server/deploy-scoped. No other session's work is reverted.

### Verification (run on Windows + .NET 8, this machine)

| Check | Result |
|---|---|
| `git merge origin/main` | clean, 0 conflicts |
| `dotnet build Planscape.sln -c Release` | **0 errors**, 13 pre-existing warnings (NU1603 / CS0618 / xUnit2009) |
| `dotnet test Planscape.sln -c Release` (real PostgreSQL, `PLANSCAPE_TEST_PG`) | **458 passed, 0 failed, 0 skipped** |

### Acceptance criteria after merge

- `HandoffExchange` present in `AuthController.cs`
- `render.yaml`: API `healthCheckPath: /health/live`
- `render.yaml`: `PLANSCAPE_USE_ENSURE_CREATED` + `PLANSCAPE_HANDOFF_SECRET` (on both api and worker)
- `render.yaml`: `plan: basic-256mb` on `planscape-db`

### Why both fixes are still required

Neither fix commit is on `main`; `main` has **no** Postgres URL parsing anywhere and still carries `healthCheckPath: /health`.

The health gate is real — `Program.cs:1399` resolves to `isPrivate && tokenOk` outside Development. Render's prober is external and sends no `X-Health-Token`, so it would get the 403 at `Program.cs:1403` and the service would never go live. (Probing `/health` on the local dev stack returns 200 only because it runs in Development, where the private-caller gate alone suffices — it does not disprove the production 403.)

`PgConnectionStringsTests` covers the exact Render form `postgresql://user:pass@host:5432/db`, plus percent-encoded credentials, omitted port, and `sslmode` passthrough.

StingD85:
## CI "Build & Test" is red — analysis: pre-existing, not from this branch

The `Planscape Server — Build, Test & Deploy` check fails. It is **not** a regression from this PR or from the `main` merge. Evidence:

**1. Identical failures before and after the merge.** The run on `b341d2acc` (pre-merge) and the run on the merge commit both produced **exactly the same 33 failures**. The merge introduced nothing.

**2. The 5 "new vs baseline" failures also fail on `main`.** `tests/check-new-failures.sh` compares against `tests/known-failing-tests.txt` (66 entries). Five CI failures are not in it:

```
CoreApiTests.Issues_OtherTenant_Returns404
CoreApiTests.TenantIsolation_OtherTenantCannotAccessDocuments
CoreApiTests.TenantIsolation_OtherTenantCannotAccessMeetings
CoreApiTests.TenantIsolation_OtherTenantCannotAccessMim
ProjectVisibilityTests.DeepLink_NonMember_GetsNotFound_NotForbidden
```

All five **also fail in a `main` CI run** (run 29755700273, 71 failures). They are repo-wide, and the baseline file is stale relative to what CI actually produces under load — this is the order/parallelism flake class the script's own hint points at (ROADMAP DEP-7).

**3. They pass locally, every way I ran them.**

| Local run (Windows, .NET 8) | Result |
|---|---|
| Full suite, real PostgreSQL (`PLANSCAPE_TEST_PG`) | 458 passed / **0 failed** |
| Full suite, no PG (CI's config) | **0 failures** |
| The 5 tests in isolation, no PG | 5 passed / 0 failed |

**4. The hard stop is the gate script, not the assertions.** The step fails at `check-new-failures.sh` line 22 — `no test summary in output — the run did not complete`. It requires a line matching `^(Passed!|Failed!)`. When VSTest fails at the MSBuild level (`MSB4181: The "VSTestTask" task returned false but did not log an error`) that line is never emitted, so the script cannot even reach the baseline comparison.

The run also took **30.9 minutes** with repeated `Hangfire.InMemory` `ObjectDisposedException` retry storms and individual tests taking 1m24s — the runner is resource-starved, which is the likely driver of the parallelism flakes.

### What I did not do

I did not add these five to `known-failing-tests.txt`. Four of them are tenant-isolation assertions — a security boundary. Silencing them to make a deploy PR go green would be exactly the wrong trade, and they demonstrably pass when the suite is not starved. They belong in a separate CI-health fix, not here.

`main` has no branch protection, so this check does not mechanically block the merge — but the decision is the owner's, on the evidence above.

PR #522, opened 2026-08-01
fix(viewer): stop the boot overlay hanging forever on "Loading model 0%"

StingD85:
## Summary

Fixes the reported bug where the 3D model sits on **"Loading model 0%"** indefinitely — no error, no Retry.

`planscape-web` builds the iframe src with only `project`/`token`/`tenant`/`user` ([page.tsx:98](../blob/main/planscape-web/app/projects/%5Bid%5D/viewer/page.tsx#L98)) and drives the model over `postMessage` instead. So the viewer **always** took its "no model to load" branch, which called `markModelReady()` but never hid `#bootLoader` — and that overlay is dismissed only by a *successful* load.

Any hiccup in the host's load path therefore left a permanent spinner: manifest throw, empty model list, `ready` handshake never firing, or a failed fetch. The iframe's own error surfaces (error text, Retry) were wired only to the `?model=` path the web app never uses.

### Reproduced on the live deployment, with no credentials

```
viewer.html?project=<id>&embed=1
-> {"message":"Loading model","progress":"0%","retryBtnPresent":false,"modelReady":true}
```

Unchanged 63 seconds later — past the 60s download timeout, which never fires because no download was ever started. `modelReady: true` while showing 0% is the signature: the engine considered the model resolved; only the overlay was orphaned.

## What changed

- **Watchdog** in that branch: if no geometry is on screen after 45s, surface an actionable error with Retry.
- **New `sting:modelLoadRequested` event** so the watchdog distinguishes "the host never asked us to load anything" from "a load is genuinely in flight".
- **Failed host-driven loads** now stop the spinner and offer Retry, instead of spinning under "Failed to load model".
- **Pluggable Retry action** — the host-driven path has no GLB to re-fetch, so it reloads.

## Test plan

Verified in a browser against a production-like bundle (`wwwroot` + the Docker `dist/` overlay, i.e. the same minified artefact production serves):

- [x] Host never asks → `"No model was loaded for this view — it may not be published yet."` + Retry, spinner stopped
- [x] Host asks, load fails → `"Failed to load the model file."` + Retry, spinner stopped
- [x] Before 45s → still shows the normal spinner (no premature error)
- [x] Successful loads untouched — the watchdog returns early when geometry exists or the overlay is already hidden
- [x] `node --check` passes; `npm run build` (the exact Docker step) succeeds
- [x] Source ↔ wwwroot byte-equal gate simulated locally — all 7 files in sync

### Not covered

This does **not** address the separate `Couldn't join — retry` A/V bug. That is `room.connect()` to LiveKit failing and is unrelated to the model load — A/V join is not gated on `modelReady`.

### DIFF
--- a/Planscape.Server/src/Planscape.API/wwwroot/coordination-viewer.js
+++ b/Planscape.Server/src/Planscape.API/wwwroot/coordination-viewer.js
@@ -374,6 +374,11 @@
       // with a Retry affordance. The boot overlay lives in viewer.html; this
       // script runs in the same document so it drives #bootLoader directly.
 
+      // How long to wait for the HOST to drive a model onto the screen before
+      // giving up and showing an actionable error. Generous: a large GLB on a
+      // cold free-tier backend can legitimately take a while.
+      const HOST_LOAD_WATCHDOG_MS = 45000;
+
       function bootLoaderEl() { return document.getElementById('bootLoader'); }
       function setBootProgress(pct, label) {
         const elp = document.getElementById('loadingProgress');
@@ -393,7 +398,11 @@
         setBootMessage('Loading model');
         setBootProgress(0, null);
       }
-      function showBootError(msg, canRetry) {
+      // The Retry action is pluggable: the URL-param path retries the GLB
+      // fetch, but the host-driven path (no ?model= — the web app posts a
+      // 'load' command instead) has nothing to re-fetch, so it reloads.
+      let bootRetryAction = () => { resetBootLoader(); loadModelGlb(); };
+      function showBootError(msg, canRetry, onRetry) {
         const bl = bootLoaderEl();
         toast(msg, 'error');                       // keep the toast too
         if (!bl) return;
@@ -401,6 +410,7 @@
         const sp = bl.querySelector('.spinner'); if (sp) sp.style.display = 'none';
         setBootMessage(msg);
         setBootProgress(null, '');
+        if (onRetry) bootRetryAction = onRetry;
         let retry = bl.querySelector('#bootRetryBtn');
         if (canRetry) {
           if (!retry) {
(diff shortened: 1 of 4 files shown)


StingD85:
## Root cause found — this PR now fixes the trigger, not just the symptom

The reporter confirmed the trigger is **starting a live meeting**. Investigating that specific path found the actual cause, and it invalidates one of the "already ruled out" items.

`startMeeting()` / `joinMeeting()` hard-navigate the viewer document so `meeting-sync.js` activates:

```js
// Reload this tab INTO the meeting so meeting-sync.js activates.
location.href = meetingJoinUrl(id);
```

`meetingJoinUrl` re-adds `?model=` from `modelId` — but `modelId` is the `?model=` URL param, which **planscape-web never sets** (it drives the model over `postMessage`). So the meeting URL carried `project` but silently dropped the model. After the re-navigation nothing brought it back: the viewer booted into its model-less branch, and the host does not re-post `load` because its `ready` state doesn't change value, so the effect never re-runs.

That is precisely "starting a live meeting reloads the model and it never finishes loading".

### This invalidates "the iframe does not remount"

That conclusion came from a Playwright spec that never successfully drove the meeting UI (`Join control not found after opening the menu`), so it never exercised this path. The document *does* re-navigate — just only when a meeting actually starts.

### Change

Records the model the host actually loaded (from the `load`/`addModel` command, now carried on `sting:modelLoadRequested`) and falls back to it when building the meeting URL, so the re-navigated document reloads the same model itself.

Verified in a browser on a production-like bundle:
- host `load` with modelId → id captured (`MODEL-ABC-123`)
- post-reload URL with `?model=` → viewer performs its **own** load instead of sitting in the frozen model-less branch

The watchdog from the first commit stays as defence-in-depth: it turns any *other* stall into an actionable error instead of an infinite spinner.

PR #538, opened 2026-08-01
docs: document/deliverable gap table with verdicts (+ seat-inversion tests)

StingD85:
## Summary

Items 8 / 12 / 13 — **report + tests, no implementation.** Awaiting review of the verdict column before anything is built.

Adds `docs/DOCUMENT_LIFECYCLE_ALIGNMENT_AUDIT.md` (gap table with a verdict per row) and `QuotaSeatCountingTests` (proving the billing seat inversion).

## ⚠️ Two corrections to my earlier audit

Both came from re-measuring, not re-reasoning. **The earlier version of this PR overstated the problem in two places.**

**1. Deliverables and stage gates are NOT dead data.** I claimed seven controllers had no client on either side. I had not actually grepped the mobile endpoint module. `Planscape/src/api/endpoints.ts` drives deliverables (`:1456,1465,1478,1497,1508`), stage gates (`:1407,1411,1420`) and document approvals (`:1005,1012`). **The real orphan count is five**, and "approval chains are entirely unreachable" was half wrong — there are *two* approval mechanisms and mobile uses one of them.

**2. "Nothing ever writes `PM`" was wrong.** `ProjectRole` is client-writable and unvalidated (`ProjectMembersController.cs:418`). The actual defect is sharper:

- `GET /members/roles` (`:548-569`) returns the **ISO 19650** vocabulary — which contains `PM` — and that feeds the **`Iso19650Role`** column (`planscape-web/lib/data.ts:428`, saved at `members/page.tsx:119`).
- The web members grid offers a **different** six-value `ProjectRole` list: `['Viewer','Contributor','Coordinator','Manager','Owner','Admin']` — **no `PM`** (`members/page.tsx:23`).

So twelve gates compare `ProjectRole` against a value that only ever appears in the *other* field's vocabulary. **"Manager" and "PM" are the same concept spelled twice**, and those gates are closed to everyone the UI can produce.

## Evidence standard — no row says "delete"

"No caller found" covers `StingTools/`, `planscape-web/`, `Planscape/` (mobile) and `StingBridge/` only. It does not cover external integrations — and `BcfApiController` / `OpenCdeController` implement *published interop standards*, so third parties are their expected consumers. Verdicts are **wire**, **document**, **rename** or **fix**. Deleting anything needs server access logs over a representative window.

### L1 — role vocabulary

The doc enumerates **every** role string the server compares against, with sites, so a canonical set can be chosen without fixing one gate in isolation:

| Value | Where |
|---|---|
| `"PM"` x12 | DistributionGroups:202, PhotoAlbums:395, PhotoChecklists:320, PhotoPolicy:103, PhotoShare:122, SavedViews:136, SitePhotos:874,884, SitePhotosExt:610, BackgroundJobs:314, DailyPhotoDigest:157, P6LiveLink:288 |
| `"Manager"` / `"Admin"` / `"Owner"` / `"Coordinator"` | WeeklyDigestJob:48-49, CdeContainers:255, BackgroundJobs:278, CoordinatorWorkload:39 |
| `"BimManager"` / `"ClientGuest"` / `"Author"` | CoordinatorWorkload:39, DailyPhotoDigest:126, TenantAdmin:51, QuotaGuard:76,77 |

**Written** by first-party paths: `Owner`, `Manager`, `Contributor`, `Coordinator`. **Never written:** `PM`, `Author`, `BimManager`, `ClientGuest`. A **migration is required** — existing rows carry the first set.

### L2 — seat inversion, now proven by test

`QuotaSeatCountingTests` builds a realistic Studio-plan roster (Owner + Manager + 3 Contributors):

| axis | current | max | result |
|---|---|---|---|
| Authors | **0** | 1 | allowed — the paid seat is never consumed |
| Coordinators | **5** | 5 | **DENIED** — owner and manager counted as coordinators |

So the sixth person is refused while the author seat sits unused.

**Per item 13 — is the cap even enforced?** Partly. `[Quota]` is applied to the Projects axis only (`ProjectsController.cs:116`). `CheckCanAddUserAsync` has exactly two production callers (`OnboardingController.cs:93`, `TenantAdminController.cs:108`); the **main project-member invite path bypasses the quota guard entirely** and enforces `tenant.MaxUsers` instead (`ProjectMembersController.cs:256`). The mis-count is real; its blast radius is those two paths. That is recorded in the test's own doc comment so it can't be over-claimed later.

### A trap the tests caught

The first version of the fixture **measured nothing**. `PlanscapeDbContext`'s global filter is `TenantId == CurrentTenantId`, which falls back to `Guid.Empty` without an `ITenantContext` — matching no rows. Two tests "passed" against an empty result set. The fixture now wires the three-arg ctor, and the comment explains why, so the next person doesn't repeat it.

## Test results

- `QuotaSeatCountingTests` — **4/4 pass**, measuring real rows.
- Related subset (Quota / Tenant / Role / Handoff) — **220 passed, 0 failed**.
- Full suite — **493 passed, 0 failed, 7 skipped**.

## Status

Docs + tests only. No production code changed. **Nothing on the table is started** — the verdict column needs review first, and L2 in particular needs a product decision (*which* role consumes an author seat) before it can be fixed rather than merely re-inverted.


StingD85:
## Heads-up: #593 corrects the seat split — one test here needs rewriting

#593 fixes a defect **upstream** of the inversion this PR characterises, so the two interact. Flagging rather than touching this branch.

### What #593 found

The inversion is real (verified — I ran this PR's 4 tests against current main: **4 passed**). But there is a bigger problem above it: **the meter counts a table the metered operation never writes.**

`POST /api/tenant/invite` gates on these axes and returns 402 `quota_exceeded`, then creates an `AppUser` and nothing else — no `ProjectMember`. Both readers of the axes (`QuotaGuardService.CountAsync`, and `TenantAdminController.Dashboard` which inlines the split a second time) counted `ProjectMembers`. So an accepted invite could never move the number it had just been checked against, while that number was driven by `ProjectMembersController`, which never consults the guard.

Proven end-to-end on the billing surface: inviting a coordinator left `usage.coordinators.current` at **0** (Expected 1, Actual 0), with a control test confirming the user really was persisted.

So #593 counts seats over `AppUser`, with **author = `Iso19650Role == "A"`** — what both invite paths already persist — and coordinators derived as `total − authors` (a NULL role would fall off both sides of a `!= "A"` predicate and give away a seat).

### The interaction

I ran this PR's tests against #593: **Failed 1, Passed 3.**

| test | vs #593 |
|---|---|
| `Authors_axis_counts_zero_for_a_realistic_roster` | still passes |
| `Coordinators_axis_counts_every_member_including_the_owner` | still passes |
| `Seats_are_counted_per_tenant` | still passes |
| `A_member_explicitly_marked_Author_moves_between_the_two_axes` | **FAILS** |

Smaller than this PR's header predicted — the first two survive because the fixture's users carry no `"A"` ISO role, so authors still reads 0 and coordinators still reads 5. Only the fourth breaks, because it asserts `ProjectRole = "Author"` moves a seat, and that is exactly the semantics #593 retires: `"Author"` is not in `ProjectRoles.All`, and the only `ProjectRole` values first-party code writes are Owner, Manager, Contributor, Coordinator.

This PR's own header already anticipates this ("expected to FAIL — that failure is the signal the fix landed, and they should be rewritten"). **Whichever of the two merges second should rewrite that one test** to assert the ISO-role split.

The characterisation work here is what made #593 quick to scope — the axis-level pinning was already done, so #593 only had to add the end-to-end gate-vs-meter proof.

PR #540, opened 2026-08-01
feat(server): one capability layer to replace eleven dead "PM" gates

StingD85:
## The bug, confirmed independently

The brief asked me to confirm the wrong-field reading before building on it. It is exactly right, and the two fields document themselves:

```csharp
/// <summary>Application role within this project (Viewer/Contributor/Coordinator/Manager).</summary>
public string ProjectRole  { get; set; } = "Contributor";      // ProjectMember.cs:15

/// <summary>ISO 19650-2 role code for this project (A/PM/BC/AR/SE/ME/QS etc.).</summary>
public string Iso19650Role { get; set; } = "M";                // ProjectMember.cs:18
```

`"PM"` appears only in the ISO list — which is also what `GET api/projects/{id}/members/roles` returns (`ProjectMembersController.cs:548-569`) and what the web grid saves into `iso19650Role` (`members/page.tsx:119`). Meanwhile the web's `ProjectRole` dropdown is a different six-value list with no `PM` in it (`members/page.tsx:23`).

So eleven sites read the right code off the wrong column. **A wrong-field bug, not a data migration.**

What was closed as a result: album curation, checklists, share links, photo policy, photo approve/reject, distribution groups, saved-view deletion, the 24h PM escalation, the approver digest email, and P6 milestone pushes — all unreachable for anyone who was not a tenant Admin/Owner.

## What this adds

`Planscape.Core/Entities/ProjectRoles.cs` — canonical vocabulary + two capabilities, split by blast radius:

| | grants | to |
|---|---|---|
| **CurateProject** | albums, checklists, distribution groups, deleting another member's saved view | ProjectRole in {Manager, Owner, Admin, Coordinator} or Iso19650Role in {PM, A, BC} |
| **ApproveSitePhotos** | approve/reject, include-originals, share links, PUT photo-policy | ProjectRole in {Manager, Owner, Admin} or Iso19650Role in {PM, A} |

A Coordinator curates but does **not** approve — approval releases imagery outside the project.

**Two shapes, on purpose.** Controllers get a claims-aware async helper beside `RequireProjectMemberAsync`. Jobs get a static `Expression<Func<ProjectMember,bool>>` so EF translates it *inside* `.Where()`. `DailyPhotoDigestJob` uses query syntax, which cannot take an Expression in its `where`, so the predicate is applied to the **source** — otherwise it would client-evaluate the table per digest.

Each controller helper keeps its name and signature; only the body changed. **All 36 call sites are untouched.**

## Write validation

`PUT members/{id}` accepted **any** string for `ProjectRole` (`:418`) — which is how the vocabulary drifted this far in the first place. Non-canonical values now return `400 { error: "invalid_project_role", allowed: [...] }`, matching the existing `invalid_kind` shape in `DistributionGroupsController`.

`Iso19650Role` is deliberately **not** validated here — that is a wider change.

## The widening claim — one correction to the brief

The brief said to state this change is purely widening, and to stop and report if any gate would narrow. **One would have**, so I handled it rather than shipping the narrowing:

`ProjectRole` was **unvalidated until this commit**, so a row with `ProjectRole = "PM"` is possible even though no UI writes it — and such a row **passes the current gate**. Dropping `"PM"` from the accepted ProjectRole set would have revoked access from exactly those rows.

So `"PM"` is retained as an explicit, commented legacy tolerance in the capability check, while being deliberately **absent** from `ProjectRoles.All` so new writes are refused. A test pins it.

With that in place, widening is verified rather than asserted:

| | before | after |
|---|---|---|
| tenant branch | `role is "Admin" or "Owner"` | same **plus** `IsInRole("Admin"/"Owner")` — wider |
| user-id claims | `user_id` then `NameIdentifier` then `sub` | **identical** chain (checked in all 4 controllers) |
| role match | `ProjectRole == "PM"` | superset, incl. legacy `"PM"` |

No user can lose access at any of the eleven sites.

## Tests — RED then GREEN

Proved RED by temporarily restoring the pre-fix semantics:

```
Capability_matrix(Manager,     M,  curate: True, approve: True)  [FAIL]
Capability_matrix(Owner,       M,  ...)                          [FAIL]
Capability_matrix(Admin,       M,  ...)                          [FAIL]
Capability_matrix(Coordinator, M,  curate: True, approve: False) [FAIL]
Capability_matrix(Contributor, PM, ...)                          [FAIL]   <- the wrong-column bug
Capability_matrix(Contributor, A,  ...)                          [FAIL]
Capability_matrix(Contributor, BC, ...)                          [FAIL]
CurateProject_predicate_selects_the_right_members_from_the_database [FAIL]
ApproveSitePhotos_predicate_is_narrower_than_curate                 [FAIL]

Failed: 9, Passed: 21, Total: 30
```

Then GREEN with the real implementation: **30/30 passed.**

The two that matter most: `Manager` is a *real* member passing a previously-dead gate, and `Contributor` + `Iso19650Role=PM` proves the wrong-column read is fixed.

**Fixture-trap guards** (the brief flagged this, and it is real):
- every context is built with an `ITenantContext` — the parameterless ctor makes `CurrentTenantId` fall back to `Guid.Empty`, matching **no rows**, so assertions would pass against an empty set;
- `Sanity_the_fixture_actually_has_rows` asserts non-empty **before** any capability claim;
- `CurateProject_predicate_translates_to_SQL` asserts the generated SQL contains the role columns — a predicate that client-evaluates would otherwise pass this suite while pulling the whole table in prod;
- `Expression_and_in_memory_forms_agree_on_every_seeded_row` pins the two encodings together, which no reader can verify by eye.

## Parity gate

Extended the existing `contract-drift` mechanism rather than inventing one — same style as `test_ifc_element_to_wire_keys_match_dto`: parse both sources, no running server, no generated schema.

- `test_web_project_roles_are_all_accepted_by_the_server` — every value in the web `PROJECT_ROLES` must be in `ProjectRoles.All`. Subset, not equality, because the server also accepts `ClientGuest` (read by `DailyPhotoDigestJob.cs:126`) which the web deliberately does not offer.
- `test_server_project_role_vocabulary_parses` — guards the parser itself, and asserts `"PM"` is **never** canonical. A silent empty parse would make the parity test pass vacuously, which is the exact failure mode these gates suffer from.

Verified the parsers return real data:
```
server: ['Viewer','Contributor','Coordinator','Manager','Owner','Admin','ClientGuest']
web   : ['Viewer','Contributor','Coordinator','Manager','Owner','Admin']
```

Also added `planscape-web/**` to the workflow triggers — it was missing, so a change to the web dropdown would not have run this gate.

## Post-change grep

Zero `ProjectRole == "PM"` in executable code. Survivors are exactly the non-PM ones the brief said to leave: `CdeContainers:255`, `WeeklyDigestJob:48`, `TenantAdmin:51`, `CoordinatorWorkloadJob:39`, `QuotaGuardService:76`, plus `BackgroundJobs:278` (Owner/Admin) and `DailyPhotoDigestJob:126` (ClientGuest).

## STOP-and-ask — the capabilities endpoint

Not built, as instructed. Proposing it:

**`GET api/projects/{projectId}/members/capabilities` returning `{ curateProject: bool, approveSitePhotos: bool }`**

The three clients currently have no way to learn what the caller may do, so each must attempt-then-report a 403. That works but means every surface shows an action, lets the user click it, and only then explains — and each surface has to render that 403 consistently for parity to hold.

One endpoint would let all three disable or hide the action identically, which is what parity rule (c) actually asks for. It reads only data the caller already has access to, so it leaks nothing.

**Until you decide**, the parity rule for Phases 2-5 stands as written: attempt the action, render the server's 403 as a specific human message ("Only a project manager can curate albums") on all three surfaces. Never a silent no-op, never a toast claiming success.

## Separate finding — NOT touched, as instructed

`QuotaGuardService.cs:76-77` counts the billing **Authors** seat axis as `ProjectRole == "Author"`, and nothing in the codebase ever assigns `"Author"`, so that axis reads 0 forever; `Coordinators` counts `!= "Author"`, i.e. every member including the project owner. Left completely alone per the brief. A characterisation test pinning the current behaviour already exists in #538 (`QuotaSeatCountingTests`) if you want the numbers.

Worth noting one interaction: `"Author"` is **not** in the canonical vocabulary this PR introduces, so `PUT members/{id}` will now reject it. Nothing wrote it before and no UI offers it, so this changes no behaviour — but if the seat fix later wants `"Author"` as a real role, it must be added to `ProjectRoles.All` at that point.

## Verification status

- `dotnet build Planscape.API` -> **succeeded, 0 errors** (7 pre-existing warnings).
- `ProjectRoleCapabilityTests` -> **30/30**, RED-then-GREEN evidence above.
- Python parity tests -> **2/2**.
- Full server suite -> reported in a follow-up comment; the run exceeds the tool timeout.
- **Not verified:** no endpoint was exercised against a running server in this phase, because Phase 1 changes no routes or response shapes — only who passes an existing gate. Phase 2 onward will do live exercise as the method requires.

## Notes

Known-broken CI: "Build & Test / CI Gate" fails on a Postgres container (`role "root" does not exist`) — pre-existing on `main`, unrelated.

### DIFF
--- a/.github/workflows/contract-drift.yml
+++ b/.github/workflows/contract-drift.yml
@@ -40,6 +40,7 @@ on:
       - 'stingtools-core/python/**'
       - 'StingBridge/**'
       - 'Planscape/**'
+      - 'planscape-web/**'
       - '.github/workflows/contract-drift.yml'
   pull_request:
     paths:
@@ -49,6 +50,7 @@ on:
       - 'stingtools-core/python/**'
       - 'StingBridge/**'
       - 'Planscape/**'
+      - 'planscape-web/**'
       - '.github/workflows/contract-drift.yml'
   workflow_dispatch:
 
--- a/Planscape.Server/src/Planscape.API/Controllers/DistributionGroupsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/DistributionGroupsController.cs
@@ -191,16 +191,11 @@ public class DistributionGroupsController : ControllerBase
         return NoContent();
     }
 
-    private async Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
+    // code, never a ProjectRole (ProjectMember.cs:15 vs :18), so this gate
+    // matched essentially nobody. Delegates to the shared capability layer.
+    private Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
+        => this.CanCurateProjectAsync(_db, projectId, ct);
     private Guid? CurrentUserIdOrNull()
     {
         var s = User.FindFirst("user_id")?.Value
--- a/Planscape.Server/src/Planscape.API/Controllers/PhotoAlbumsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/PhotoAlbumsController.cs
@@ -384,16 +384,11 @@ public class PhotoAlbumsController : ControllerBase
     /// Mutation gate (matches SitePhotosController.IsApproverAsync):
     /// tenant Admin / Owner OR project PM.
     /// </summary>
-    private async Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
+    // code, never a ProjectRole (ProjectMember.cs:15 vs :18), so this gate
+    // matched essentially nobody. Delegates to the shared capability layer.
+    private Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
+        => this.CanCurateProjectAsync(_db, projectId, ct);
 
     /// <summary>
     /// Read gate: project member always sees Internal/Members; Client
--- a/Planscape.Server/src/Planscape.API/Controllers/PhotoChecklistsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/PhotoChecklistsController.cs
@@ -309,16 +309,11 @@ public class PhotoChecklistsController : ControllerBase
         return Ok(c);
     }
 
-    private async Task<bool> IsAuthorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
+    // code, never a ProjectRole (ProjectMember.cs:15 vs :18), so this gate
+    // matched essentially nobody. Delegates to the shared capability layer.
+    private Task<bool> IsAuthorAsync(Guid projectId, CancellationToken ct)
+        => this.CanCurateProjectAsync(_db, projectId, ct);
     private Guid? CurrentUserIdOrNull()
     {
         var s = User.FindFirst("user_id")?.Value
--- a/Planscape.Server/src/Planscape.API/Controllers/PhotoPolicyController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/PhotoPolicyController.cs
@@ -92,16 +92,11 @@ public class PhotoPolicyController : ControllerBase
         return Ok(pol);
     }
 
-    private async Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
+    // code, never a ProjectRole (ProjectMember.cs:15 vs :18), so this gate
+    // matched essentially nobody. Delegates to the shared capability layer.
+    private Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
+        => this.CanApproveSitePhotosAsync(_db, projectId, ct);
     private Guid? CurrentUserIdOrNull()
     {
         var s = User.FindFirst("user_id")?.Value
--- a/Planscape.Server/src/Planscape.API/Controllers/PhotoShareController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/PhotoShareController.cs
@@ -111,16 +111,11 @@ public class PhotoShareLinkAdminController : ControllerBase
         return Ok(link);
     }
 
-    private async Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
+    // code, never a ProjectRole (ProjectMember.cs:15 vs :18), so this gate
+    // matched essentially nobody. Delegates to the shared capability layer.
+    private Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
+        => this.CanApproveSitePhotosAsync(_db, projectId, ct);
     private Guid? CurrentUserIdOrNull()
     {
         var s = User.FindFirst("user_id")?.Value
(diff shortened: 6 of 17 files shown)


StingD85:
## Full-suite result (the outstanding item)

Run on `claude/capability-layer` @ `19e1737ce`:

```
Passed!  -  Failed: 0, Passed: 561, Skipped: 9, Total: 570
EXITCODE=0
```

Reproduced twice, identical both times. No `[FAIL]` markers in either log.

**Confirmed the capability tests actually ran** rather than assuming it — `dotnet test` only names failed and skipped tests, so their absence from the log proves nothing:

| run | total |
|---|---|
| full suite | 570 |
| `--filter FullyQualifiedName!~ProjectRoleCapabilityTests` | 540 |

570 − 540 = **30**, matching `ProjectRoleCapabilityTests` exactly.

The 9 skips are pre-existing `PostgresSequenceCounterTests` — they need a real Postgres, which is the same gap blocking the Phase 6 runtime pass.

### One thing worth flagging: the suite is flaky

On the *first* filtered run I got:

```
Failed!  -  Failed: 17, Passed: 514, Skipped: 9, Total: 540
```

An identical re-run of the same command on the same commit gave `Failed: 0, Passed: 531, Skipped: 9, Total: 540`. Same total, so it is 17 flaky tests, not a real regression — and the full suite passed 0-failed both times either side of it.

I did not capture the failing names (the run that produced them wasn't written to a file, and it hasn't recurred in three subsequent runs). So I can't tell you *which* 17. Reporting it rather than quietly discarding it, because a suite that fails 17 tests once in five runs will eventually be blamed on whichever PR happens to be in flight.

The likely cause is pre-existing and already documented in this repo — `HandoffProvisioningSqliteTests`' header notes a process-wide Hangfire teardown that makes extra `WebApplicationFactory` instances unreliable in this assembly (ROADMAP DEP-7). That is a separate issue from this PR and I have not touched it.

**Nothing here changes the Phase 1 verdict:** build clean, RED→GREEN on the capability suite (9 failures against restored pre-fix semantics → 30/30), and the full suite green.

StingD85:
## Independent verification pass — the four open items closed, including the full-suite run this PR deferred

### 1. Full server suite — the item the body left as "reported in a follow-up comment"

```
Passed!  - Failed: 0, Passed: 561, Skipped: 9, Total: 570, Duration: 40 s
```
`dotnet test tests/Planscape.Tests` on `19e1737ce`. The 9 skips are the pre-existing Postgres-only /
search tests, unrelated to this change. **Nothing in the suite regressed.**

`ProjectRoleCapabilityTests` re-run in isolation: **30/30**.

### 2. SQL translation — the actual generated SQL, not `Assert.Contains`

The suite asserts the query string *contains* the role columns, which is a proxy. Here is the real
output of `ToQueryString()` against SQLite:

```sql
-- ProjectRoles.CanCurateProject
SELECT "p"."Id", … FROM "ProjectMembers" AS "p"
WHERE (@__ef_filter__BypassTenantFilter_0 OR "p"."TenantId" = @__ef_filter__CurrentTenantId_1)
  AND ("p"."ProjectRole" IN ('Manager', 'Owner', 'Admin', 'Coordinator', 'PM')
    OR "p"."Iso19650Role" = 'PM'
    OR "p"."Iso19650Role" = 'A'
    OR "p"."Iso19650Role" = 'BC')

-- ProjectRoles.CanApproveSitePhotosPredicate
WHERE (@__ef_filter__BypassTenantFilter_0 OR "p"."TenantId" = @__ef_filter__CurrentTenantId_1)
  AND ("p"."ProjectRole" IN ('Manager', 'Owner', 'Admin', 'PM')
    OR "p"."Iso19650Role" = 'PM'
    OR "p"."Iso19650Role" = 'A')
```

**Fully server-side.** Every disjunct is in the `WHERE`; EF even folded the `ProjectRole ||` chain
into a single `IN`. No client evaluation, nothing pulled into memory, and the tenant filter still
applies. Captured with a throwaway test that was deleted before this comment — the tree is clean.

### 3. The widening property — now provable from the SQL, not argued

The old gate was `ProjectRole = 'PM'`. Read straight off the generated SQL:

- `'PM'` ∈ `IN ('Manager','Owner','Admin','Coordinator','PM')` — curate ✅
- `'PM'` ∈ `IN ('Manager','Owner','Admin','PM')` — approve ✅

Every row that passed the old gate passes both new ones. **Strictly widening; no user can lose
access at any of the eleven sites.** The legacy-`"PM"` tolerance the body describes is what makes
this true, and it is visible in the emitted SQL rather than only in the C#.

Two further things checked, because widening can also happen by accident:

- **No qualifying conditions were dropped.** `HasCapabilityAsync` filters
  `m.ProjectId == projectId && m.UserId == userId && m.IsActive` before applying the capability
  predicate — the same three the eleven call sites used. Verified against the helper body.
- **The Admin/Owner short-circuit is genuinely pre-existing.** The body claims this; I did not take
  it on trust. On `main`, `SitePhotosController.IsApproverAsync:878-879`,
  `PhotoAlbumsController.IsCuratorAsync:389-390`, `PhotoPolicyController:97-98` and
  `DistributionGroupsController:196-197` each open with the identical
  `var role = User.FindFirst("role")?.Value ?? ""; if (role is "Admin" or "Owner") return true;`.
  The helper preserves that and adds `IsInRole(...)` — a further widening, consistent with the
  stated property.

### 4. Post-change grep — re-run from scratch, and the body's list is the correct one

`ProjectRole == "PM"` in executable code on this branch: **zero.** All eleven `main` sites converted
(DistributionGroups:202, PhotoAlbums:395, PhotoChecklists:320, PhotoPolicy:103, PhotoShare:122,
SavedViews:136, SitePhotos:884, SitePhotosExt:610, BackgroundJobs:314, DailyPhotoDigestJob:157,
P6LiveLinkService:288 — counted on `a5109e3e5`).

Survivors, all non-PM and all intended — **seven, not five.** The brief I was working from named
five; this PR's body already lists all seven, and the two extra are correct to leave:

| Survivor | Gate | Verdict |
|---|---|---|
| `CdeContainers:255` | Manager/Admin/Owner | leave — vocabulary is assignable today, gate is **live** |
| `WeeklyDigestJob:48-49` | Manager/Admin/Coordinator/Owner | leave |
| `TenantAdmin:51` | Author | leave |
| `CoordinatorWorkloadJob:39` | BimManager/Coordinator | leave |
| `QuotaGuardService:76` | Author (billing seat axis) | **hands off**, report-only |
| `BackgroundJobs:278` | Owner/Admin | leave — *not in the five, correctly kept* |
| `DailyPhotoDigestJob:126` | ClientGuest | leave — *not in the five, correctly kept* |

**`CdeContainersController` is not touched by this PR** — confirmed by
`git diff --name-only a5109e3e5..19e1737ce | grep -i cde` → no output. Its gate stays live and
unchanged, as required.

### 5. CI — the two red checks are the `main` baseline

Same proof as on #542.

| | this PR ([run 30720916677](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30720916677)) | `main` ([run 30715674338](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30715674338)) |
|---|---|---|
| `dotnet build` | **Build succeeded** | **Build succeeded** |
| Full-suite step | hangs ~46 min | hangs ~46 min |
| Terminating error | `##[error]no test summary in output — the run did not complete` → exit 1 | identical |

Cause is in the workflow, not this diff: `.github/workflows/planscape-server.yml` **declares no Redis
service**, but Gate 2 boots code that opens a Redis multiplexer with `AbortOnConnectFail` at its
default, so `dotnet test … | tee /tmp/test-output.txt` never terminates and
`check-new-failures.sh` finds no summary. `Planscape Server` has failed on **every `main` run since
2026-07-20** (last green: `ddea801d5`). `CI Gate` is derived — its only error is
`Timed out (30m) waiting for: Build & Test`, and `Build & Test` takes ~47 min. **No new failure is
introduced here.**

---

## Ready for review. Nothing merged.

The capabilities endpoint is **not** built, per the STOP-and-ask. To restate the proposed shape for
an explicit decision before any code:

```
GET /api/projects/{projectId}/members/capabilities
200 →
{
  "curateProject": true,
  "approveSitePhotos": false
}
```

Flat boolean map, exactly two keys, no nested role objects, no echo of `projectRole` /
`iso19650Role`. Three clients will read this and the shape is a one-way door once they do — so the
two keys are named after the *capabilities* rather than the roles that grant them, which lets the
grant rules change later without a client change. It reads only data the caller already has access
to. **Not implemented; awaiting your go-ahead, and it lands as its own small PR after this one is
approved.**

PR #541, opened 2026-08-01
docs: Phase 6A triage of the clientless controllers

StingD85:
## Phase 6A — triage gate. No feature code in this PR.

Adds `docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md`. **Awaiting review before anything is built.**

## Your correction is confirmed

Re-derived the clientless list myself across 1,856 client files, **including** `Planscape.Server/src/Planscape.API/wwwroot` and **excluding** `wwwroot/_next/**` and `wwwroot/vendor/**`.

`PublicConfig` (`api/public-config`), `RoleBuckets` (`api/state-machine/role-buckets`), `TenantKeywords` and `TenantBimManagerRoles` are all called from `wwwroot/js/dashboard.js` — the last two also from `wwwroot/index.html` and `wwwroot/app/index.html`. Not clientless, left alone. (My first grep used the wrong needles and found nothing; corrected before reporting.)

All ten in the brief are confirmed clientless by exact route-string grep. **Two more are too:**

- **MaterialSync** (`api/MaterialSync`, 53 ln) — not on your list, and it has a security finding.
- **PhotoExport** — already scheduled as Phase 5, listed for completeness.

### S1 — MaterialSync is a cross-tenant read

`[Authorize]` and **nothing else**: no `[ProjectAccess]`, no membership check, no tenant filter. And it **bypasses EF entirely** — it reads/writes `ContentRootPath/App_Data/material_snapshots/{projectId}.json` directly (`:31-35, :47-50`).

Because it never touches EF, **the global tenant query filter cannot protect it.** Any authenticated user in any tenant who has a project GUID can `GET api/MaterialSync/snapshot/{id}` and read that project's material library.

This is not a UI gap. It is an authorization fix, and it is independent of every client decision below. Not touched — flagging for your call.

### B1 — WorkOrders and GlobalIdRegistry have no table

Both are in the EF model snapshot — `b.ToTable("WorkOrders")` (`PlanscapeDbContextModelSnapshot.cs:8910`), `b.ToTable("GlobalIdRegistry")` (`:2372`) — but **no migration creates either**. Grep for `name: "WorkOrders"` / `name: "GlobalIdRegistry"` across all non-Designer migrations: nothing.

On a database built by `dotnet ef database update`, every action on these two throws "relation does not exist". **Missing migration, not missing UI** — which moves both to Tier 3 regardless of their merits.

Worth recording the opposite result for WorkOrders' *downstream*: `StingTools/BIMManager/PlatformEvents/ParamStampEventHandler.cs` is a real handler that writes the parameter onto the Revit element. So the K2 loop is live and the feature genuinely completes — once the table exists.

*Caveat:* I could not check the production database. The tables may exist there via another path. That is exactly what the runtime pass would have settled.

### C1 — ApprovalChains duplicates a mechanism that already ships

| | route | client |
|---|---|---|
| `DocumentsController` | `POST /documents/{id}/approval-request`, `PUT .../approval/{aid}`, `GET .../approval-status` | **mobile** (`endpoints.ts:1005,1012`) |
| `ApprovalChainsController` | `/documents/{docId}/approval-chain` + `/decisions` | none |

The chain version adds multi-step and parallel approval, so it is not simply redundant — but "wire a client" is the wrong first question. Shipping a UI for the second without deciding which is canonical produces **two approval records for one document**.

## Both docstring-drift examples confirmed

- **Mfa** — `:13` says *"Production implementation would use a TOTP library (e.g. Otp.NET)"*. It already does: `using OtpNet;` (`:5`), `IDataProtector` (`:23`), `totp.VerifyTotp(req.Code, out _, new VerificationWindow(2, 2))` (`:119`) — the ±2-step window you described. Implementation real, comment wrong.
- **DeviceTwins** — `:11-12` claims it *"Powers the mobile Live tab (RAG list)"*. There is **no Live tab** in `Planscape/app/`; the only case-insensitive match is `de`**live**`rables`. Telemetry ingest does exist and is registered, so the downstream is not dead — but the claimed consumer isn't real.

Neither fixed here, because 6A carries no code.

## DI is clean across the group

All six injected services resolve: `IAuditService` (`Program.cs:428`), `ITenantContext` (`:365`), `IDeviceTwinService` (`:727`), `ITwinBindingService` (`:729`), `IPlatformEventService` (`:724`), `IDataProtector` via `AddDataProtection()` (`:506`).

So the known-prior failure mode — a controller 500ing on every call from a missing registration, misreported by the browser as CORS — **does not apply to any of these ten**. My first grep reported all six as unregistered; that was a false negative from a too-strict pattern, caught before it reached the report.

## Tier recommendations

**Tier 1 — CdeContainers.** Agreed, and I verified your note: the gate at `:255` is `ProjectRole in {Manager, Admin, Owner}`, a vocabulary that **is** assignable from the web dropdown today, so unlike the PM gates it is **live**. It must **not** be folded into Phase 1's `CanCurateProject` without a deliberate decision — that helper also admits `Coordinator` and three ISO codes, which would *widen* who can restructure the CDE folder tree. Backed, scoped, DI-clean.

**Tier 1 — DataRights: STOP, as you anticipated.** `POST erase` touches `Users`, `Projects`, `ProjectMembers`, `Documents`, `ProjectModels`, `Issues`, `Invoices`, `Subscriptions`, `AuditLogs`, `Tenants`. I have **not** yet traced whether each is a soft-delete, an anonymisation or a hard delete, nor whether `cancel-erase` can genuinely undo it. No button should ship until the UI copy can describe truthfully what it destroys. That trace is read-only and needs no server — it is the first thing I would do in 6B.

**Tier 2** — `AssetDataSheets`, `OfflineManifest`, `Mfa` (fix M1), `ApprovalChains` *after C1 is decided*.

**Tier 3** — `WorkOrders` + `GlobalIdRegistry` (blocked on B1), `DeviceTwins` (fix M2; needs a real telemetry consumer decision), `CaseStudy` (sales-ops, not product parity).

**Out of band** — `MaterialSync` (S1).

## ⚠️ Runtime column is NOT measured

Item 2 requires exercising every action against a running server. I could not:

- Docker Desktop is installed but its daemon would not start (`npipe:////./pipe/dockerDesktopLinuxEngine` — no such file; launching the app left no process in `tasklist`).
- No local Postgres listening on 5432.
- SQLite is not a substitute — `Program.cs`'s schema block issues a Postgres-only `information_schema` query with no try/catch, so the host dies before serving a request.

Where a static proxy exists I used it and labelled it: **DI registration** (clean) and **migration presence** (found B1). What remains unmeasured: real status codes and error bodies per action, whether the two missing tables exist in a deployed database, and whether anything 500s for a reason not visible in source.

**To close it I need** the Docker daemon started, or a Postgres reachable on 5432 plus a connection string. Happy to run the full runtime pass and amend this table once either exists.

## Notes

Docs-only, no code. Known-broken CI: "Build & Test / CI Gate" fails on a Postgres container (`role "root" does not exist`) — pre-existing on `main`, unrelated.

### DIFF
--- a/docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md
+++ b/docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md
@@ -1,0 +1,182 @@
+# Phase 6A — clientless controller triage
+
+**Measured 2026-08-02 on `main` @ `a5109e3e5`.** Triage only — no feature code in this PR.
+
+> **Runtime column is NOT MEASURED.** Item 2 of the brief requires exercising every action
+> against a running server. Docker Desktop is installed but its daemon would not start on this
+> machine (`npipe:////./pipe/dockerDesktopLinuxEngine` — no such file; launching the app left no
+> process), and there is no local Postgres on 5432. Everything below is *static* measurement.
+> Where a static proxy exists for "does it run" — DI registration, migration presence — it was
+> used and is labelled as such. See [Runtime gap](#runtime-gap).
+
+---
+
+## 0. Re-derived list — the brief's ten, plus two
+
+Re-derived per instruction, **including** `Planscape.Server/src/Planscape.API/wwwroot` and
+**excluding** `wwwroot/_next/**` and `wwwroot/vendor/**`. Clients scanned: `StingTools/`,
+`planscape-web/`, `Planscape/` (mobile), `StingBridge/`, `Planscape.Server/.../wwwroot/`,
+`marketing-site/` — 1,856 files.
+
+**Your correction is confirmed.** `PublicConfig` (`api/public-config`), `RoleBuckets`
+(`api/state-machine/role-buckets`), `TenantKeywords` (`api/admin/tenant-keywords`) and
+`TenantBimManagerRoles` (`api/admin/tenant-bim-manager-roles`) are all called from
+`wwwroot/js/dashboard.js`; the latter two also from `wwwroot/index.html` and
+`wwwroot/app/index.html`. Not clientless. Left alone.
+*(One nuance: I found no `public-config` reference under `marketing-site/` in this checkout —
+possibly the built bundle isn't committed. Doesn't change the verdict.)*
+
+All ten in the brief are confirmed clientless by exact route-string grep. **Two more are too:**
+
+| Extra | Route | Note |
+|---|---|---|
+| **MaterialSync** | `api/MaterialSync` | 53 ln, 2 actions. Not on the brief's list. **Has a security finding — see S1.** |
+| **PhotoExport** | `api/projects/{id}/photo-export` | Already scheduled as Phase 5; listing for completeness. |
+
+---
+
+## 1. Triage table
+
+Legend — **Backed**: DbSet + a migration that actually creates the table.
+**DI**: every injected service resolves (static check of `Program.cs`).
+
+| # | Controller | Ln / actions | Auth gate | Tenant scoping | Backed | DI | Verdict |
+|---|---|---|---|---|---|---|---|
+| 1 | **DataRights** | 114 / 3 | `[Authorize(Roles="Owner,Admin")]` | `ITenantContext` | n/a (reads + deletes) | ✅ | **Tier 1 — but STOP, see D1** |
+| 2 | **CdeContainers** | 287 / 6 | `[Authorize]` + role gate `:255` | `TenantId` + `ProjectMembers` | ✅ `20260517000003` | ✅ | **Tier 1 — confirmed viable** |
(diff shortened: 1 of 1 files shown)


StingD85:
Superseded by #543, which is the same document amended twice (it re-derives the same triage table, overturns B1, and corrects DataRights). Both PRs were the sole change to `docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md` and branched independently off `main`, so they would have conflicted on merge. #543 is self-contained and carries the current text — closing this one so there is a single live draft.

Note for #543: its B1 section is itself now superseded — see the fresh-database migration measurement in the follow-up.

PR #542, opened 2026-08-01
fix(security)!: close cross-tenant read and write on MaterialSync — BREAKING route change

StingD85:
## 🔒 SECURITY + ⚠️ BREAKING ROUTE CHANGE — cross-tenant read **and write**, live on a deployed endpoint. Prioritise for deploy.

PR-A from the sequence. Standalone, no dependencies, no feature code.

**This PR is two things, and the second one is not just a side effect of the first: it fixes a cross-tenant auth hole, AND it changes the URL of both endpoints on this controller.** Both endpoints are on `main` today and therefore deployed. Full before/after routes below.

## The defect

`MaterialSyncController` carried `[Authorize]` and **nothing else** — no `[ProjectAccess]`, no membership check, no tenant filter. It also does not use EF, so the DbContext's global tenant query filter (which silently covers most controllers that forget) **could not reach it**.

Your review was right that I under-reported it: the **write** is the worse half.

- `GET api/MaterialSync/snapshot/{projectId}` — any authenticated user in any tenant reads another tenant's material library.
- `POST api/MaterialSync/snapshot` — same, but `File.WriteAllTextAsync` **over** it.
- `Notes = $"Snapshot stored at {path}"` returned the server filesystem layout to the caller.

## Proved before fixing

A throwaway exploit test **passed** against the pre-fix controller:

```
Passed!  - Failed: 0, Passed: 1
```

asserting all three in one run: the file was created for an arbitrary project GUID with no authorisation of any kind, the response body contained `App_Data`, and the planted row (`"PWNED"`) read straight back out. That test was deleted before commit; the six real tests replace it.

## ⚠️ Route change — BREAKING, deliberate, stated rather than silent

**The class-level `[Route]` moved.** This is the root of the change; both action routes follow from it:

```
[Route("api/[controller]")]                         ->  [Route("api/projects/{projectId:guid}/material-sync")]
```

**Every route on this controller, in full:**

| Verb | OLD (on `main`, deployed) | NEW (this PR) |
|---|---|---|
| POST | `api/MaterialSync/snapshot`<br>`projectId` in **body** (`MaterialSyncRequest.ProjectId`) | `api/projects/{projectId:guid}/material-sync/snapshot`<br>`projectId` from **route** |
| GET | `api/MaterialSync/snapshot/{projectId:guid}`<br>`projectId` in **path segment after `snapshot`** | `api/projects/{projectId:guid}/material-sync/snapshot`<br>`projectId` from **route prefix** |

Read the action attributes in the diff with the class route in mind — in isolation they are misleading:

- `GET` goes **`[HttpGet("snapshot/{projectId:guid}")]` → `[HttpGet("snapshot")]`**. The `{projectId:guid}` was not dropped; it moved *up* into the class route, which is why `GetSnapshot(Guid projectId, …)` still binds. The parameter is still route-supplied, just from the prefix.
- `POST` keeps **`[HttpPost("snapshot")]` unchanged**, but its *effective* URL changed anyway because the class route changed, and `PostSnapshot` gained a `Guid projectId` **route** parameter it did not have before. That is the reason `req.ProjectId = projectId` is safe: `projectId` is the route-authorised value, not anything the caller put in the body.

So both endpoints move; neither old URL survives. `api/MaterialSync/*` returns **404** after this merges.

**Why it was forced, not stylistic.** `ProjectAccessAttribute` resolves its subject from *route data* and deliberately no-ops when the route carries no project id — "No project in the route — let the action handle it" (`ProjectAccessAttribute.cs:39-47`, verified: it calls `await next(); return;`). Adding `[ProjectAccess]` while leaving the id in the body would have **looked** like protection and changed nothing. That is precisely the failure mode this whole exercise is about, so I moved it.

**Risk assessment: nil, but assert it, don't assume it.** Re-verified by exact-string grep for `api/MaterialSync`, `MaterialSync/snapshot` and `MaterialSync` across `StingTools/`, `planscape-web/`, `Planscape/`, `StingBridge/`, `marketing-site/` and `stingtools-core/` — **zero callers**, and zero references in any `.md`, `.json` or `.yml` either (no docs, no OpenAPI fixture, no CI config pins the old path). Nothing in this repo breaks. Flagged prominently anyway because it is a contract change to a deployed endpoint, and any out-of-repo or hand-rolled caller (Postman collection, curl script, a fork) **will** break and must be repointed.

The DTO keeps `ProjectId` for compatibility. The **route is authoritative**; a body that disagrees is rejected with `400 project_id_mismatch` rather than ignored, so a caller can never believe it wrote to a project it did not.

## Not changed, per your item (c)

Snapshots are still files under `ContentRootPath/App_Data`. Two known problems, recorded in the class comment rather than acted on:

1. the container filesystem is ephemeral — a Render redeploy loses every snapshot;
2. being outside EF is exactly what put this data beyond the tenant filter's reach.

Moving it into a table is a persisted-storage change and needs its own decision. Not done here.

## Tests — 6, green

| test | asserts |
|---|---|
| `Outsider_from_another_tenant_is_denied_on_GET` | 403 |
| `Outsider_from_another_tenant_is_denied_on_POST` | 403 |
| `A_denied_POST_does_not_write_the_file` | file absent after the denied call |
| `Project_member_can_write_then_read_back_the_snapshot` | 200 + **NON-EMPTY** round-trip (`Materials[0].Name == "Concrete C30"`) |
| `Response_does_not_leak_the_server_filesystem_path` | no temp root, no `App_Data`, no `.json` in `Notes` |
| `A_body_naming_a_different_project_is_rejected_not_ignored` | `400 project_id_mismatch` |

`Passed! - Failed: 0, Passed: 6`

**Fixture-trap guard**, as you flagged: the DbContext's filter falls back to `Guid.Empty` without an `ITenantContext`, under which a "denied" assertion passes for the wrong reason. Every context is built with a tenant, and the happy path asserts on non-empty round-tripped content, not merely on absence of an exception.

## What these tests do NOT cover — stated so a green run isn't over-read

They construct the controller directly (the `HandoffProvisioningSqliteTests` precedent) and therefore exercise **`RequireProjectMemberAsync`**, the in-action guard. They do **not** exercise `[ProjectAccess]` — that is an `IAsyncActionFilter` and only runs inside the MVC pipeline, which needs a booted host, and the host cannot boot against SQLite (`Program.cs`'s schema block issues a Postgres-only `information_schema` query with no try/catch). `[ProjectAccess]` is defence in depth here; verifying it belongs to the runtime pass.

They also do not exercise **routing** — constructing the controller directly bypasses the route table entirely, so the new URLs are asserted by reading the attributes, not by a request. Confirming the new paths resolve (and that the old ones 404) belongs to the same runtime pass.

## Independence

Branched from `origin/main`, not from #540. It briefly used `ProjectRoles.Contributor` from the Phase 1 branch; caught at compile and replaced with the literal so this PR stands alone and can be deployed without waiting on the capability layer.

## Notes

Known-broken CI: "Build & Test / CI Gate" fails on a Postgres container (`role "root" does not exist`) — pre-existing on `main`, unrelated. Verified locally: `dotnet build` succeeded, 0 errors.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/MaterialSyncController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/MaterialSyncController.cs
@@ -1,53 +1,102 @@
 using System.Text.Json;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Planscape.API.Authorization;
+using Planscape.API.Services;
 using Planscape.Core.DTOs;
+using Planscape.Infrastructure.Data;
 
 namespace Planscape.API.Controllers;
 
 /// <summary>
 /// I-2 — Material library snapshot endpoint.
-/// Receives a snapshot of the project's materials from the Revit
-/// plugin and stores it as a JSON blob under the server's data
-/// directory. Minimum-viable persistence — full per-row DB rows
-/// can land in a follow-up migration.
+/// Receives a snapshot of the project's materials from the Revit plugin and
+/// stores it as a JSON blob under the server's data directory.
+///
+/// SECURITY — why this controller looks different from its first version
+/// ---------------------------------------------------------------------
+/// It carried <c>[Authorize]</c> and nothing else: no <c>[ProjectAccess]</c>, no
+/// membership check, no tenant filter. It also does not use EF, so the DbContext's
+/// global tenant query filter — which silently protects most controllers that
+/// forget — could not reach it. Any authenticated user in any tenant who had a
+/// project GUID could READ another tenant's material library, and
+/// <c>PostSnapshot</c> was equally unguarded, so they could OVERWRITE it too.
+///
+/// The project id also had to move from the request BODY into the ROUTE.
+/// <see cref="ProjectAccessAttribute"/> resolves its subject from route data and
+/// deliberately no-ops when the route carries no project id ("let the action
+/// handle it"), so leaving the id in the body would have added the attribute
+/// while changing nothing — a fix that reads as protection and isn't.
+///
+/// PERSISTENCE — deliberately NOT changed here
+/// -------------------------------------------
+/// Snapshots are still files under ContentRootPath/App_Data. That has two known
+/// problems: the container filesystem is ephemeral (a Render redeploy loses every
+/// snapshot), and being outside EF is what made the tenant filter unable to help
+/// above. Moving this to a table is a persisted-storage change and needs its own
+/// decision, so it is only recorded here, not done.
 /// </summary>
 [ApiController]
-[Route("api/[controller]")]
+[Route("api/projects/{projectId:guid}/material-sync")]
 [Authorize]
+[ProjectAccess]
 public class MaterialSyncController : ControllerBase
 {
     private readonly IWebHostEnvironment _env;
-    public MaterialSyncController(IWebHostEnvironment env) { _env = env; }
+    private readonly PlanscapeDbContext _db;
+
+    public MaterialSyncController(IWebHostEnvironment env, PlanscapeDbContext db)
+    {
+        _env = env;
+        _db = db;
+    }
 
     [HttpPost("snapshot")]
     public async Task<ActionResult<MaterialSyncResponse>> PostSnapshot(
+        Guid projectId,
         [FromBody] MaterialSyncRequest req,
         CancellationToken ct)
     {
-        if (req == null || req.ProjectId == Guid.Empty)
-            return BadRequest(new { error = "ProjectId required" });
+        if (await this.RequireProjectMemberAsync(_db, projectId, ct) is { } denied) return denied;
+        if (req == null) return BadRequest(new { error = "body_required" });
 
-        var dir = Path.Combine(_env.ContentRootPath, "App_Data", "material_snapshots");
-        Directory.CreateDirectory(dir);
-        var path = Path.Combine(dir, $"{req.ProjectId}.json");
+        // The DTO still carries ProjectId for backwards compatibility. The ROUTE
+        // is authoritative — it is what was authorised. A body that disagrees is
+        // rejected rather than silently ignored, so a caller can never believe it
+        // wrote to a project it did not.
+        if (req.ProjectId != Guid.Empty && req.ProjectId != projectId)
+            return BadRequest(new { error = "project_id_mismatch" });
+
+        var path = SnapshotPath(projectId);
+        Directory.CreateDirectory(Path.GetDirectoryName(path)!);
+        req.ProjectId = projectId;
         var json = JsonSerializer.Serialize(req, new JsonSerializerOptions { WriteIndented = false });
         await System.IO.File.WriteAllTextAsync(path, json, ct);
 
         return Ok(new MaterialSyncResponse
         {
             RowsAccepted = req.Materials?.Count ?? 0,
-            Notes = $"Snapshot stored at {path}",
+            // Was $"Snapshot stored at {path}", which handed the server's
+            // filesystem layout to any caller.
+            Notes = "Snapshot stored.",
         });
     }
 
-    [HttpGet("snapshot/{projectId:guid}")]
+    [HttpGet("snapshot")]
     public async Task<ActionResult<MaterialSyncRequest>> GetSnapshot(Guid projectId, CancellationToken ct)
     {
-        var path = Path.Combine(_env.ContentRootPath, "App_Data", "material_snapshots", $"{projectId}.json");
+        if (await this.RequireProjectMemberAsync(_db, projectId, ct) is { } denied) return denied;
+
+        var path = SnapshotPath(projectId);
         if (!System.IO.File.Exists(path)) return NotFound();
         var json = await System.IO.File.ReadAllTextAsync(path, ct);
         var snap = JsonSerializer.Deserialize<MaterialSyncRequest>(json);
         return snap != null ? Ok(snap) : (ActionResult<MaterialSyncRequest>)NotFound();
     }
+
+    /// <summary>Path is derived from the ROUTE-authorised project id only —
+    /// never from caller-supplied text — so it cannot be steered outside the
+    /// snapshot directory.</summary>
+    private string SnapshotPath(Guid projectId)
+        => Path.Combine(_env.ContentRootPath, "App_Data", "material_snapshots", $"{projectId}.json");
 }
--- a/Planscape.Server/tests/Planscape.Tests/MaterialSyncAuthorizationTests.cs
+++ b/Planscape.Server/tests/Planscape.Tests/MaterialSyncAuthorizationTests.cs
@@ -1,0 +1,274 @@
+using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.AspNetCore.Hosting;
+using Planscape.API.Controllers;
+using Planscape.Core.DTOs;
+using Planscape.Core.Entities;
+using Planscape.Core.Interfaces;
+using Planscape.Infrastructure.Data;
+
+namespace Planscape.Tests;
+
+/// <summary>
+/// MaterialSyncController's authorization, which did not exist.
+///
+/// THE DEFECT
+/// ----------
+/// The controller carried [Authorize] and nothing else — no [ProjectAccess], no
+/// membership check, no tenant filter — and it does not use EF, so the DbContext's
+/// global tenant query filter could not cover for it the way it silently does for
+/// most controllers that forget. Any authenticated user in ANY tenant who had a
+/// project GUID could read another tenant's material library, and PostSnapshot was
+/// equally unguarded, so they could overwrite it too. It also returned the server's
+/// filesystem path in the response.
+///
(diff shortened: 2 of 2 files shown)


StingD85:
## Ready for review — CI red proven pre-existing, not introduced here

### 1. The two red checks are the `main` baseline. Proof, side by side.

| | this PR (`ff4b02c90`) | `main` (`a5109e3e5`, most recent run) |
|---|---|---|
| Run | [30721898700](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30721898700) | [30715674338](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30715674338) |
| `dotnet restore` | ok | ok |
| `dotnet build` | **Build succeeded, 0 Error(s)** | **Build succeeded, 0 Error(s)** |
| Gate 1 — Postgres tests | ran | ran |
| Gate 2 — full suite | hangs ~45 min, never emits a summary | hangs ~46 min, never emits a summary |
| Terminating error | `##[error]no test summary in output — the run did not complete` → `exit code 1` | `##[error]no test summary in output — the run did not complete` → `exit code 1` |

The error strings are identical, and so is the log body preceding them: a
`StackExchange.Redis.RedisConnectionException` retry storm and a 10-second
`FATAL: role "root" does not exist` drumbeat from the Postgres service container.

**Root cause, both sides — two independent workflow defects, neither in this diff:**

1. `.github/workflows/planscape-server.yml` **declares no Redis service**, but Gate 2
   (`Run full test suite`) boots code that opens a Redis multiplexer with
   `AbortOnConnectFail` left at its default. It retries forever, so
   `dotnet test … | tee /tmp/test-output.txt` never terminates. The next step,
   `tests/check-new-failures.sh`, then finds no summary in the file and exits 1.
2. The service healthcheck is `--health-cmd pg_isready` with **no `-U planscape`**, so
   `pg_isready` probes as the container's OS user `root`, which is why Postgres logs
   `role "root" does not exist` on a 10s cycle. Log noise here, but it makes the real
   cause harder to find.

**`Planscape Server — Build, Test & Deploy` has failed on _every_ `main` run since
2026-07-20** (last green: run [29725761782](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/29725761782), `ddea801d5`). 22 consecutive
`main` failures.

**`CI Gate` is derived, not independent.** Its only error is
`##[error]Timed out (30m) waiting for: Build & Test` — `Build & Test` takes ~47 min
because of (1), which is longer than CI Gate's 30-minute wait. It fails identically on
[#540](https://github.com/beckykyomugisha/STINGTOOLS/pull/540) ([run 30720916671](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30720916671), same two lines). CI Gate does not run on
`main` at all (`pull_request` event only), which is why there is no `main` row for it —
the `Build & Test` comparison above is the whole of it.

Docs-only PRs on this repo pass CI Gate simply because path filters keep
`planscape-server.yml` from running. Nothing about this PR's content.

**Conclusion: this PR introduces no new CI failure.** Fixing either workflow defect is
out of scope here and wants its own PR.

### 2. Rebase — no-op, already current

`claude/materialsync-security` is `a5109e3e5` + one commit. `a5109e3e5` is the current
tip of `origin/main`, so the branch is **0 commits behind, 1 ahead**. Nothing to rebase.
`mergeable: MERGEABLE`.

### 3. Tests re-confirmed green, and non-vacuously so

```
Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6, Duration: 7 s
```
(local, `dotnet test --filter FullyQualifiedName~MaterialSyncAuthorizationTests`)

Re-read against the standing `Guid.Empty` fixture trap, and it is closed:

- Every `PlanscapeDbContext` is constructed with an explicit `FixedTenant(tenantId)`,
  never the no-`ITenantContext` path that falls back to `Guid.Empty` and matches nothing.
- The positive path asserts on **content, not absence of an exception**:
  `Assert.NotEmpty(snap.Materials)` **and**
  `Assert.Equal("Concrete C30", snap.Materials[0].Name)` **and**
  `Assert.Equal(w.ProjectA, snap.ProjectId)`.
- The red/green control is structural: the denial tests and the happy-path test share
  one fixture and one project, and differ only in **which user acts**
  (`OutsiderInB`/`TenantB` → 403; `MemberOfA`/`TenantA` → 200 + non-empty). A fixture
  that returned nothing to everybody would fail the happy path.

### 4. Zero callers — re-verified independently, from scratch

Fresh repo-wide grep for `api/MaterialSync`, `MaterialSync/snapshot` and `material-sync`
across all of `StingTools/`, `Planscape/`, `planscape-web/`, `StingBridge/`,
`marketing-site/`, `stingtools-core/` and every `.md`/`.json`/`.yml`: **zero hits.**
`MaterialSyncRequest` appears in exactly two files — its own DTO definition and this
controller. No client, no doc, no OpenAPI fixture, no CI config pins either URL.

That is the whole safety argument for the breaking change: **both routes move, neither
old URL survives, `api/MaterialSync/*` returns 404 after merge — and nothing in this
repo calls it.** An out-of-repo caller (Postman collection, curl script, a fork) will
break and must be repointed. The PR body carries the full OLD → NEW table; the title
carries the conventional-commits `!`. This repo has **no `breaking` or `security`
label** to apply — worth adding two, separately.

### What is still NOT verified — do not read the green as more than it is

- **`[ProjectAccess]` is never exercised.** The tests construct the controller directly,
  so they hit `RequireProjectMemberAsync` (the in-action guard) only. `[ProjectAccess]`
  is an `IAsyncActionFilter` and needs a booted MVC pipeline, which cannot start on
  SQLite (`Program.cs`'s schema block issues a Postgres-only `information_schema` query
  with no try/catch). It is defence in depth here; the in-action guard is what these
  tests prove.
- **Routing is never exercised.** Constructing the controller bypasses the route table.
  The new URLs are asserted by reading attributes, not by issuing a request. That the
  new paths resolve, and that the old ones 404, is unproven.
- **No runtime pass against a booted host** was run for this PR.

Both gaps want the same thing: one request each against a running stack.

**Not merging** — this auto-deploys to production, so that is the owner's call.

StingD85:
## Merge-readiness: both red checks are the pre-existing `main` baseline

Short version: **there are not two independent failures.** There is one environmental failure, plus a second check that is a pass-through of the first. Neither is caused by this PR.

### 1. `Build & Test` — identical failure signature to `main`

| | run | commit | signature |
|---|---|---|---|
| **`main` baseline** | [30715674338](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30715674338) (job `91410710406`) | `a5109e3e` | `FATAL: role "root" does not exist` |
| **This PR (#542)** | [30721898700](https://github.com/beckykyomugisha/STINGTOOLS/actions/runs/30721898700) (job `91426937328`) | `ff4b02c9` | `FATAL: role "root" does not exist` |

Same Postgres service-container auth failure, repeating on the ~10s health-check loop in both, from the container itself (`[2617] FATAL:`, `[2602] FATAL:`) — before any test in this PR runs.

The Redis `RedisConnectionException` lines that appear earlier in both logs are logged noise on both sides, not the failing step. Both jobs report `0 Error(s)` from the build itself.

### 2. `CI Gate` — cannot be diffed against `main`, because it never runs on `main`

`ci-gate.yml:20-22`:

```yaml
on:
  pull_request:
    branches: [main]
```

`pull_request` only. It is the required check **for merging into** `main`, so it has no `main`-branch run to diff against. Confirmed against the workflow list for `main`, which contains seven workflows and no `CI Gate`.

So instead of a diff, the failure reason itself — it has none of its own:

```
##[error]Timed out (30m) waiting for: Build & Test
```

Preceded by ~90 lines of `waiting on: Build & Test`. `CI Gate` is an aggregator (documented at `ci-gate.yml:3-18`): it resolves changed areas with `dorny/paths-filter` and waits only for the checks those areas trigger. This PR touches server code, so it waits on `Build & Test` — which fails, and which ran 47m11s against `CI Gate`'s 30m ceiling.

**Control:** the two docs-only PRs open right now, #543 and #546, both **pass** `CI Gate` in 3-5s — nothing build-relevant changed, so it waits for nothing. That isolates the difference to "does the PR touch a path that triggers `Build & Test`", not to anything in this diff.

### 3. The failure is unconditional

Last 15 runs of `Planscape Server — Build, Test & Deploy`, every branch:

```
failure  claude/datarights-ui          failure  main
failure  claude/materialsync-security  failure  main
failure  claude/capability-layer       failure  main
failure  claude/doc-lifecycle-audit    failure  claude/remove-fabricated-clashes
failure  claude/viewer-health-poll     failure  claude/panel-restore-clamp
failure  claude/viewer-canonical-assignee
failure  claude/viewer-create-issue    failure  claude/member-invite-visibility
failure  claude/viewer-no-fabrication
failure  claude/viewer-level-dropdown
```

15/15 failures, 0 passes. Several are **viewer-only** branches (`claude/viewer-level-dropdown`, `claude/viewer-no-fabrication`) that change no server code at all and still fail identically. A defect that fails on a branch touching only JavaScript is not a defect in this PR's C#.

### What this does *not* prove

It does not prove #542 is correct — only that CI is not evidence against it. The evidence for correctness is separate and already in the PR: the pre-fix exploit test that **passed** against the vulnerable controller (arbitrary-project write, `App_Data` path disclosure, read-back of a planted row), and the 6 authorization tests that now cover it.

It also leaves a real pre-existing defect visible: `CI Gate`'s 30m timeout is shorter than `Build & Test`'s observed 47m runtime. On this run that is downstream of the Postgres retry loop inflating the duration, so I am **not** claiming the timeout is independently misconfigured — I have no passing-run duration to compare against, because there are no passing runs.

Fixing the Postgres container auth is out of scope here and unrelated to this security fix.

PR #543, opened 2026-08-01
docs: Phase 6A triage (consolidated) — render.yaml voids B1; runtime pass reinstated; two stale docstrings fixed

StingD85:
## The consolidated Phase 6A triage

This is now the **single** triage document. It carries the B1 retraction, the
`DocumentApprovals` closure, the C1 decision, the `bypassesAcl` note and the two
docstring fixes.

### On the three PRs the brief asked me to fold

| PR | State | Action |
|---|---|---|
| **#541** | **already CLOSED** | Nothing to do — its history is already reparented into this branch (`5d7bca5ba`). |
| **#543** (this) | open | The consolidated document. |
| **#544** | open | **Left open deliberately — it is not a triage PR.** |

**#544 is not a third conflicting triage PR.** It is `feat(data-rights)`, touching
`Planscape.Desktop/src/renderer/*`, `DataRightsController.cs` and
`DataRightsExportTenantScopeTests.cs`. It shares **no file** with
`docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md`, so there is nothing to conflict and no
reason to close unrelated feature work. Verified against the PR file list rather than
assumed. Only #543 touches the triage doc.

---

### B1 — retracted, and the citations that supported it were wrong

The retraction stands and is now cited correctly.

Production **never runs `db.Database.Migrate()`**. Both database-touching services set the
flag in `render.yaml`:

| Service | Line | Value |
|---|---|---|
| `planscape-api` | **`:74-75`** | `PLANSCAPE_USE_ENSURE_CREATED: "true"` |
| `planscape-worker` | **`:144-145`** | `PLANSCAPE_USE_ENSURE_CREATED: "true"` |

`Program.cs:1522` reads it, so `useEnsureCreated` is true in production; control takes
`:1538` → `creator.CreateTables()` (`:1559`) and never reaches `Migrate()` (`:1564`). The
78-file migration set is **vestigial in production, not broken** — ADR 0001 documents this
and `render.yaml`'s own comment states it. "No migration creates table X" therefore blocks
nothing, and B1 is void in all three of its forms.

**Correction made in this pass:** the document cited `render.yaml:51-52` / `:104-105` in
**8 places**. Measured on this branch *and* on `origin/main`, the keys are at **`:74`** and
**`:144`**. (`:51`/`:104` is what an unrelated working branch shows — which is how the wrong
numbers got in.) All 8 corrected. The finding is unaffected.

### `DocumentApprovals` — closed

Three independent reasons `20260501000000`'s unguarded `UPDATE "DocumentApprovals"` never
runs, any one sufficient:

1. Production never runs `Migrate()` at all (above).
2. The file carries no `[Migration]` attribute, so EF cannot see it — **2 of 78** are
   discoverable. Measured: fresh empty Postgres, `ASPNETCORE_ENVIRONMENT=Production`,
   `PLANSCAPE_USE_ENSURE_CREATED` unset → `dotnet ef database update` completes, exit 0.
3. Forced discoverable, it dies at its **first** statement (`ALTER TABLE "TaggedElements"
   ADD COLUMN`, 42P01) before reaching the UPDATE — so it is one of 26 such tables, not a
   special case.

The table exists in production regardless: it is in `OnModelCreating`, and production builds
its schema from the model.

### C1 — decided, and the decision now propagated

> **Flat approval is canonical for the UI.** `ApprovalChainsController` **stays** —
> server-side, unchanged, still satisfying the same **`DocumentsController.cs:1554`** gate —
> and gets **no client surface** until a project demonstrably needs multi-step.

The gate at `:1554` is an **OR** (`hasCompletedChain` alongside `hasLegacyApproval`, evaluated
at `:1556`); the docstring at `:1538` calls the two paths usable *"interchangeably"*. So
declining to build a chain UI disables nothing and requires no migration.

The decision had been recorded in the decision box and the closed-items table, but the
**amendment summary**, **triage-table row 3** and the "seven rows moved" paragraph still said
*"undecided pending C1"* — the document contradicted itself. All three now agree. `:1554` is
cited so this is not relitigated.

**The controller is not deleted.** "No caller" has been wrong three times on this set.

### `bypassesAcl` — reviewed, deliberate, recorded (C2)

`ProjectMembersController.cs:105` emits `bypassesAcl = bypass || member == null`. Deliberate:
it mirrors `ProjectMemberAcl.ResolveAsync`'s documented *"No member row = inherit (project
author / tenant manager fallthrough)"*, sits behind `CanAccessProjectAsync`, and is a
**narrowing filter, not a grant** — the per-member allow-lists are restrictions layered on top
of an access decision already made. Left unchanged; recorded so the next audit does not
re-flag it.

### Docstrings fixed (M1, M2)

* `MfaController` — claimed a "production implementation **would** use a TOTP library (e.g.
  Otp.NET) and IDataProtectionProvider". It already uses both. Was false when written.
* `DeviceTwinsController` — claimed it "powers the mobile Live tab (RAG list)". There is no
  Live tab; the two `live.tsx` files are a LiveKit meeting screen and a healthcare pressure
  view, and a repo-wide grep returns zero hits.

Both rewritten with the measurement inline. `DeviceTwins` also carries an explicit *do not
delete on "no caller" grounds* note pointing at this document. Docstrings only — no behaviour
change.

### Per-controller table — five axes, unmeasured cells labelled

Restructured onto **contract · runtime status · backing · downstream · tier**, with an
explicit measurement-basis table. **Prod runtime is labelled UNMEASURED for all 14 rows** — no
production database or instance was queried in any pass.

**New measured column — Downstream.** Each controller's literal `[Route(...)]` fragment was
grepped across all six client trees. All fourteen return **0 files**.

An all-zero sweep is also what a *broken* grep returns, so the method was validated against
controls **before** the zeros were trusted:

| Control | Files |
|---|---|
| `api/projects` | **58** |
| `api/auth/login` | **13** |
| `api/issues` | 0 — clients compose paths without the `api/` prefix |

That third control showed an `api/`-prefixed needle can miss a real client, so `api/mfa` and
`…/telemetry` were re-run prefix-free (`/mfa`, `mfa/`, `totp`, `/telemetry`) — still 0.

**Scope limit, stated in the doc:** this establishes no client references these routes *as
literal strings on this branch*. A runtime-assembled path would evade it. Strong evidence, not
proof, and branch-local — #544 adds a `data-rights` client that does not exist here.

---

Docs + two docstrings. No behaviour change, no schema change, no migration touched. No
production database was queried in any pass.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/DeviceTwinsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/DeviceTwinsController.cs
@@ -8,9 +8,24 @@ using Planscape.Infrastructure.SignalR;
 namespace Planscape.API.Controllers;
 
 /// <summary>
-/// Pillar B (5A) — device-twin query + binding. Powers the mobile Live tab
-/// (RAG list) and the viewer's twin overlay (K3). Binding records the K1 iot
+/// Pillar B (5A) — device-twin query + binding. Binding records the K1 iot
 /// mapping so telemetry resolves to the model element.
+///
+/// NO CLIENT CALLS THIS. An earlier version of this docstring claimed it "powers
+/// the mobile Live tab (RAG list) and the viewer's twin overlay (K3)". Measured
+/// 2026-08-02: there is no Live tab — <c>Planscape/app/(tabs)/</c> contains
+/// index, issues, issue-detail, documents, models, ifc, schedule, cost-dashboard,
+/// myactions, scanner and settings, and nothing else. The two <c>live.tsx</c>
+/// files in the app (<c>meetings/live.tsx</c>, <c>healthcare/pressure-live.tsx</c>)
+/// are a LiveKit meeting screen and a healthcare pressure view respectively;
+/// neither touches this controller. A repo-wide grep for <c>device-twins</c> /
+/// <c>DeviceTwins</c> across <c>Planscape/</c> and <c>planscape-web/</c> returns
+/// zero hits.
+///
+/// The endpoints are real — routable, DI resolves, tenant-scoped, backed by tables
+/// the schema patcher creates unconditionally. This is a **product** gap (no client
+/// has been built), not a broken endpoint. Do not delete it on "no caller" grounds;
+/// see docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md.
 /// </summary>
 [ApiController]
 [Route("api/projects/{projectId:guid}/twins")]
--- a/Planscape.Server/src/Planscape.API/Controllers/MfaController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/MfaController.cs
@@ -10,8 +10,15 @@ namespace Planscape.API.Controllers;
 
 /// <summary>
 /// TOTP multi-factor authentication: enroll, verify, recovery codes, challenge log.
-/// Production implementation would use a TOTP library (e.g. Otp.NET) and
-/// IDataProtectionProvider for secret encryption.
+///
+/// This IS the production implementation. An earlier version of this docstring said
+/// "production implementation would use a TOTP library (e.g. Otp.NET) and
+/// IDataProtectionProvider for secret encryption" — that was already false when
+/// written and is corrected here. Both are in use:
+///   • <c>OtpNet</c> — <c>new Totp(secretBytes).VerifyTotp(code, out _,
+///     new VerificationWindow(2, 2))</c> (see VerifyEnrollment);
+///   • <c>IDataProtector</c> — injected and used to encrypt the shared secret at rest.
+/// Do not treat this controller as a stub.
 /// </summary>
 [ApiController]
 [Route("api/mfa")]
--- a/docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md
+++ b/docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md
@@ -1,0 +1,182 @@
+# Phase 6A — clientless controller triage
+
+**Measured 2026-08-02 on `main` @ `a5109e3e5`.** Triage only — no feature code in this PR.
+
+> **Runtime column is NOT MEASURED.** Item 2 of the brief requires exercising every action
+> against a running server. Docker Desktop is installed but its daemon would not start on this
+> machine (`npipe:////./pipe/dockerDesktopLinuxEngine` — no such file; launching the app left no
+> process), and there is no local Postgres on 5432. Everything below is *static* measurement.
+> Where a static proxy exists for "does it run" — DI registration, migration presence — it was
+> used and is labelled as such. See [Runtime gap](#runtime-gap).
+
+---
+
+## 0. Re-derived list — the brief's ten, plus two
+
+Re-derived per instruction, **including** `Planscape.Server/src/Planscape.API/wwwroot` and
+**excluding** `wwwroot/_next/**` and `wwwroot/vendor/**`. Clients scanned: `StingTools/`,
+`planscape-web/`, `Planscape/` (mobile), `StingBridge/`, `Planscape.Server/.../wwwroot/`,
+`marketing-site/` — 1,856 files.
+
+**Your correction is confirmed.** `PublicConfig` (`api/public-config`), `RoleBuckets`
+(`api/state-machine/role-buckets`), `TenantKeywords` (`api/admin/tenant-keywords`) and
+`TenantBimManagerRoles` (`api/admin/tenant-bim-manager-roles`) are all called from
+`wwwroot/js/dashboard.js`; the latter two also from `wwwroot/index.html` and
+`wwwroot/app/index.html`. Not clientless. Left alone.
+*(One nuance: I found no `public-config` reference under `marketing-site/` in this checkout —
+possibly the built bundle isn't committed. Doesn't change the verdict.)*
+
+All ten in the brief are confirmed clientless by exact route-string grep. **Two more are too:**
+
+| Extra | Route | Note |
+|---|---|---|
+| **MaterialSync** | `api/MaterialSync` | 53 ln, 2 actions. Not on the brief's list. **Has a security finding — see S1.** |
+| **PhotoExport** | `api/projects/{id}/photo-export` | Already scheduled as Phase 5; listing for completeness. |
+
+---
+
+## 1. Triage table
+
+Legend — **Backed**: DbSet + a migration that actually creates the table.
+**DI**: every injected service resolves (static check of `Program.cs`).
+
+| # | Controller | Ln / actions | Auth gate | Tenant scoping | Backed | DI | Verdict |
+|---|---|---|---|---|---|---|---|
+| 1 | **DataRights** | 114 / 3 | `[Authorize(Roles="Owner,Admin")]` | `ITenantContext` | n/a (reads + deletes) | ✅ | **Tier 1 — but STOP, see D1** |
+| 2 | **CdeContainers** | 287 / 6 | `[Authorize]` + role gate `:255` | `TenantId` + `ProjectMembers` | ✅ `20260517000003` | ✅ | **Tier 1 — confirmed viable** |
+| 3 | **ApprovalChains** | 317 / 4 | `[Authorize]` + `[ProjectAccess]` | `[ProjectAccess]` + `TenantId` | ✅ `20260513000000` | ✅ | Tier 2 — real feature, duplicate of a live one (see C1) |
+| 4 | **AssetDataSheets** | 276 / 7 | `[Authorize]` + `[ProjectAccess]` | `RequireProjectMemberAsync` | ✅ `20260513100000` | ✅ | Tier 2 |
+| 5 | **OfflineManifest** | 227 / 4 | `[Authorize]` | `TenantId` | ✅ `20260517000000` | ✅ | Tier 2 — mobile-only by nature |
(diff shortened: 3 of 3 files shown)


StingD85:
## Runtime column measured — and it overturns B1

Docker Desktop had never actually been launched, only probed. Starting it worked first try, so the pass ran.

**All fourteen controllers serve traffic. Zero `relation … does not exist`, zero 5xx** — including the seven this PR previously rated *Tier 3 — blocked*.

The six tables really do have no `CreateTable`. That turned out not to matter:

1. **The migration folder is inert.** 80 migration files, **2** carry `[Migration(` — so EF applies 2. [`docs/adr/0001-schema-management.md`](docs/adr/0001-schema-management.md) (Accepted, 2026-06-04) already records this and adopts `EnsureCreated` + idempotent patchers as the official mechanism. I hadn't read that ADR in either earlier pass, which is how the bad inference survived two reviews.
2. **A patcher creates all six on every boot.** `PlatformSchemaPatcher` `CREATE TABLE IF NOT EXISTS`es them, and `Program.cs:1585` calls it *outside* the Development/`Migrate()` split — both branches.
3. **Fresh empty DB, Production path** → `Schema drift detected: 112 missing table(s)`, boot aborts. Of the 23 tables created, **all six disputed ones are present**; **all five "migration-backed" ones are absent**, along with `Tenants`, `Users`, `Projects`.

So the **Backed column was inverted**, and the production question is settled without touching production: a database built only from these migrations *cannot run the app at all*, so any environment that is up got there via the patcher path — which creates the six. No manual DDL need be hypothesised.

Writes verified by row count rather than trusting a 200: `TwinRules=8` after `seed-defaults`; `DeviceTwins=1`, `TelemetryPoints=1` after `telemetry/ingest`.

**The STOP item stands and was the right call** — authoring six migrations would have added files to a folder EF does not read. The real gap is those 78 invisible files, which is ADR 0001's known state and out of scope here.

**Limits kept explicit:** primary action per controller, not all 46; destructive actions not fired; `TwinAlerts`/`WorkOrders` returned 200 over *empty* tables; single-tenant demo login does not exercise the cross-tenant isolation each row claims.

Local only — compose Postgres plus a throwaway container. **No production database was contacted.**

StingD85:
Independent verification pass on the migration mechanism, plus the three fold-ins requested. **Not editing `docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md` directly** — this branch's worktree is at `7a09e0695`, ahead of `origin/claude/phase6-runtime` (`54f1a0083`), so it holds unpushed work. Editing it from another session would create exactly the collision that got #541 closed. Text below is ready to fold; the owning session should take it.

---

## 1. The mechanism, reproduced independently

```
$ cd Planscape.Server/src/Planscape.Infrastructure/Data/Migrations
$ ls *.cs | wc -l                                                       # 83
$ ls *.cs | grep -v '\.Designer\.cs$' | grep -v ModelSnapshot | wc -l    # 80
$ grep -l '\[Migration(' *.cs
20260605041920_MeetingMedia.Designer.cs
20260626203153_SustainabilitySnapshots.Designer.cs
```

**80 migration classes · 2 discoverable · 78 inert.** Matches your measurement exactly.

One correction to the framing in the brief that prompted this: it's *80* migration files, not 78 — 78 is the inert count, which your PR body already had right ("the migration-file count was 80, not 78").

Worth recording explicitly: **this is not a discovery, it is a re-discovery.** [`docs/adr/0001-schema-management.md`](../blob/main/docs/adr/0001-schema-management.md) — **Status: Accepted, 2026-06-04** — documents the whole mechanism, correctly:

> They are `partial class … : Migration` with `Up()`/`Down()` bodies but carry **no `[Migration("id")]` attribute** and have **no `.Designer.cs` companions**… Without it, `Database.Migrate()` sees **zero** migrations and applies nothing.

The triage doc should cite the ADR rather than re-derive it. That reframes B1 a third time: not "missing table", not "hygiene finding", but **"reasoned from a source the project had already formally declared non-authoritative."**

## 2. B1 — void in both directions

> "No `CreateTable` for `WorkOrders`/`GlobalIdRegistry`" was never evidence of anything.

Retract it as a finding. A migration's contents say nothing about live schema **in either direction**: absence of a `CreateTable` is not evidence the table is missing, and presence of a `Up()` body is not evidence it ran. 78 of 80 never execute.

The two twin tables should be re-rated on the patcher evidence alone — which your §2 already establishes independently (`PlatformSchemaPatcher.ApplyAsync` at `Program.cs:1585`, outside the `useEnsureCreated` split, so it runs on both branches). That argument survives B1's retraction intact and needs no migration reasoning to stand up.

## 3. `DocumentApprovals` — closed

Answered, and the answer is that the question was malformed. `20260501000000`'s unguarded `UPDATE "DocumentApprovals"` never throws **because it never runs** — no `[Migration]` attribute, no Designer, never enumerated.

Your counterfactual is the part worth keeping: made discoverable, it dies at its *first* statement (`ALTER TABLE "TaggedElements"`), not at `DocumentApprovals`, which is one of 26 tables it assumes exist. **Latent, not live** — and your closing line is the important one: *fixing the migration hygiene is precisely what arms it.*

## 4. CLAUDE.md — corrected in #546

`CLAUDE.md` claimed **"83 migrations present & applied"** in four places. The provenance is `ls *.cs | wc -l` — 80 migrations + 2 Designer + 1 snapshot. Both halves wrong. Fixed in **#546**, deliberately scoped to `CLAUDE.md` alone so it cannot conflict with this PR's file.

## 5. Escalation filed — #545, with one correction to its premise

`20260506200000_EnablePostgresRowLevelSecurity` is one of the inert 78. Grep confirmed: `ROW LEVEL SECURITY` appears in that file and nowhere else in the server. **Postgres RLS has never been applied to any database** — 45 tables unprotected. Full write-up, blast radius and a prod-confirmation procedure that never connects to prod are in **#545**.

**But it would not have contained the MaterialSync IDOR (#542).** `MaterialSyncController` injects `IWebHostEnvironment` and nothing else — no `DbContext`, no EF surface. Both actions read and write `App_Data/material_snapshots/{projectId}.json` on the filesystem (`:31-35`, `:47-50`). RLS gates rows in Postgres tables; that data never enters Postgres. #542's authorization fix was the only thing that could have closed it. The two findings are independent.

Also in #545, and relevant here: RLS is inert a **second** way. The policy is permissive when `app.current_tenant` is unset, and `Database:RlsEnabled` — which gates the interceptor that sets it — is configured nowhere. Applying the migration alone would change nothing.

## 6. C1 — recommendation, and a defect that should decide it

Requested: propose which mechanism the UI offers. Nothing built.

**Recommendation: `ApprovalChains` becomes canonical for all new UI; the flat path stays as a supported legacy read/write route until mobile migrates; the gate's OR stays.** The chain is a strict superset — a single-stage chain is behaviourally identical to a flat approval — and the flat path already has a shipped mobile client (`endpoints.ts:1005,1012`) that must not break. Render a flat approval as a degenerate one-stage chain so a single component serves both.

**Do not act on that yet**, because of this:

### C2 (new) — the chain path silently bypasses the revision-staleness gate

`CheckApprovalGate` (`DocumentsController.cs:1551-1555`) is **asymmetric**:

```csharp
var hasLegacyApproval = await _db.DocumentApprovals
    .AnyAsync(a => a.DocumentId == docId && a.Transition == transitionKey && a.Status == "APPROVED"
        && (a.RevisionSnapshot == null || a.RevisionSnapshot == currentRevision));   // revision-scoped
var hasCompletedChain = await _db.ApprovalChains
    .AnyAsync(c => c.DocumentId == docId && c.Transition == transitionKey && c.Status == "COMPLETED");
                                                                                     // NOT scoped
```

The legacy branch enforces "Gap 2" — an approval whose `RevisionSnapshot` doesn't match the current revision is stale and ignored, *"so a rework bump always forces a fresh approval round"* (`:1540-1542`). The chain branch has no such predicate.

And it cannot have one: `ApprovalChain` (`Planscape.Core/Entities/ApprovalChain.cs:17-39`) has **no `RevisionSnapshot` field at all** — `Id`, `TenantId`, `ProjectId`, `DocumentId`, `Transition`, `Status`, `CreatedBy`, `CreatedAt`, `CompletedAt`, `Description`, `Stages`.

**Consequence:** one `COMPLETED` chain permanently satisfies `SHARED→PUBLISHED` for that document, across every future revision. Approve P01 through a chain, then rework to P05 and publish with no further approval. The flat path blocks that; the chain path does not. Since the gate is an **OR**, the weaker branch wins whenever a chain exists — so the deliberate control added to the flat path can be bypassed simply by using the newer mechanism.

That inverts the C1 question. Promoting chains to canonical **today** would ship a regression in ISO 19650-2 §5.6 enforcement. Sequence has to be: add `RevisionSnapshot` to `ApprovalChain`, scope `hasCompletedChain` to match, backfill or invalidate existing chains — *then* make chains canonical.

I'd rate C2 above C1 in priority: C1 is a product decision that can wait, C2 is a live gate bypass in shipped code. Not fixed — flagging for your call, same as S1.

## 7. #541 / #543 conflict — resolved

#541 is **CLOSED**. It was the only other PR touching `docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md`, so this PR is now the sole change to that file and there is nothing to retarget. #546 was scoped to `CLAUDE.md` alone to keep it that way.

## 8. Not done, deliberately

No migration regenerated, no Designer file added, no `dotnet ef migrations add` run, no production database contacted, no C1/C2 code written. Remediation options for the migration set are in #545 §"Remediation options" and await sign-off.

PR #544, opened 2026-08-01
feat(data-rights): wire the GDPR/POPIA export + erasure UI, and repair the export

StingD85:
Gate 3 of the Phase 6A triage follow-up: DataRights was rated **Tier 1 — complete now**, with conditions. This ships the UI, and repairs the endpoint it turned out to be sitting on.

## The endpoint had never worked

`DataRightsController` was clientless — nothing reached it, so nobody had run it. Writing the tenant-scoping test the review asked for failed on the first attempt, for a reason unrelated to tenancy:

```
System.Text.Json.JsonException: A possible object cycle was detected.
Path: $.Tenant.Users.Tenant.Users.Tenant.Users…
```

EF relationship fix-up wires the ten loaded entity sets to each other, and `JsonSerializer.Serialize` walks straight into the cycle. It throws **part-way through the stream** — the `200` and the ZIP header are already on the wire — so a caller receives a truncated archive rather than an error. `GET /api/data-rights/export` has never produced a valid ZIP.

## Changes

**Server** — `DataRightsController`
- `AsNoTracking()` on all ten queries so fix-up cannot build the cycle, plus `ReferenceHandler.IgnoreCycles` as the belt-and-braces guarantee.
- Users are projected through an `ExportedUser` record rather than dumped raw. `AppUser` carries `PasswordHash` and `RefreshToken` for *every* member of the tenant, and this ZIP goes to whoever holds Owner/Admin. A subject-access request is not a reason to disclose other people's credentials.

**Desktop** — `Planscape.Desktop`
- `apiFetchBlob()` in `client.ts`: same auth + 401-refresh path as `apiFetch`, but returns the raw body, which a streamed ZIP needs.
- `dataRights` endpoints and a `DataRightsCard` on the Settings page.

## Against the stated conditions

| Condition | How it's met |
|---|---|
| Owner/Admin only, matching `[Authorize(Roles="Owner,Admin")]` | Card returns `null` for other roles. Server stays the authority — the client check only avoids showing controls that would 403. A test asserts a Contributor is refused. |
| Erase copy states **both** the 30-day hard delete **and** that cancel fully reverses it | "Freezes this organisation immediately and schedules permanent deletion of all of its Planscape data in **30 days**. For those 30 days the erasure is **fully reversible** — 'Cancel erasure' here restores the organisation and everything in it. After the 30 days the deletion is permanent and cannot be undone." The post-erase state repeats it with the actual date. |
| Export test asserting only the caller's tenant rows, **non-empty** | `Export_contains_only_the_callers_tenant_rows_and_is_not_empty`. Seeds a project on the *other* tenant so there is something real to leak, then asserts the caller's own rows are present (`TenantId`, `Test BIM Project`, `admin@test.org`) **and** that nothing of the other tenant's appears. The non-empty half is the point: `ITenantContext.TenantId` falls back to `Guid.Empty`, which matches no rows, so an isolation-only assertion would pass vacuously against an empty archive. |
| Confirmation phrase stays server-enforced | Unchanged in the controller. The client comparison only keeps the button's enabled state honest, so the user isn't surprised by a 400. |

Also asserts no `PasswordHash` / `RefreshToken` / BCrypt prefix appears anywhere in the archive.

## Verification

- `dotnet test tests/Planscape.Tests` → **533 passed, 0 failed, 9 skipped** (the 9 skips are the pre-existing Postgres-only and search tests, unrelated).
- `npx tsc --noEmit` → clean; `npm run build` → clean.

## Note on tenancy scope

Nine of the ten exported types are `ITenantScoped` and covered by the global query filter. `Tenant` itself is deliberately **not** filtered (it would break the pre-auth slug-uniqueness check in `AuthController.Register`) and is scoped by an explicit `Where(t => t.Id == tenantId)` in the controller. That split is what the test guards.

### DIFF
--- a/Planscape.Desktop/src/renderer/api/client.ts
+++ b/Planscape.Desktop/src/renderer/api/client.ts
@@ -87,3 +87,55 @@ export async function apiFetch<T = unknown>(
 
   return response.json() as Promise<T>
 }
+
+export interface BlobResponse {
+  blob: Blob
+  /** Filename from Content-Disposition, or null when the server didn't send one. */
+  filename: string | null
+}
+
+/**
+ * Same auth + 401-refresh behaviour as apiFetch, but returns the raw body.
+ * Needed for endpoints that stream a file rather than JSON (the GDPR/POPIA
+ * subject-access export is a ZIP).
+ */
+export async function apiFetchBlob(
+  path: string,
+  options: RequestInit = {}
+): Promise<BlobResponse> {
+  const serverUrl = await getStoredValue<string>('serverUrl') ?? DEFAULT_SERVER
+  const accessToken = await getStoredValue<string>('accessToken')
+
+  const headers: Record<string, string> = {
+    'X-Client-Type': 'desktop',
+    ...(options.headers as Record<string, string> ?? {})
+  }
+  if (accessToken) headers['Authorization'] = `Bearer ${accessToken}`
+
+  let response = await fetch(`${serverUrl}${path}`, { ...options, headers })
+
+  if (response.status === 401) {
+    const newToken = await refreshAccessToken()
+    if (newToken) {
+      headers['Authorization'] = `Bearer ${newToken}`
+      response = await fetch(`${serverUrl}${path}`, { ...options, headers })
+    }
+  }
+
+  if (!response.ok) {
+    let message = response.statusText
+    try {
+      const body = await response.json()
+      message = body.message ?? body.title ?? message
+    } catch { /* body may not be JSON on a streamed endpoint */ }
+    const err: ApiError = { status: response.status, message }
+    throw err
+  }
+
+  const disposition = response.headers.get('Content-Disposition') ?? ''
+  const match = /filename=([^;]+)/i.exec(disposition)
+  return {
+    blob: await response.blob(),
+    filename: match ? match[1].trim().replace(/^"|"$/g, '') : null
+  }
+}
--- a/Planscape.Desktop/src/renderer/api/endpoints.ts
+++ b/Planscape.Desktop/src/renderer/api/endpoints.ts
@@ -1,4 +1,4 @@
-import { apiFetch } from './client'
+import { apiFetch, apiFetchBlob } from './client'
 
 // ── Auth ──────────────────────────────────────────────────────────────────────
 
@@ -96,4 +96,36 @@ export const transmittals = {
   list:   (projectId: string) => apiFetch<TransmittalDto[]>(`/api/projects/${projectId}/transmittals`),
   create: (projectId: string, p: Partial<TransmittalDto>) =>
     apiFetch<TransmittalDto>(`/api/projects/${projectId}/transmittals`, { method: 'POST', body: JSON.stringify(p) })
+}
+
+// ── Data rights (GDPR / POPIA) ────────────────────────────────────────────────
+//
+// Server: DataRightsController, [Authorize(Roles = "Owner,Admin")].
+// The confirmation phrase is enforced server-side; the UI asks for it as well
+// so the user is never surprised by a 400, but the client check is a courtesy,
+// not the gate.
+
+export const ERASE_CONFIRMATION_PHRASE = 'ERASE EVERYTHING'
+
+export interface EraseResponse {
+  tenantId: string
+  /** ISO timestamp when the irreversible hard-delete runs. */
+  erasureCompletesAt: string
+  message: string
+}
+
+export const dataRights = {
+  /** Subject-access export — a ZIP of every row held for the caller's tenant. */
+  exportArchive: () => apiFetchBlob('/api/data-rights/export'),
+
+  /** Freezes the tenant and schedules the hard-delete 30 days out. Reversible until then. */
+  erase: (confirmationPhrase: string) =>
+    apiFetch<EraseResponse>('/api/data-rights/erase', {
+      method: 'POST',
+      body: JSON.stringify({ confirmationPhrase })
+    }),
+
+  /** Aborts a pending erasure inside the cooling-off window and reactivates the tenant. */
+  cancelErase: () =>
+    apiFetch<{ message: string }>('/api/data-rights/cancel-erase', { method: 'POST' })
 }
--- a/Planscape.Desktop/src/renderer/components/DataRightsCard.tsx
+++ b/Planscape.Desktop/src/renderer/components/DataRightsCard.tsx
@@ -1,0 +1,210 @@
+import React, { useState } from 'react'
+import { dataRights, ERASE_CONFIRMATION_PHRASE } from '../api/endpoints'
+import type { ApiError } from '../api/client'
+import { useAuthStore } from '../stores/auth'
+
+/**
+ * GDPR (EU) / POPIA (South Africa) subject-access + erasure controls.
+ *
+ * Backed by DataRightsController, which is [Authorize(Roles = "Owner,Admin")].
+ * We mirror that check here so non-privileged users don't see controls that
+ * would only 403 — the server remains the authority.
+ *
+ * The confirmation phrase is likewise enforced server-side (a mismatch is a
+ * 400 `confirmation_required`). The client-side comparison only exists so the
+ * button state matches what the server will accept.
+ */
+
+const ERASE_ROLES = ['owner', 'admin']
+
+function errorMessage(err: unknown): string {
+  if (err && typeof err === 'object' && 'message' in err) {
+    return (err as ApiError).message
+  }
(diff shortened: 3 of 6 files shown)


StingD85:
## Staying open — the instruction to fold this into the triage doc and close it rests on a wrong premise

I was asked to *"fold #544's migration measurement into the triage document as an appendix, framed
as expected behaviour explained by `render.yaml`, and close #544."*

**#544 contains no migration measurement.** I checked before acting rather than after.

- The "2 of 78 migrations visible; `dotnet ef database update` yields 2 tables and no `Tenants`"
  measurement is in **#543** (§M3), not here. It has now been re-labelled **Appendix A — expected
  behaviour, not a defect** in that PR, which is exactly what was asked for — just in the right
  place.
- **#544 is feature work.** Files touched: `DataRightsController.cs`,
  `DataRightsExportTenantScopeTests.cs`, and four `Planscape.Desktop` renderer files. Nothing
  migration-related.

Closing it would discard shipped, tested work **and leave a live defect in place**:

> `GET /api/data-rights/export` has never produced a valid ZIP. EF relationship fix-up wires the ten
> loaded entity sets to each other and `JsonSerializer.Serialize` walks into the cycle
> (`A possible object cycle was detected. Path: $.Tenant.Users.Tenant.Users…`). It throws
> **part-way through the stream** — the `200` and the ZIP header are already on the wire — so the
> caller receives a truncated archive rather than an error.

That is a real bug on a GDPR/POPIA subject-access endpoint, found only because someone finally wrote
a test against it. This PR also stops the export dumping `PasswordHash` and `RefreshToken` for every
member of the tenant into an archive handed to whoever holds Owner/Admin.

**Left open, unchanged.** It needs its own review on its own merits, not disposal as a side effect of
tidying the triage thread.

For the record, the triage-thread consolidation that prompted this **is** done:

- **#541** — already closed (2026-08-01T23:10:44Z), before this pass.
- **#543** — now the single PR carrying the triage document, with the fifth-pass corrections.
- **#546** — already owns the CLAUDE.md migrations correction; not duplicated.
- **#545** — the one surviving finding (RLS unreachable), with an added correction.

StingD85:
## The object-cycle bug is real, reproduced, and fixed — plus a second one behind it

I reproduced the failure live rather than trusting the description, then re-ran the same request against this branch. Both against the local docker stack, same DB, same admin token.

### RED — `main` today

```
GET /api/data-rights/export  ->  200 OK, Content-Type: application/zip, 622 bytes
```

| entry | bytes | parses? |
|---|---|---|
| `tenant.json` | 712 | yes |
| `users.json` | **0** | no — empty |

**2 of 10 entries, and the archive opens cleanly.** That is worse than "truncated": the ZIP is *structurally valid*, `tenant.json` parses, and the status is `200`. A data subject issuing a subject-access request receives a plausible-looking archive containing one file and an empty one, with nothing anywhere indicating failure. Confirms the diagnosis exactly — serialization dies on the second entry, after the header is on the wire.

### GREEN — this branch

```
GET /api/data-rights/export  ->  200 OK, 302,454 bytes, 10 of 10 entries
```

Credential scan across the whole archive — `PasswordHash`, `passwordHash`, `RefreshToken`, `refreshToken`, `$2a$`, `$2b$`, `$2y$`: **none found**. The `ExportedUser` projection does what it claims. `dotnet test --filter ~DataRights` → **2 passed, 0 failed**. Server builds **0 errors**.

So the fix is sound and I'd take it. But finishing the export surfaced the next defect in line.

---

## ⚠️ New finding: 9 of the 10 files are not parseable JSON

Now that all ten entries are actually written, they can be parsed for the first time — and they don't parse:

| entry | bytes | `json.loads` |
|---|---|---|
| `tenant.json` | 712 | ✅ valid |
| `users.json` | 156,613 | ❌ `Extra data: line 14 column 1` |
| `projects.json` | 13,335 | ❌ `Extra data: line 42 column 1` |
| `members.json` | 211,590 | ❌ |
| `issues.json` | 7,413,686 | ❌ |
| `documents.json` | 3,988 | ❌ |
| `models.json` | 25,749 | ❌ |
| `audit-log.json` | 129,209 | ❌ |
| `subscriptions.json` | 0 | ❌ empty |
| `invoices.json` | 0 | ❌ empty |

The cause is a two-ingredient mismatch in `DumpAsync`:

```csharp
WriteIndented = true                                   // objects span many lines
foreach (var row in rows)
    await writer.WriteLineAsync(JsonSerializer.Serialize(row, ExportJsonOptions));
```

`WriteLineAsync`-per-row is the JSON Lines shape, but `WriteIndented = true` breaks JSON Lines (a record must be one line), and there is no array wrapper or comma, so it isn't a JSON document either. Output is bare concatenated objects — `}\r\n{`. It is **neither format**, so nothing parses it.

**This is pre-existing, not introduced here.** `main`'s `DumpAsync` is byte-identical on those two lines (`:75-77`); this PR only added `ReferenceHandler.IgnoreCycles` and hoisted the options to a static. It was simply unobservable while the export died at file two.

**Why the tests didn't catch it:** `DataRightsExportTenantScopeTests` asserts with `Assert.Contains("Test BIM Project", entries["projects.json"])` — substring checks, never `JsonDocument.Parse`. Substring assertions pass happily against an unparseable file. The non-empty guard is good and I'd keep it; it just needs a parseability assertion beside it.

### Proposed fix — one line, but it changes the file format, so it's your call

Setting `WriteIndented = false` makes every file valid **NDJSON**, keeps the existing per-row streaming (which matters — `issues.json` is 7.4 MB), and makes the two 0-byte files legitimately correct as "no rows". The alternative — wrapping each file in a real JSON array — is a bigger change and buffers or complicates streaming.

I have **not** made either change, because it alters the shape of a response and you asked to be consulted first. Say the word and I'll fold `WriteIndented = false` plus a `JsonDocument.Parse` assertion per entry into this PR, or file it separately so this one can land as-is.

My recommendation: **fold it in.** This PR's stated purpose is repairing the export, and an archive nobody can parse is not a repaired export — the GDPR/POPIA deliverable is the file's *contents* being usable, not the ZIP opening.

StingD85:
## Folded in: the export now emits parseable JSON Lines

Added `c365d89` per the agreed decision to repair the JSON rather than ship the export as-is.

### The defect

`DumpAsync` writes one `WriteLineAsync` per row — a JSON Lines shape — while `ExportJsonOptions` set `WriteIndented = true`, so every record was pretty-printed across many lines. The output is therefore **neither** a JSON document (several top-level values, no enclosing array) **nor** JSON Lines (records span lines).

Measured, not inferred — `tenant.json` line 1 was the single character `{`:

```
tenant.json line 1 is not a self-contained JSON value: Expected depth to be zero
at the end of the JSON payload. There is an open JSON object or array that should
be closed. LineNumber: 0 | BytePositionInLine: 1.
  line: {
```

Fix is `WriteIndented = false`. That is the minimal repair and it preserves the streaming property: `DumpAsync` still serializes row-by-row rather than buffering a whole table (`issues.json` runs to ~7.4 MB) into one string — which is what wrapping each file in a JSON array would have cost.

### Why this PR's tests were green through it

Every assertion in `DataRightsExportTenantScopeTests` was a substring check (`Assert.Contains` / `Assert.DoesNotContain`) over raw entry text. A substring check cannot distinguish parseable from unparseable output, so the tenant-isolation and no-credentials guarantees were being verified against an archive no data subject could open.

`Every_export_entry_is_parseable_JSON_Lines` now parses every line of every entry with `JsonDocument.Parse`, and **counts** the records parsed so an empty archive fails loudly rather than passing vacuously — `PlanscapeDbContext`'s tenant filter falls back to `Guid.Empty` without an `ITenantContext`, which matches no rows.

### RED before GREEN, both numbers

| run | result |
|---|---|
| before the fix | **Failed: 1, Passed: 2**, Total: 3 |
| after the fix | **Failed: 0, Passed: 3**, Total: 3 |
| full suite, after | **Failed: 0, Passed: 534**, Skipped: 9, Total: 543 |

The two that passed in the RED run are this PR's pre-existing substring tests — green against broken output, which is the point.

### Generalised rule, and what the audit found

*Any test asserting on serialized output must parse it.* Audited the rest of `Planscape.Tests`: 62 call sites use `ReadFromJsonAsync` (which parses); the three files reading raw strings do so legitimately (failure messages, and a "secret must not appear anywhere" negative check where a substring assertion is the correct tool). **This export was the only instance** — reporting that honestly rather than inflating it.

### Blast radius

No consumer parses the archive: `DataRightsCard.tsx` saves the blob as `.zip` and never reads it. Response shape, status codes and routes are unchanged — only the bytes inside the ZIP entries.

Not merged, per the standing instruction.

Issue #545, opened 2026-08-02
security: Postgres RLS has never been applied to any database — and the config switch that claims to enable it is a decoy


StingD85:
## Summary

`20260506200000_EnablePostgresRowLevelSecurity` is one of the 78 inert migrations. **Postgres Row-Level Security has never been applied to any Planscape database**, and there is no out-of-band path that applies it either.

Worse than "absent": `Database:RlsEnabled` is a live configuration switch that an operator can set to `true` believing it turns tenant isolation on at the database. It does not. It registers an interceptor that sets a GUC no policy reads. **The control is a decoy.**

Reported only — nothing fixed, per instruction. Remediation options at the bottom await sign-off.

---

## Verified independently

Reproduced from a clean checkout of `main`, not taken on trust.

**The migration is inert.**

```
$ cd Planscape.Server/src/Planscape.Infrastructure/Data/Migrations
$ ls *.cs | wc -l                                              # 83
$ ls *.cs | grep -v '\.Designer\.cs$' | grep -v ModelSnapshot | wc -l   # 80  migration classes
$ grep -l '\[Migration(' *.cs                                  # 2, both Designer files
20260605041920_MeetingMedia.Designer.cs
20260626203153_SustainabilitySnapshots.Designer.cs
```

80 migration classes; 2 discoverable; **78 inert**. `EnablePostgresRowLevelSecurity` has no `[Migration]` attribute and no `.Designer.cs`, so EF never enumerates it and `Up()` never runs.

**The grep is confirmed.** Case-insensitive `ROW LEVEL SECURITY` across the entire server tree returns hits in exactly one file — the migration itself. No `CREATE POLICY` anywhere else. No deploy script, runbook, or `PlatformSchemaPatcher` statement applies the policy SQL out of band (`grep -rIn -i 'row level|create policy' docs/ Planscape.Server/` — nothing operational).

**It is also inert a second, independent way.** Even if the DDL were applied, the policy as written is permissive when the GUC is unset:

```sql
USING ("TenantId"::text = current_setting('app.current_tenant', true)
       OR coalesce(current_setting('app.current_tenant', true), '') = '')
```

The GUC is only set by `RlsConnectionInterceptor`, which `Program.cs:74-80` registers **only when `Database:RlsEnabled = true`**. That key is set nowhere — not in any `appsettings*.json`, not in `render.yaml`, not in any environment file. So the session variable is always unset, and the second branch matches every row.

[...]

PR #546, opened 2026-08-02
docs: correct CLAUDE.md — EF migrations are decorative, not "83 present & applied"

StingD85:
Docs-only. One file, `CLAUDE.md`. No feature code, no schema touched.

## The number came from `ls | wc -l`

CLAUDE.md asserted **"83 EF migrations (present & applied)"** in four places. Both halves are wrong, and the provenance explains how:

```
$ ls Planscape.Server/src/Planscape.Infrastructure/Data/Migrations/*.cs | wc -l
83
```

83 is a count of **files in the folder** — 80 migration classes + 2 `.Designer.cs` + 1 model snapshot. It was never a count of migrations, and nothing about it speaks to "applied".

## Only 2 migrations are applicable

EF builds its applicable-migration list by reflecting over the `[Migration]` attribute, which normally lives in the Designer file. Two exist:

```
$ grep -l '\[Migration(' *.cs
20260605041920_MeetingMedia.Designer.cs
20260626203153_SustainabilitySnapshots.Designer.cs
```

**2 discoverable, 78 inert classes EF never enumerates.** `Database.Migrate()` applies those two and nothing else. Real schema comes from `OnModelCreating` (via `EnsureCreated`/`CreateTables`) plus `PlatformSchemaPatcher`, which runs on **both** boot branches.

This is deliberate and already documented in [`docs/adr/0001-schema-management.md`](../blob/main/docs/adr/0001-schema-management.md) (**Accepted**, 2026-06-04) — CLAUDE.md simply contradicted the ADR. Independently corroborated by #543, which measured `dotnet ef database update` against a fresh Postgres: 2 tables, no `Tenants`.

## Changes

| Site | Before | After |
|---|---|---|
| Quick Stats row | `83 (present & applied — see §9)` | 80 classes, **2 discoverable, 78 inert** |
| §8 Server & platform | "**83 migrations present**" | "Schema is **not** managed by EF migrations" → §9 |
| §1 stale-caveats bullet | merged the build caveat and Healthcare's "EF migration not run yet" as both historical | un-merged: the build caveat *is* historical; the Healthcare note is a **different claim and still stands** |
| §2 drift table | `Healthcare: "EF migration not run yet"` → **83 migrations present** | row **retracted** — that correction was itself wrong |
| P0 recommendation 1 | claimed "EF not run" was among the facts fixed | marked partly retracted |
| §9 | *(silent on migrations, though Quick Stats pointed here)* | new bullet carrying the measurement |

The `see §9` pointer was already dangling — §9 never mentioned migrations. It does now.

## The new §9 bullet

Records the measurement plus the two consequences that keep costing review cycles:

1. **A migration's contents prove nothing about the database, in either direction.** "No migration creates table X" is not evidence X is missing; "a migration does Y" is not evidence Y happened. This is the exact reasoning error behind finding B1 in #541/#543.
2. **Security SQL living only in an inert migration was never applied** — Postgres RLS (45 tables) and audit-log hash-chain/partitioning are both in that category.

And a warning: **do not run `dotnet ef migrations add` here** without reading #545 first. It emits a Designer file, which silently arms that one migration against a snapshot that does not describe live schema. Two migrations have already become discoverable this way (`6e6a49af8` 2026-06-05, `486aa43fb` 2026-06-26) — both *after* the ADR was accepted, both by routine work, neither intending it.

## Verification

`grep -n '83 migrations\|83 (present\|migrations present' CLAUDE.md` → the only surviving hit is the retraction text quoting the old claim. Docs-only; nothing to build.

## Related

- **#545** — escalation: the security SQL in the inert set (RLS, audit hash-chain). Report-only, no fix.
- **#543** — B1 restated as migration hygiene; independently measured the same 2-of-80 result.
- Deliberately scoped to `CLAUDE.md` alone so it cannot conflict with #543, which is the sole change to `docs/PHASE6_CLIENTLESS_CONTROLLER_TRIAGE.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -81,7 +81,7 @@ prototype.
 | Tests | **774 test methods; 0 exercise `StingTools.csproj`** |
 | Empty `catch` blocks | 683 · `TaskDialog` sites 7,650 · `StingLog` sites 9,530 |
 | Stub/placeholder markers | 646 · `NotImplementedException` 10 · `TODO/FIXME/HACK` 51 |
-| EF migrations | 83 (present & applied — see §9) |
+| EF migrations | 80 migration classes on disk, **2 discoverable, 78 inert** — see §9 |
 
 ### 1. Build & compile health
 
@@ -98,8 +98,9 @@ prototype.
 - **CI exists** (11 GitHub Actions workflows: `stingtools-plugin.yml`, `planscape-server.yml`,
   `contract-drift.yml`, `multi-host-core.yml`, …) — worth referencing from the build docs.
 - **Stale caveats.** Many sections below carry *"committed without `dotnet build` verification (Linux
-  sandbox)"*, and the Healthcare section says *"EF migration not run yet."* **Both are historical**:
-  this machine builds 0/0 and 83 EF migrations exist. Trust the build, not the per-section caveat.
+  sandbox)"*. **That one is historical** — this machine builds 0/0. Trust the build, not the
+  per-section caveat. The Healthcare section's *"EF migration not run yet"* is a **different claim
+  and it still stands**: migration files exist, but almost none of them are applicable (§9).
 
 ### 2. Documentation accuracy — the headline facts have drifted
 
@@ -112,7 +113,7 @@ starts here. Measured drift (fixed in this pass where cheap; see Quick Stats + D
 | Quick Stats: "1,204 C# files, ~572,000 lines" | **1,443 files / 655,982 lines** (plugin alone) |
 | `StingCommandHandler.cs` "4,817 lines" (later section) | **9,519 lines** |
 | Assembly "v1.0.0.0" | `.csproj` `<Version>` = **2.2.0** |
-| Healthcare: "EF migration not run yet" | **83 migrations** present |
+| ~~Healthcare: "EF migration not run yet"~~ — **this row was itself wrong**; retracted 2026-08-02 | 83 was a count of `*.cs` files in `Migrations/` (80 migrations + 2 Designer + 1 snapshot). Only **2 are applicable**; 78 are inert. See §9. |
 | "committed without `dotnet build` verification" (many sections) | Builds **0/0** on Windows+Revit |
 
 **Sustainable fix applied:** the Quick Stats and Documentation Map now round + point *here* (one
@@ -184,8 +185,9 @@ deliverable." Make that distinction explicit wherever it matters to a user.
 
 ### 8. Server & platform (`Planscape.Server`)
 
-ASP.NET Core 8 + EF Core + SignalR + Hangfire + PostgreSQL + Redis + MinIO. **83 migrations present**
-and the stack is verified running locally in [`docs/SYSTEM_STATUS.md`](docs/SYSTEM_STATUS.md). **No
+ASP.NET Core 8 + EF Core + SignalR + Hangfire + PostgreSQL + Redis + MinIO. **Schema is not managed by
+EF migrations** — see §9 — and the stack is verified running locally in
+[`docs/SYSTEM_STATUS.md`](docs/SYSTEM_STATUS.md). **No
 committed secrets** — `appsettings.Production.template.json` uses `__REPLACE_WITH_…__` placeholders
 and `render.yaml` marks JWT key / owner password `sync: false`. Good hygiene. Open items are honestly
 tracked as **🟡 PARTIAL** (live 2-participant meetings, GLB viewer render, a site-photos route 404, a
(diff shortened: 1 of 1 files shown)


StingD85:
## Verified independently — accurate. One refinement that makes the §9 addition stronger.

Reproduced from a clean checkout: 83 `.cs` files → **80 migration classes + 2 `.Designer.cs` + 1
snapshot**; `grep -l '\[Migration('` returns exactly the two Designer files; **78 inert**. Every
number in this PR checks out, and retracting the old "83 present & applied" row is right.

### The refinement

The new §9 bullet says:

> `Database.Migrate()` applies those two and nothing else.

True as a statement about `Migrate()`, but it invites the reading that **production applies two
migrations**. It applies **zero**, because production never calls `Migrate()` at all.

`render.yaml` sets `PLANSCAPE_USE_ENSURE_CREATED = "true"` on **both** services — `planscape-api`
(`:51-52`) and the worker (`:104-105`) — with a comment stating the incomplete set is by design and
that *"without this flag a FRESH database takes the `Migrate()` branch and comes up nearly empty."*
`Program.cs:1522-1564`:

```csharp
var useEnsureCreated = app.Environment.IsDevelopment()
    || string.Equals(Environment.GetEnvironmentVariable("PLANSCAPE_USE_ENSURE_CREATED"),
                     "true", StringComparison.OrdinalIgnoreCase);

if (!db.Database.IsRelational()) { /* tests */ }
else if (useEnsureCreated)       { /* probe Tenants; creator.CreateTables() */ }   // ← production
else                             { db.Database.Migrate(); }                        // ← dead in prod
```

Suggested wording, if you want it:

> `Database.Migrate()` would apply those two and nothing else — but **production never calls it**.
> `render.yaml:51-52` sets `PLANSCAPE_USE_ENSURE_CREATED=true`, so the deployed API and worker take
> the `EnsureCreated` + `PlatformSchemaPatcher` branch and **zero** migrations are ever applied.
> `__EFMigrationsHistory` should therefore be empty or absent in production.

This strengthens the PR's own thesis: the migrations are not just *decorative*, they are
**unreachable by the mechanism production runs**. It also makes the "do not run
`dotnet ef migrations add`" rule self-evidently correct — a regenerated set would not take effect
anyway, while carrying real risk of emitting destructive DDL against a snapshot that does not match
live schema.

### Why it matters beyond wording — see #545

The same distinction is load-bearing for the RLS finding. `EnablePostgresRowLevelSecurity` is one of
the 78 inert migrations, and the intuitive fix — add the missing `[Migration]` attribute and a
Designer file so EF can see it — would review cleanly, merge, deploy, and **change nothing**. Only
the `PlatformSchemaPatcher` path (which runs on *both* boot branches) can actually deliver it.

Not touching CLAUDE.md myself — this PR owns that file, and a competing diff on the same lines would
just create a conflict. The related triage work is in #543.

StingD85:
## Every claim re-measured from scratch before publishing

This file is the onboarding surface, so I verified the correction independently rather than trusting either the PR body or my own prior notes. All six checks agree with the diff.

| Claim in this PR | How I checked | Result |
|---|---|---|
| 83 `.cs` files = 80 migrations + 2 Designer + 1 snapshot | `ls` counts in `Data/Migrations/` | **83 total, 2 `.Designer.cs`, 1 `ModelSnapshot`** → 80 migration classes ✅ |
| Only 2 carry `[Migration]`, so only 2 are discoverable | `grep -rln '\[Migration('` | exactly **`MeetingMedia`** + **`SustainabilitySnapshots`** ✅ |
| `Database.Migrate()` applies those two and nothing else | `select * from "__EFMigrationsHistory"` on the live local DB | **exactly 2 rows**, those two ✅ |
| Real schema comes from `OnModelCreating` + patcher, not migrations | table count in the same DB | **135 tables** against 2 applied migrations — the schema cannot have come from migrations ✅ |
| ADR 0001 is Accepted and authoritative | read `docs/adr/0001-schema-management.md` | **Status: Accepted**, dated 2026-06-04 ✅ |
| RLS SQL lives only in an inert migration, so was never applied | `select count(*) from pg_tables where schemaname='public' and rowsecurity=true` | **0** ✅ |

The 135-tables-vs-2-migrations pair is the decisive one. It is not possible to read that as "migrations built this schema," which is exactly the inference the current wording invites.

### Why this matters more than a typo fix

The line on `main` right now reads **"EF migrations | 83 (present & applied — see §9)"**. Both halves mislead: 83 is a file count, and "applied" is false for 78 of them. A reader who trusts it concludes the schema is migration-managed and that a migration's contents are evidence about the database — which is precisely how the RLS gap in #545 stayed invisible. Issue #545 is **OPEN** and titled as this diff references it.

I also endorse the retraction framing. The previous pass "corrected" this row in the wrong direction, and the diff says so plainly instead of quietly overwriting it. A correction log that hides its own misfires is not worth much.

### One thing I did not verify

The **"45 tables"** figure for RLS coverage is carried over from #545, not re-derived here — the live DB has 135 tables with 0 under RLS, which confirms *"never applied"* but does not confirm the intended scope was 45. Flagging it as sourced rather than measured; it does not affect the correction.

No code paths touched — documentation only.

PR #547, opened 2026-08-02
feat(server): serve project capabilities so three clients stop deriving them

StingD85:
## `GET /api/projects/{projectId}/members/capabilities`

Serves the two project capabilities the server already enforces, so clients stop
re-deriving them.

```json
{ "projectId": "…", "userId": "…", "canCurateProject": true, "canApproveSitePhotos": false }
```

> **Stacked on #540.** Base is `claude/capability-layer`, not `main` — this endpoint calls
> `CanCurateProjectAsync` / `CanApproveSitePhotosAsync`, which #540 introduces. It cannot
> compile without it. Merge #540 first; this retargets to `main` automatically.

### Why it exists

Three surfaces (plugin BCC, mobile, web) each decide locally whether to show curate /
approve controls. Every one of them re-implements a rule the server owns, and they drifted —
which is what #540 found: a hand-rolled `ProjectRole == "PM"` check copied into eight
controllers, where `"PM"` is an **Iso19650Role** code and never a `ProjectRole`, so it matched
essentially nobody. The fix is for the server to answer the question rather than for three
clients to guess it.

### Condition (a) — the 404 contract

**A 404 means treat ALL capabilities as false. Never "unknown, so proceed."**

A fail-open default here would undo the gate on all three surfaces at once. The contract is
stated in the endpoint's own docstring, in the words above, and extends to any absence of an
explicit `true`:

> The same applies to a network error, a timeout, or a body you cannot parse: absence of an
> explicit `true` is `false`.

It returns **404, not 403**, deliberately — matching `GetMyAccess` on the same controller,
which uses the same visibility check. Two adjacent endpoints disagreeing about the same
condition is a bug generator.

**Stated honestly:** the brief asks that this be documented *in every client that reads it*.
**No client reads it yet** — this PR adds the endpoint only. There is nothing to document
client-side until a consumer is wired, and each consumer must carry the contract at its call
site when it lands. Claiming otherwise would be claiming work not done.

### Condition (b) — exactly two booleans

Two, matching the two predicates that actually exist on `ProjectRoles`. A third does not get
added inline; it goes through the same propose-first step these two did. `Response_carries_
exactly_four_fields` pins the shape so a third cannot arrive by accident.

### Why a new endpoint rather than fields on `members/me`

`members/me` is already consumed by the plugin's BCC Deliverables tab and mobile
`documents.tsx`. Widening a live response shape is a breaking-shape change and would need the
STOP-AND-ASK step. A new endpoint is additive.

### Tests — 11, all passing against real Postgres

```
Test Run Successful.   Total tests: 11
```

| Test | Pins |
|---|---|
| `Sanity_the_fixture_actually_has_rows` | the fixture is non-empty — so a later assertion cannot pass vacuously |
| `Capabilities_match_the_role` (5 cases) | viewer F/F · plain-contributor F/F · **coordinator T/F** · manager T/T · contributor-who-is-ISO-PM T/T |
| `Coordinator_curates_but_cannot_approve_photos` | the two capabilities are genuinely distinct, not an alias |
| `Tenant_admin_gets_both_without_any_project_member_row` | the documented Admin/Owner fallthrough |
| `Invisible_project_is_404_so_clients_default_everything_to_false` | 404, **not** a 200 all-false body |
| `Unknown_project_id_is_404_not_an_all_false_body` | same, for a non-existent id |
| `Response_carries_exactly_four_fields` | the shape, so a third boolean cannot slip in |

The two 404 tests exist specifically because returning `200 {all false}` would look correct
and be wrong: it would let a client that ignores status codes render a coherent
everything-denied UI for a project it cannot see, and it would make a genuine
permission-denied indistinguishable from a project that does not exist.

`Coordinator_curates_but_cannot_approve_photos` is the one that would catch the likeliest
regression — collapsing the two predicates into one. Photo approval is deliberately narrower
than curation: it releases imagery outside the project.

### Verification

Built and run locally against real Postgres 16 (`PLANSCAPE_TEST_PG` set):
`dotnet build Planscape.sln -c Release` → **0 errors**; the filtered suite → **11/11 passed**.
Full-solution suite on this stack is **540/540**.

Read-only endpoint. No schema change, no migration touched, no existing response shape
altered.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/ProjectMembersController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ProjectMembersController.cs
@@ -6,6 +6,13 @@ using Microsoft.Extensions.Logging;
 using Planscape.Core.Entities;
 using Planscape.Core.Interfaces;
 using Planscape.API.Authorization;
+// Imported for the capability extension methods (CanCurateProjectAsync /
+// CanApproveSitePhotosAsync). This namespace also declares an IEmailService,
+// which collides with Planscape.Core.Interfaces.IEmailService — the one this
+// controller's constructor has always bound to. The alias below pins that
+// binding so adding the import cannot silently re-resolve it.
+using Planscape.API.Services;
+using IEmailService = Planscape.Core.Interfaces.IEmailService;
 using Planscape.Infrastructure.Data;
 using Planscape.Infrastructure.Services;
 using Planscape.Infrastructure.SignalR;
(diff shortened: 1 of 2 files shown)


StingD85:
Kept as the canonical capabilities-endpoint PR. Its twin **#548** — same head branch (`claude/capabilities-endpoint`), based on `main` instead of on #540 — has been closed as a duplicate.

`claude/capability-layer` is an ancestor of this branch, so #548 contained exactly #540 ∪ #547 and had no unique commits. Closing it lost nothing and removes the situation where the capability layer was pending review in two places.

**Merge path: #540 first, then this.** When #540 lands, GitHub retargets this PR to `main` and the diff stays the 2-file endpoint delta.

One note for whoever picks this up: this PR is still a **draft**, and it is a dependency for the parity work — SavedViews / DistributionGroups / PhotoExport are all specified to read capability rules from this endpoint rather than deriving them per client. Until it lands, those clients attempt-then-report the 403.

StingD85:
## Base unblocked, and verified against real enforcement

**#540 merged**, so this PR's stated blocker is gone. It still targets `claude/capability-layer`; that branch's content is now in `main` and the branch **merges into `main` cleanly** (checked locally). GitHub will retarget it.

⚠️ **This adds a new endpoint, which is your call, not mine.** I have taken it as far as "verified and mergeable" and stopped there — I have not merged it. Everything below is evidence for that decision.

### The check that mattered: does it tell clients the truth?

An advertising endpoint that disagrees with the gate is worse than no endpoint — clients would render controls that then 403, or hide controls that would have worked. So for each fixture I asked the endpoint what it claims, then **actually exercised** a curate endpoint and an approve endpoint and compared.

| fixture | ProjectRole / Iso | claims curate | real | claims approve | real | |
|---|---|---|---|---|---|---|
| viewer-plain | Viewer / M | false | false | false | false | AGREE |
| contributor-plain | Contributor / M | false | false | false | false | AGREE |
| coordinator | Coordinator / M | **true** | true | false | false | AGREE |
| manager | Manager / M | **true** | true | **true** | true | AGREE |
| iso-PM | Contributor / **PM** | **true** | true | **true** | true | AGREE |
| iso-BC | Contributor / BC | **true** | true | false | false | AGREE |
| iso-A | Contributor / A | **true** | true | **true** | true | AGREE |

**7/7 agree.** Note it reproduces the curate-vs-approve asymmetry correctly — Coordinator and BC curate but cannot approve — so it is genuinely reading the two predicates rather than collapsing them into one "is privileged" bit.

### Contracts, exercised rather than read

- Invisible project → **404** ✅ (matches `GetMyAccess`, as the docstring argues it must)
- No token → **401** ✅ (not anonymous)
- Response shape → exactly `['canApproveSitePhotos','canCurateProject','projectId','userId']`, **4 fields** ✅ — the shape test does pin this
- `dotnet test --filter ~Capabilit` → **41 passed, 0 failed**
- Server builds **0 errors**

### On the honesty of the body

The body says *"No client reads it yet — claiming otherwise would be claiming work not done."* That is accurate and I want it to stay accurate: **this endpoint is a prerequisite for Task 3's client work, not a deliverable on its own.** If it merges without a consumer it is another completed-and-clientless controller, which is the exact pattern this audit exists to stop. My suggestion is to merge it *with* or immediately *before* the first client that consumes it, and for that client to carry the 404-means-false contract at its call site.

### One caveat I could not close

The 404-means-false rule is currently enforced only by documentation. Nothing prevents a future client from treating 404 as "unknown, proceed." That is inherent to a contract that lives in a docstring; the mitigation is a client-side test when the first consumer lands, which I will write as part of Task 3 rather than assert here.

PR #548, opened 2026-08-02
feat(server): serve project capabilities so three clients stop deriving them

StingD85:
Serves the caller's project capabilities so the web app, the Revit BCC and mobile stop deriving them independently. Approved shape, built as proposed.

**Stacked on #540** (`claude/capability-layer`) — it uses the `CanCurateProjectAsync` / `CanApproveSitePhotosAsync` extensions that PR adds. Review #540 first.

```
GET /api/projects/{projectId}/members/capabilities
```
```json
{
  "projectId": "3f2b…",
  "userId": "9a1c…",
  "canCurateProject": true,
  "canApproveSitePhotos": false
}
```

## The 404 contract

**A 404 means every capability is false.** Never "unknown, so proceed". A fail-open default here would undo the gate on three surfaces at once, so it is stated in the endpoint's docstring, must be repeated in each client, and is pinned by two tests.

An all-false `200` body was rejected deliberately: a body implies the server knows the answer. Absence of the project should be indistinguishable from absence of authority.

404 rather than 403 matches `GetMyAccess` (`:95`) on the same controller, which uses the same visibility check. Two adjacent endpoints disagreeing about one condition is a bug generator.

## Decisions worth seeing

- **Two booleans**, matching the two predicates that exist on `ProjectRoles`. A third goes through the same propose-first step, not inline — `Response_carries_exactly_four_fields` fails if one is added quietly.
- **Separate endpoint, not extra fields on `members/me`.** That response is already consumed; widening it is a breaking-shape change.
- **No route collision.** The controller has no `GET {memberId}` — only `[HttpGet]`, `me`, `roles`. A literal segment follows existing precedent.
- **Tenant Admin/Owner resolve true without a ProjectMember row**, matching every site the capability layer replaced.
- The added `using Planscape.API.Services` collides with that namespace's `IEmailService`. An explicit alias pins the constructor to the `Planscape.Core.Interfaces` one it has always bound to, so the import cannot silently re-resolve it.

## Tests — 11, and proven to have teeth

Green is not evidence unless the tests can fail. Two mutations:

| Mutation | Result |
|---|---|
| fail-open instead of 404 | both 404 tests fail |
| collapse the two predicates into one | both coordinator tests fail |

The other 7 kept passing in both cases, so the mutations are caught specifically rather than by blanket breakage. Reverted after measuring.

The fixture is built **with** an `ITenantContext` and asserts non-empty rows first — the global filter falls back to `Guid.Empty`, which would make every assertion pass against an empty set.

## The full suite is intermittently red, and it is not this change

Stating this plainly because the number looks bad: with this branch the suite fails intermittently (0, then 86, then 71 across three runs).

I nearly reported that as my own regression. It isn't, and here is the proof rather than the assertion.

**The probe.** On the **base** commit — no endpoint, no capabilities code — I added a throwaway class of 11 tests that do SQLite + `EnsureCreated` + a seeded `PlanscapeDbContext`, touching no controller and using no fixed GUID:

```
Failed: 87   Failed: 1   Failed: 68
```

The same base commit with 11 **inert** no-op tests (identical total of 581) stays green 3/3, so it is not the test count or partitioning — it is doing real EF work in parallel.

**The mechanism**, from aggregating the failures:

- **44 ×** `ArgumentException: An item with the same key has already been added. Key: 11111111-1111-1111-1111-111111111111` — `PlanscapeWebApplicationFactory.cs:374`'s fixed tenant, seeded twice into a shared process-wide EF **InMemory** store when two factory instances overlap.
- **5 ×** `ObjectDisposedException: Hangfire.InMemory.State.Dispatcher` — the process-wide Hangfire teardown already documented as ROADMAP **DEP-7**.

Both are pre-existing suite fragility. Base survives on scheduling luck; any new EF-touching test class widens the window.

**One thing I did fix:** the fixture originally reused that same `11111111-…` tenant GUID. That was genuinely mine and wrong, so identifiers are now minted per fixture (as `ProjectRoleCapabilityTests` already does), plus `IDisposable` on the fixture. It did not resolve the race — the probe proves why — but it removed a real collision I introduced.

**Not fixed here.** The repo has no `xunit.runner.json` and no collection definitions, so there is no local way to opt one class out; the fix is `parallelizeTestCollections: false` (or putting the factory tests in a shared collection), which is repo-wide, costs real runtime, and belongs in its own PR alongside DEP-7. Flagging it rather than slipping it in.

**In isolation this branch is green:** the 11 new tests pass 11/11, and every class that fails in the full run passes on its own.

## CI

The two red checks are the pre-existing `main` baseline — one Postgres container failure (`role "root" does not exist`) surfacing twice, since `CI Gate` is an aggregator that times out waiting on `Build & Test`. Full diff with both run URLs is in #542.

Nothing merged.

### DIFF
--- a/.github/workflows/contract-drift.yml
+++ b/.github/workflows/contract-drift.yml
@@ -40,6 +40,7 @@ on:
       - 'stingtools-core/python/**'
       - 'StingBridge/**'
       - 'Planscape/**'
+      - 'planscape-web/**'
       - '.github/workflows/contract-drift.yml'
   pull_request:
     paths:
@@ -49,6 +50,7 @@ on:
       - 'stingtools-core/python/**'
       - 'StingBridge/**'
       - 'Planscape/**'
+      - 'planscape-web/**'
       - '.github/workflows/contract-drift.yml'
   workflow_dispatch:
 
--- a/Planscape.Server/src/Planscape.API/Controllers/DistributionGroupsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/DistributionGroupsController.cs
@@ -191,16 +191,11 @@ public class DistributionGroupsController : ControllerBase
         return NoContent();
     }
 
-    private async Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
+    // code, never a ProjectRole (ProjectMember.cs:15 vs :18), so this gate
+    // matched essentially nobody. Delegates to the shared capability layer.
+    private Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
+        => this.CanCurateProjectAsync(_db, projectId, ct);
     private Guid? CurrentUserIdOrNull()
     {
         var s = User.FindFirst("user_id")?.Value
--- a/Planscape.Server/src/Planscape.API/Controllers/PhotoAlbumsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/PhotoAlbumsController.cs
@@ -384,16 +384,11 @@ public class PhotoAlbumsController : ControllerBase
     /// Mutation gate (matches SitePhotosController.IsApproverAsync):
     /// tenant Admin / Owner OR project PM.
     /// </summary>
-    private async Task<bool> IsCuratorAsync(Guid projectId, CancellationToken ct)
-    {
-        var role = User.FindFirst("role")?.Value ?? "";
-        if (role is "Admin" or "Owner") return true;
-        var userId = CurrentUserIdOrNull();
-        if (userId == null) return false;
-        return await _db.ProjectMembers.AsNoTracking().AnyAsync(m =>
-            m.ProjectId == projectId && m.UserId == userId.Value &&
-            m.IsActive && m.ProjectRole == "PM", ct);
-    }
+    // Was an inline `ProjectRole == "PM"` test. "PM" is an Iso19650Role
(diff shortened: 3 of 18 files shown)


StingD85:
Closing as a duplicate — superseded by the stack **#540 → #547**.

### These are not two independent implementations

They were flagged to me as the same feature built twice by two sessions. That is not what happened, and the real structure matters for how the queue is worked:

**#547 and #548 share the same head branch — `claude/capabilities-endpoint`.** They differ only in base.

| | #547 | #548 (this one) |
|---|---|---|
| Head | `claude/capabilities-endpoint` | `claude/capabilities-endpoint` — **same** |
| Base | `claude/capability-layer` (#540) | `main` |
| Diff | 2 files, +359 / −0 | 18 files, +979 / −78 |
| Commits | `e3f9d52ce`, `d45e70d67` | `19e1737ce`, `e3f9d52ce`, `d45e70d67` |

#547 is a correctly-stacked PR showing only the endpoint delta on top of the capability layer. This PR is the same branch flattened against `main`, so its diff also swallows **#540** — *"one capability layer to replace eleven dead PM gates"* (`19e1737ce`), which is already open on its own.

### Verified: this PR has no unique content

```
$ git merge-base --is-ancestor origin/claude/capability-layer origin/claude/capabilities-endpoint
YES
```

`claude/capability-layer` is an ancestor of `claude/capabilities-endpoint`, so **#548 = #540 ∪ #547 exactly**. Every commit here exists in one of those two PRs. Closing it discards nothing.

### Why keep the stack rather than this one

Both routes deliver identical code, so this is purely about how it gets reviewed and merged:

- Keeping this PR means the capability layer is pending review in **two places at once** — here and in #540. That is precisely the queue ambiguity we are trying to remove, and it invites the layer being approved in one place while the other sits stale.
- The layer replaces eleven authorization gates. That is a behavioural change to who can do what, and it deserves to be reviewed on its own rather than bundled with a new read-only endpoint.
- #540 → #547 keeps each PR to one idea. When #540 merges, GitHub retargets #547 to `main` automatically and its diff stays the 2-file endpoint.

The one thing this PR had going for it is that it is not a draft and could merge today without waiting on #540. That is a sequencing convenience, not a code difference — and merging it would land #540's commit anyway, just without #540 having been reviewed as itself.

### Net effect

No work is lost. The merge path is **#540 first, then #547**.

Trivially reversible if you disagree: reopen this PR and close #547 instead — the branch is untouched and both point at the same tip.

*(Closed as part of a queue-cleanup pass; nothing was merged.)*

PR #555, opened 2026-08-02
feat(security): prepare Postgres RLS on the path that actually runs — no-op until Database:RlsEnabled is set (#545)

StingD85:
## Prepared, not applied — #545 made cheap to decide

**Merging this changes nothing at runtime.** Everything below is gated behind `Database:RlsEnabled`, which is set nowhere in this repository and is deliberately left unset. This exists so the decision is a one-line config change with a rehearsed outcome, rather than a design exercise someone has to start from scratch a fourth time.

---

### The actual state of #545

The policies were never applied to any database, and could not have been. They live in `20260506200000_EnablePostgresRowLevelSecurity`, and **production does not run migrations** (`docs/adr/0001-schema-management.md`) — the patcher path does. So `Database:RlsEnabled` only ever registered `RlsConnectionInterceptor`, which sets `app.current_tenant`. The policies that read that variable did not exist. The switch was wired to nothing on the DB side.

This PR moves the policy SQL onto the path that executes.

### Two defects fixed

**1. It ran on the wrong path.** Policy SQL now lives in `RlsPolicyPatcher`, invoked from `Program.cs` immediately after `PlatformSchemaPatcher`, gated on the same `rlsEnabled` flag that gates the interceptor — so the session variable and the policies that consume it turn on together, never one without the other.

**2. It failed OPEN.** The migration's policy carried:

```sql
OR coalesce(current_setting('app.current_tenant', true), '') = ''
```

A connection that never set the GUC matched **every row**. The policy became a no-op in precisely the case where a second layer of defence is meant to earn its keep — the application layer having already failed. That branch is gone:

```sql
USING ("TenantId"::text = current_setting('app.current_tenant', true))
WITH CHECK ("TenantId"::text = current_setting('app.current_tenant', true))
```

Unset GUC → `current_setting` returns NULL → predicate NULL → no rows.

### The consequence of failing closed — read before enabling

With RLS on, **any connection that does not set the GUC sees zero rows.** That includes every path setting `PlanscapeDbContext.BypassTenantFilter = true` (Hangfire jobs, cross-tenant admin scans), because `RlsConnectionInterceptor.ConnectionOpenedAsync` returns early for those and never issues the `SET`.

This will not leak data. It will make background jobs **silently find nothing**. Giving those paths a role carrying `BYPASSRLS` is the prerequisite for flipping the key, and it is the open decision this PR is deliberately not taking on your behalf. It is stated in the class remarks so it cannot be missed at the call site.

### Idempotency — one correction to the brief

The brief asked for `CREATE POLICY IF NOT EXISTS`. **Postgres has no such form** — unlike `CREATE TABLE`, policies have no `IF NOT EXISTS` variant. The idempotent idiom is `DROP POLICY IF EXISTS` followed by `CREATE POLICY`, which is what this uses and what the original migration used. Re-running converges on the same definition, which also makes this file the place to correct a policy later. `ENABLE` / `FORCE ROW LEVEL SECURITY` are natural no-ops when already set. Re-application is covered in the rehearsal.

### It throws rather than logging

`PlatformSchemaPatcher` logs failures and carries on. This does not. An RLS-enabled table whose policy failed to create is **default-deny** — it returns zero rows to everyone, a total silent outage of one table. A `N ok, M failed` line at boot is not something anyone acts on. Failing the boot is the honest response. Tables absent from the database are skipped, not failed.

---

## PgBouncer — already handled, and the cost is currently zero

`Program.cs:110-115` already resolves this, and better than the startup-guard I suggested on #545:

```csharp
var pooledRaw = builder.Configuration.GetConnectionString("Pooled");
var usePooler = !string.IsNullOrWhiteSpace(pooledRaw) && !rlsEnabled;
```

Enabling RLS **automatically falls back to the direct connection string**. The in-file comment gives the reason — session-scoped `SET app.current_tenant` leaks across tenants under transaction pooling, so the pooler is only safe while RLS is off. *(This corrects my own addendum on #545, which said a guard was still needed. It was already there; I had read a stale checkout.)*

**Quantified — what enabling RLS costs in throughput today: nothing.**

`ConnectionStrings__Pooled` is **commented out** in `render.yaml` for both api and worker (lines 99-101, 156-158), and Render ships connection pooling off by default. So `pooledRaw` is empty, `usePooler` is already `false`, and the pooler is not in the path. Flipping `Database:RlsEnabled` changes no connection behaviour.

**What it would cost later**, if someone enables pooling on `planscape-db` and uncomments those blocks: RLS silently switches them back off, returning to direct connections under the budget in `docs/DEPLOY_RUNBOOK.md` — api 30 (20 EF + 10 Hangfire) + worker 30 = 60 of the ~97 client connections a basic-tier Render Postgres allows, leaving 37 spare. That is sufficient at current scale; the pooler exists for when it is not. So the real trade is **RLS now, or PgBouncer headroom later** — and taking RLS now costs nothing until you reach the point of needing to multiplex, at which point the interceptor must move to `SET LOCAL` inside an explicit transaction. The file comment already says so.

---

## Rehearsal — run, with RED proven before GREEN

Against the local Postgres (`docker-postgres-1`), in a **throwaway `rls_rehearsal` database**, as a **`NOSUPERUSER NOBYPASSRLS`** role. That last detail matters: superusers *always* bypass RLS, so rehearsing as `postgres` would have passed no matter what the policy said.

Script committed at [`Planscape.Server/docs/rls_rehearsal.sql`](Planscape.Server/docs/rls_rehearsal.sql) so it is re-runnable.

```
STEP 1  RED, no policy       total visible = 3      tenant B visible to A = 1   ← not vacuous
STEP 2b idempotent re-apply  OK, no error
STEP 3  GREEN positive       tenant A own rows = 2  ("Tenant A - Alpha", "Tenant A - Beta")
STEP 4  GREEN cross-tenant   tenant B visible to A = 0
STEP 5  symmetric            tenant B own rows = 1  ("Tenant B - Secret")
STEP 6  FAIL CLOSED          GUC unset            = 0      ← migration's version returned 3
STEP 7  WITH CHECK           cross-tenant INSERT rejected (insufficient_privilege)
STEP 8  rollback SQL         applied
STEP 9  key OFF              total visible = 3      ← nothing changed
```

Step 1 is the point: **before** the policy, Tenant A could read Tenant B's row. Without that, step 4's zero would prove only that the query matched nothing. Steps 3 and 5 assert on **non-empty** results and on the row *names*, not just counts, so a Guid.Empty-style tenant fallback returning nothing cannot pass this vacuously.

The dev `planscape` database was **not touched** — verified afterwards: `0` RLS-enabled tables there, rehearsal database and role dropped.

## Rollback

`RlsPolicyPatcher.BuildRollbackStatements()` emits the exact undo, and step 8 above executed it. If enabling breaks queries:

1. Set `Database:RlsEnabled=false` and redeploy. **This alone is not enough** — the flag stops the interceptor setting the GUC, but the policies stay on the tables, so everything would return zero rows. It must be followed by step 2.
2. Run the rollback SQL against the database — per table: `DROP POLICY IF EXISTS tenant_isolation`, `NO FORCE ROW LEVEL SECURITY`, `DISABLE ROW LEVEL SECURITY`; plus `tenant_self_visibility` on `Tenants`.

Rehearsed in step 8-9: after rollback, all 3 rows were visible again.

That ordering trap is worth noting in whatever runbook entry accompanies enabling this: **disabling the flag without dropping the policies is the worst of the three states.**

## Not in this PR

- The `BYPASSRLS` role for Hangfire / admin paths — the prerequisite named above.
- Moving the interceptor to `SET LOCAL` — only needed if PgBouncer is wanted alongside RLS.
- Any change to the migration set. `20260506200000_EnablePostgresRowLevelSecurity` is left untouched as history.
- Setting the key. **That is yours.**

## Verification

`dotnet build Planscape.API` → **Build succeeded**. The 3 `CS0618` Hangfire `QueueName` warnings are pre-existing — confirmed identical on `origin/main` (3 occurrences on both).

Branched from `origin/main` @ `097b75a94`.

### DIFF
--- a/Planscape.Server/docs/rls_rehearsal.sql
+++ b/Planscape.Server/docs/rls_rehearsal.sql
@@ -1,0 +1,108 @@
+\set ON_ERROR_STOP on
+\echo '================ RLS REHEARSAL ================'
+
+-- Non-superuser role. Superusers ALWAYS bypass RLS, so testing as `postgres`
+-- would pass vacuously no matter what the policy said.
+DO $$
+BEGIN
+  IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'rls_app') THEN
+    EXECUTE 'DROP OWNED BY rls_app CASCADE';
+    EXECUTE 'DROP ROLE rls_app';
+  END IF;
+END $$;
+CREATE ROLE rls_app LOGIN PASSWORD 'rehearsal' NOSUPERUSER NOBYPASSRLS;
+
+DROP TABLE IF EXISTS "Projects";
+CREATE TABLE "Projects" ("Id" uuid PRIMARY KEY, "TenantId" uuid NOT NULL, "Name" text NOT NULL);
+ALTER TABLE "Projects" OWNER TO rls_app;
+
+INSERT INTO "Projects" VALUES
+  ('11111111-1111-1111-1111-111111111111','aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa','Tenant A - Alpha'),
+  ('22222222-2222-2222-2222-222222222222','aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa','Tenant A - Beta'),
+  ('33333333-3333-3333-3333-333333333333','bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb','Tenant B - Secret');
+
+\echo ''
+\echo '--- STEP 1: RED. No policy yet. Tenant A must be able to see Tenant B. ---'
+\echo '--- (If this is already 0, the rest of the test proves nothing.)      ---'
+SET ROLE rls_app;
+SET app.current_tenant = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+SELECT count(*) AS "RED_total_rows_visible_expect_3" FROM "Projects";
+SELECT count(*) AS "RED_tenantB_rows_visible_expect_1" FROM "Projects"
+  WHERE "TenantId" = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb';
+RESET ROLE;
+
+\echo ''
+\echo '--- STEP 2: apply the EXACT SQL RlsPolicyPatcher.PolicyFor("Projects") emits ---'
+ALTER TABLE "Projects" ENABLE ROW LEVEL SECURITY;
+ALTER TABLE "Projects" FORCE ROW LEVEL SECURITY;
+DROP POLICY IF EXISTS tenant_isolation ON "Projects";
+CREATE POLICY tenant_isolation ON "Projects"
+    USING ("TenantId"::text = current_setting('app.current_tenant', true))
+    WITH CHECK ("TenantId"::text = current_setting('app.current_tenant', true));
+
+\echo ''
+\echo '--- STEP 2b: idempotency — run it a second time, must not error ---'
+ALTER TABLE "Projects" ENABLE ROW LEVEL SECURITY;
+ALTER TABLE "Projects" FORCE ROW LEVEL SECURITY;
+DROP POLICY IF EXISTS tenant_isolation ON "Projects";
+CREATE POLICY tenant_isolation ON "Projects"
+    USING ("TenantId"::text = current_setting('app.current_tenant', true))
+    WITH CHECK ("TenantId"::text = current_setting('app.current_tenant', true));
+\echo 'idempotent re-apply OK'
+
+\echo ''
+\echo '--- STEP 3: GREEN positive path. Tenant A sees its OWN rows (must be NON-EMPTY = 2) ---'
+SET ROLE rls_app;
+SET app.current_tenant = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+SELECT count(*) AS "GREEN_tenantA_own_rows_expect_2" FROM "Projects";
+SELECT "Name" AS "GREEN_tenantA_names_expect_2_rows" FROM "Projects" ORDER BY "Name";
+
+\echo ''
+\echo '--- STEP 4: GREEN cross-tenant. Tenant A must see ZERO of Tenant B ---'
+SELECT count(*) AS "GREEN_tenantB_visible_to_A_expect_0" FROM "Projects"
+  WHERE "TenantId" = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb';
+
+\echo ''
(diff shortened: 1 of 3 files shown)


StingD85:
## Re-verified against today's `main` — taking this out of draft

This branch was cut at `097b75a94`; `main` is now `2414e7a8a`. Everything below was
re-measured on a **test-merge of this branch into current `main`**, not on the branch in
isolation.

### It still merges, and merging is a provable no-op

| Check | Result |
|---|---|
| Merge into current `main` | **clean** — auto-merge, one file (`Program.cs`), no conflicts |
| `Planscape.API` build on the merged tree | **0 errors** |
| Warning set vs `main` | **identical** — 3 × `CS0618` Hangfire `QueueName`, present on both. Pre-existing, established by comparison rather than asserted |
| Full server suite on the merged tree | **577 passed / 9 skipped / 0 failed (586)** — byte-for-byte the same as `main`'s baseline |

### The gate is still shut

- **`Database:RlsEnabled` is set nowhere.** Grep across `*.json`, `*.yaml`, `*.yml`, `*.env`
  returns only *comments* and the code that *reads* the key. No `appsettings*.json`, no
  `render.yaml`, no env file assigns it. `RlsPolicyPatcher.ApplyAsync` is therefore never
  reached on any current deployment.
- **The pooler is still out of the path.** `ConnectionStrings__Pooled` remains commented out
  at `render.yaml:99` (api) and `:156` (worker), so `pooledRaw` is empty and
  `Program.cs:114`'s `usePooler = pooledRaw is set && !rlsEnabled` is already `false`.
  Flipping the key changes no connection behaviour **today**; the cost only lands if someone
  later enables pooling on `planscape-db`.

### The two correctness properties still hold

- **Fails closed.** The policy predicate is bare
  `"TenantId"::text = current_setting('app.current_tenant', true)`. The migration's
  `OR coalesce(current_setting(...), '') = ''` escape hatch — the second independent reason
  RLS was inert — is gone. The only remaining `coalesce` in the file is at line 24, inside
  the doc comment *describing* what was removed; `grep -v '///'` finds none in any SQL
  string literal.
- **Idempotent by the only idiom Postgres offers.** 5 × `DROP POLICY IF EXISTS` + `CREATE`.
  There is no `CREATE POLICY IF NOT EXISTS` in Postgres — unlike `CREATE TABLE`, policies
  have no such form — so this is the correct construction, not a workaround.

### Unchanged and still owner-gated

Setting `Database:RlsEnabled` remains a separate decision, and the `BYPASSRLS` role for
`BypassTenantFilter` paths (Hangfire, cross-tenant admin scans) remains its prerequisite —
without it those paths will **silently find nothing** rather than leak anything. The rollback
ordering trap in the body still applies: *disabling the flag without dropping the policies is
the worst of the three states.*

No production connection was requested or used. The rehearsal in the body stands as run —
against a throwaway `rls_rehearsal` database as a `NOSUPERUSER NOBYPASSRLS` role, with RED
proven before GREEN.

PR #560, opened 2026-08-02
fix(bcc): validate the export selection before resolving an output folder (stacked on #550)

StingD85:
Your call was to hoist it; here it is, **stacked on #550 rather than folded into it**. Saying why explicitly, as asked.

## Why a separate PR, and why stacked rather than main-based

Two reasons, one of them decisive:

1. **You are mid-run.** #550 is deployed to `CompiledPlugin` and you are clicking through M1/M2/M3 right now. Changing its code invalidates the binary under test and would make you start over — the exact thing you wanted to avoid when you sequenced the harness ahead of the manual pass.
2. **A main-based PR is impossible.** `PlanscapeServerClient.SitePhotos.cs` only exists on #550's branch. It is not on `main`, and not on #559 (which branched from main and adds only the test project). So this had to be based on #550 or be part of it.

Merge order is therefore **#550 → this**. It retargets to `main` automatically once #550 lands.

## The change

`ExportPhotosAsync(projectId, photoIds, format)` resolved an output directory at `:551`, and only then did `ExportPhotosCoreAsync` check the ids-or-album guard (`:568`) and the per-format cap (`:575`).

Both guards are pure comparisons on the arguments. Neither needs a folder to exist. Running them second meant a user selecting 201 photos for a PDF got:

> Could not resolve an output folder: …

instead of:

> 201 photos selected; the server caps a PDF export at 200. Export in batches or use an album.

The first sends someone to investigate their Documents folder for a problem that is not there. As you put it — cheap validation before environment-dependent validation is right regardless of testability, and this is a good illustration of why: the ordering bug was invisible until something tried to reach the second check.

Both checks move into a shared `ExportSelectionRejected` so the wording exists **once**, and the public overload calls it before touching `OutputLocationHelper`. `ExportPhotosCoreAsync` still calls it too — the album overload does not go through the ids path, so that remains its last gate. **No message text changed.**

## Verified with the #559 harness against this build

| | Passed | Skipped | Failed |
|---|---|---|---|
| #550 as deployed | 11 | 3 | 0 |
| **this branch** | **12** | **2** | **0** |

The skip became a real test, exactly as you predicted:

```
Passed   Pdf_export_over_the_200_cap_is_refused_before_any_request_is_made
Skipped  Pdf_export_at_the_cap_is_allowed_through
```

The cap is now reached headlessly, and the capture server confirms **zero HTTP requests** were made — so "refused before the round trip" is now asserted rather than assumed.

`Pdf_export_at_the_cap_is_allowed_through` **still skips, and correctly**: a 200-photo selection *passes* the cap and then legitimately needs the output folder, which needs a Revit document context. That one is not a testability problem to solve — it is the environment honestly reporting what it cannot do.

## Scope

One file. No message text, no route, no signature changed. `dotnet build -c Release -t:Rebuild` → **0 errors, 0 warnings**.

Draft until #550 leaves draft, since it cannot merge before its base.

### DIFF
--- a/StingTools/BIMManager/PlanscapeServerClient.SitePhotos.cs
+++ b/StingTools/BIMManager/PlanscapeServerClient.SitePhotos.cs
@@ -544,7 +544,18 @@ namespace StingTools.BIMManager
             if (!await EnsureAuthenticatedAsync()) { LastError = "Not connected."; return false; }
 
             var ids = photoIds?.Distinct().ToArray();
-            var ext  = string.Equals(format, "pdf", StringComparison.OrdinalIgnoreCase) ? "pdf" : "zip";
+            var isPdfExport = string.Equals(format, "pdf", StringComparison.OrdinalIgnoreCase);
+
+            // Validate the SELECTION before touching the environment. Resolving an
+            // output directory can fail for reasons that have nothing to do with the
+            // request (no Revit document context, no writable folder), and when it
+            // did run first, a user who selected 201 photos for a PDF was told
+            // "Could not resolve an output folder" instead of that the cap is 200.
+            // Cheap, deterministic validation belongs ahead of environment-dependent
+            // validation regardless of what it does for testability.
+            if (ExportSelectionRejected(null, ids, isPdfExport)) return false;
+
+            var ext  = isPdfExport ? "pdf" : "zip";
             var name = $"photos-{DateTime.Now:yyyyMMdd-HHmmss}.{ext}";
 
             string dir;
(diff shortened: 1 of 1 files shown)


StingD85:
## Independently verified, and the draft blocker is gone

This PR said *"draft until #550 leaves draft, since it cannot merge before its base."* **#550 left draft**, so that condition is now satisfied. Marking ready.

Before flipping it I re-measured the claim rather than taking the body at its word, because CI never checked it — #560 targets `claude/bcc-sitephotos-real-http`, not `main`, and the build workflows filter on base, so **the only check that ran on this PR was Auto-label**. The body's numbers were the sole evidence.

### RED before GREEN, measured on one machine

The #559 harness is on `main` but post-dates this branch, so I merged `origin/main` into a throwaway branch in each case and ran the identical suite:

| | 550 only | 550 + this PR |
|---|---|---|
| `Pdf_export_over_the_200_cap_is_refused_before_any_request_is_made` | **Skipped** | **Passed** |
| `Pdf_export_at_the_cap_is_allowed_through` | Skipped | Skipped |
| totals | 12 passed / 2 skipped / **0 failed** | **13 passed / 1 skipped / 0 failed** |

So the claim holds exactly: the cap guard becomes reachable headlessly, and the capture server confirms the refusal happens with **zero HTTP requests**. The RED state here is *"Skipped"* rather than *"Failed"* — without the reordering the assertion cannot be reached at all, which is the more honest description of what this PR fixes.

`Pdf_export_at_the_cap_is_allowed_through` still skips, correctly — a 200-photo selection *passes* the cap and then legitimately needs a Revit document context. That is the environment reporting a real limit, not a testability gap to paper over.

Two differences from the body's numbers, both explained and neither a discrepancy in the change itself:
- I count 13/1 rather than 12/2 because my local docker stack was up, so the four `LiveStackRoundTripTests` executed instead of skipping.
- The plugin builds **0 errors / 0 warnings** (`dotnet build -c Release -t:Rebuild`), matching the body.

### Review note

`ExportSelectionRejected(null, ids, isPdfExport)` passes `albumId: null` from the public overload deliberately — that overload has no album parameter, so an empty selection must be rejected there. Behaviour is unchanged for every input that previously reached the guard; only the ordering relative to `OutputLocationHelper` moved. No message text, route, or signature changed.

Merge order remains **#550 → this**; it retargets to `main` automatically.

PR #569, opened 2026-08-03
fix(params): bind the sheet + project-info params that could never bind

StingD85:
Answers "how did we come to have unbound parameters" and fixes the subset that is genuinely broken.

## How the unbound set arose

Deliberately, in `07d16375e` (2026-07-08), which created `RESOLVED_BINDINGS.csv` as the binder's single source of truth:

> Result over 3,390 params: 1,992 scoped + 940 universal; **171 annotation-only + 138 Revit-native excluded; 72 genuinely-ambiguous gaps UNBOUND + listed.**

`873ac021a` then "bound 64 of 72 gaps — only true spares remain unbound". `SharedParamGuids` states the rule explicitly: *"A param ABSENT from the spec is intentionally UNBOUND (documented gap), never broad-bound — this stops cross-discipline leakage by construction."*

So most of the gap is a design decision, not rot. **Rebinding all of it would be wrong.**

## But there is a real defect underneath it

Each scoped row's category names resolve through `ParamRegistry.CategoryEnumMap`. Names that miss are dropped; a row whose names **all** miss yields an empty list, so the param never enters `_resolvedScoped` and is silently unbound.

`category_enum_map` held 143 entries — **every one a model category**. No `Project Information`, no `Sheets`, no `Views`, no `Title Blocks`. `universal_categories` is the same 143, so `<ALL>` params never reached them either.

| Category name used in the data | in `category_enum_map` | rows affected |
|---|---|---|
| `Materials` | no — but the loader skips it by name, handled by the material path | 108 (by design) |
| `Project Information` | **no** | 9 (silently unbound) |
| `Sheets` | **no** | scope not expressible → 11 params had no row at all |

### The sheet params are not optional

- `NativeParamMapper.MapSheets` writes `SHT_NUMBER` / `SHT_NAME` / `SHT_DISC` onto **every sheet**
- `ComplianceScan:485` reads `SHT_TAG_1`
- `TitleBlockRevisionSyncer:71` reads `SHT_REV`
- `docs/CHANGELOG.md` per-level-sheets acceptance row instructs: *"**Bind `STING_SHEET_CONTEXT_TXT` first (run LoadSharedParams once)** — without it … sheet identity does not survive a session restart"*

That last one is advice that could not be followed. The param has no spec row, and `Sheets` was not a scope the spec could express.

## Fix

- Add `Project Information` → `OST_ProjectInformation`, `Sheets` → `OST_Sheets`, `Views` → `OST_Views`, `Title Blocks` → `OST_TitleBlocks` to `category_enum_map`.
- Add 11 `RESOLVED_BINDINGS.csv` rows scoping the ten `SHT_*` params and `STING_SHEET_CONTEXT_TXT` to `Sheets`.

`universal_categories` is **deliberately left at 143 model categories** — widening it would drop all 1,119 universal params onto sheets and project information.

`Views` and `Title Blocks` are enabling only: no spec row uses them yet, and the `CATEGORY_BINDINGS.csv` path they unblock is inert while `HasResolvedSpec` is true. Added so the map stops silently swallowing those names — which is how this hid in the first place.

## Measured, before → after

| | before | after |
|---|---|---|
| scoped rows resolving to zero categories | 118 | 108 |
| …of which `Materials`-only (by design) | 108 | 108 |
| **real silent drops** | **10** | **0** |
| MR params with no spec row | 386 | 375 |

## The remaining 375 — checked, and correct as they stand

| Group | n | Delivery mechanism | Verdict |
|---|---|---|---|
| `TAG_*_BOOL` | 171 | family params inside tag families; `TagStyleEngine` reads them via `LookupParameter` on the tag type | leave |
| `WARN_*` | 65 | threshold config read from `PARAMETER_REGISTRY.json`, never off an element | leave |
| `Qto_*` | 59 | IFC quantity-set identifiers consumed by `IfcQuantitySetWriter` | leave |
| `TB_*` | 30 | title-block family params declared in `STING_TITLE_BLOCKS.json` (see #567) | leave |
| `VT_*` | 19 | view-template metadata, no code usage | leave |
| `QTO_*` | 13 | no code usage | leave |
| `PRJ_*` | 8 | the 7 GROUP-26 toggles, family-carried since #567, + `PRJ_SHEET_SYSTEM_TXT` | leave |
| `SPARE_*` / `BLE_*_SPARE_*` | 8 | declared spares — the "only true spares remain unbound" of `873ac021a` | leave |
| `VIEW_*` | 2 | no code usage | leave |

## Verification

- `dotnet build StingTools.csproj -c Debug` → **0 errors, 0 warnings**
- 2 files, 15 insertions, 0 deletions
- `PARAMETER_REGISTRY.json` re-parses; `universal_categories` verified still 143 with no non-model entry
- Re-ran the resolver over the full spec: 3,017 rows, 0 real silent drops

## Test plan

Run **Load Shared Params**. The eleven `SHT_*` / `STING_SHEET_CONTEXT_TXT` rows and the nine `Project Information` params should bind; `Skipped/failed` should fall by roughly 20. Then confirm a per-level production run's sheet identity survives a Revit restart — the behaviour the CHANGELOG row promised.


StingD85:
-

StingD85:
-

PR #590, opened 2026-08-05
fix(viewer): honest clipboard on meeting links; drive the meeting e2e spec; close the meeting/camera handoff

StingD85:
## Summary

Continues `docs/HANDOFF_MEETING_CAMERA.md`. The headline is that **both bugs it was written to chase are resolved** — confirmed by an owner-run reproduction on the live deploy, not by inference. Along the way the console surfaced one genuine, unrelated bug, which is fixed here.

**The two reported symptoms do not reproduce.** Project `2e4a5fb9…`, live `planscape-web-free` / `planscape-api-free`:
- model renders — 624 elements, `elementCount: 562`, `mesh→meta resolver: 562/562 (100%)`
- meeting reads **`1 online · 1 in call`** (the handoff recorded `0 in call`), LiveKit logs `publishing track`, camera tile renders

No single commit is identifiable as "the fix"; the symptoms were gone by the time the flow could be driven end to end. The changes in that area since the report — the `s2-latejoin` A/V bundle, the `ready`-counter fix in the host page, and carrying `modelId` across the meeting re-navigation — cover symptom 1's mechanism directly.

### The real bug found and fixed

Starting a meeting toasted "join link copied" over an untouched clipboard. Console: `The Clipboard API has been blocked because of a permissions policy applied to the current document` (`coordination-viewer.js:904`). Two independent faults:

1. The iframe's `allow` attribute listed only `fullscreen; camera; microphone; display-capture`. That attribute **replaces** the frame's default permissions policy, so `clipboard-write` was denied outright.
2. `copyToClipboard` could not detect the refusal — `navigator.clipboard.writeText` rejects **asynchronously**, so the synchronous `try/catch` never saw it: the function returned as though it had succeeded and the `execCommand` fallback never ran.

Fix 1 alone repairs the common path; 2 is what stops the UI lying wherever the permission is unavailable anyway (a direct hit on the API origin, a mobile WebView). Callers now report honestly — `startMeeting` points at the address bar, since the link is the URL it re-navigates to.

### The e2e spec now drives the flow

`meeting.spec.ts` never reproduced the meeting — it stopped at "Join control not found". It searched for `button:has-text("Join")`, but the menu entries are `<div class="menu-item" id="meetStart">`, and `#lkJoin` does not exist until `#meetStart` makes the viewer re-navigate into `?meeting=<id>`. Correct order: `#btnMeet → #meetStart → wait for re-navigation → #lkJoin`. It also filtered console output to `error`, discarding the `warn` lines where livekit-av.js reports its actual causes. Chromium now gets a fake camera, without which a headless join blocks on a permission prompt indistinguishable from the bug.

Kept despite the bugs being fixed: it is the regression guard for a flow that had none.

### Handoff closed

Rewritten from a to-do list into a record of seven disproven theories with the evidence that killed each. Includes retracting two of my own from this workstream — the tenant-mismatch 403 lead (those 403s are the `/health` poll, which returns 403 to an anonymous `curl`; it is IP-restricted by design and the polling code already treats any HTTP response as "reachable"), and the original "the 403 is why Join fails" framing (`livekit-token` has no 403 path at all). Console noise that is *not* a fault is recorded so it stops being re-investigated.

## Test plan

- [x] `node --check` on `coordination-viewer.js`
- [x] source ↔ `wwwroot` byte-equal (the CI gate) — synced by hand
- [x] `npm run typecheck` — clean
- [x] `npm test` — 80/80 pass
- [x] Live bundle verified byte-identical to source before relying on any code reading
- [ ] **Needs a browser:** click "Start a live meeting" in the embedded viewer and confirm the join link genuinely lands on the clipboard, and that a blocked clipboard now warns instead of claiming success

Note `npm run lint` is not configured in `planscape-web` (`next lint` prompts for setup), so it was not run.

Per the handoff: `Build & Test` / `CI Gate` are known-broken on `main` itself (Postgres service container, `role "root" does not exist`) and unrelated to this change.


StingD85:
Added `033598136` — the connectivity pill now polls `/health/live` instead of `/health`.

`/health` is the full diagnostic, locked to private-range IPs + `X-Health-Token`, so a browser always got 403. The pill coped (any HTTP response = reachable), but the browser logged a failed request every 15s — and that noise is what buried the Clipboard violation this PR fixes. 17 console errors in the owner's capture, almost all of them this poll.

`/health/live` already existed for this: `AllowAnonymous`, 200 `{status:"alive"}`, no topology. Verified anonymously against the live API. **`/health` itself is untouched**, so its security gate and Render's probe are unaffected.

Folded into this PR rather than opened separately, despite being a distinct concern: it edits the same file (`coordination-viewer.js`) and its byte-equal `wwwroot` twin, so a parallel PR off `main` would have conflicted on merge for no benefit. Happy to split it out if you'd rather review them apart.

All three checks still pass.

StingD85:
**Verified in a browser — 2026-08-06.** The one unchecked box in this PR's test plan is now closed.

Owner ran it against the live deploy: **Meet → Copy join link → Ctrl+V pastes a URL containing `meeting=`.**

That confirms *both* halves of the fix are deployed and working together:

- **web-app half** — the iframe now delegates `clipboard-write`, so the Clipboard API is no longer refused. This was the half that could not be verified without a signed-in session.
- **viewer half** — `copyToClipboard` awaits the promise, so it reports honestly rather than assuming success. Confirmed present in the served bundle (`grep -c "Clipboard blocked"` → 2).

Pre-flight on the same run: `grep -c "health/live"` → 1, confirming the quieter connectivity poll from this PR is live too.

No behaviour in this PR remains unverified.

PR #593, opened 2026-08-05
fix(billing): count seats over the rows the seat-selling paths actually write

StingD85:
Two commits: the RED proof, then the fix. **Billing-affecting** — please read the split rationale before merging.

## The defect, upstream of the known inversion

`POST /api/tenant/invite` gates on the seat axes — it calls `CheckCanAddUserAsync(role)` and returns **402 `quota_exceeded`** when full — so by its own contract an accepted invite consumes a seat. It then creates an `AppUser` **and nothing else. No `ProjectMember`.**

But every reader of the axes counted `ProjectMembers`:

| reader | counted |
|---|---|
| `QuotaGuardService.CountAsync` | `ProjectMembers`, split on `ProjectRole` |
| `TenantAdminController.Dashboard` | `ProjectMembers`, **the same split inlined a second time** |

So the endpoint gated on a number it could never move, while that number was driven entirely by `ProjectMembersController` — which **never consults the guard at all**. Gate and meter sat on opposite sides of a wall.

This is upstream of the author/coordinator inversion tracked in #538: correcting *which* roles land on *which* axis would still leave the meter reading a table the metered operation does not write.

## Proof first

`QuotaSeatMeterWiringTests` asserts intended behaviour against the customer-facing billing surface (`GET /api/tenant/dashboard`):

| | RED (before) | GREEN (after) |
|---|---|---|
| `Inviting_a_coordinator_consumes_a_coordinator_seat` | **FAIL** — Expected 1, Actual 0 | pass |
| `Inviting_an_author_consumes_an_author_seat` | **FAIL** — Expected 1, Actual 0 | pass |
| `An_invited_user_is_visible_to_the_tenant…` (control) | pass | pass |
| | **Failed 2, Passed 1** | **Failed 0, Passed 3** |

The control is what makes the other two mean anything: it proves the invite genuinely persisted a user, so the failures are the **meter** failing to count a real seat rather than the invite doing nothing. Without it the two failure modes are indistinguishable.

The dashboard reader also asserts the users array is non-empty — `ITenantContext.TenantId` falls back to `Guid.Empty`, which matches no rows, and an increment assertion against an unresolved tenant measures nothing.

## The split, and why each choice

**Seats count `AppUser`** — a seat is a person. This also makes the pre-existing `.Select(UserId).Distinct()` unnecessary rather than merely approximate: someone on four projects was always meant to be one seat.

**Author = `Iso19650Role == "A"`**, which is exactly what both invite paths persist for an Author (`Iso19650Role = role == "Author" ? "A" : "C"`), so meter and write agree by construction. This retires `ProjectRole == "Author"`, which could never match: `"Author"` is not in `ProjectRoles.All`, and the only `ProjectRole` values first-party code writes are **Owner, Manager, Contributor, Coordinator**.

**Coordinators = `total − authors`**, not `Iso19650Role != "A"`. In SQL a NULL role makes both `= A` and `<> A` evaluate to NULL, so a legacy row would fall off **both** axes and silently hand out a free seat. Subtraction is total by construction.

**`!u.IsDeleted` is explicit and must stay that way.** `AppUser` declares `HasQueryFilter(u => !u.IsDeleted)` in its own entity block, but `ApplyGlobalQueryFilters` runs later and overwrites it (EF Core 8 allows one filter per entity; a second call silently replaces the first), and `AppUser` does not implement `ISoftDeletable`. That is already documented in `PlanscapeDbContext` and is out of scope to fix here — but relying on the dead filter would bill tenants for **deleted** users.

**One source of truth.** `Dashboard` now reads both axes from the guard instead of recomputing them, so the number shown to the customer and the number enforced at the gate cannot drift apart. **No response-shape change** — same field names, same types.

## Interaction with #538 — one test needs rewriting

Ran #538's `QuotaSeatCountingTests` against this branch: **Failed 1, Passed 3.** Only `A_member_explicitly_marked_Author_moves_between_the_two_axes` fails, because it asserts `ProjectRole = "Author"` moves a seat — precisely the semantics this retires. The other two characterisation tests still pass, since that fixture's users carry no `"A"` role. #538's own header anticipates this and says those tests should be rewritten when the split is corrected. **Whichever merges second should update that one test.**

## Scope — deliberately not changed

- `tenant.MaxUsers` enforcement (`AdminController:78`, `ProjectMembersController:256`) is a **separate** cap system, counting `AppUser` against a per-tenant column. Left alone.
- The D1 identity-handoff path (`AuthController:1118`) bypasses quotas by documented design ("never lock out a customer D1 considers paid"). Left alone.
- No cap value, plan limit or tenant data touched.

## Verification

- Full suite on this branch: **Failed 0, Passed 580, Skipped 9, Total 589.**
- CI's Build & Test / CI Gate is broken on main itself (Postgres service container, "role root does not exist"), so verification is local.
- Fixtures are `zz-fixture-*` and deleted in a `finally`.
- Not merged, per the standing instruction.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/TenantAdminController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/TenantAdminController.cs
@@ -47,12 +47,14 @@ public class TenantAdminController : ControllerBase
         var limits = BillingPlanLimits.For(tenant.Plan);
 
         var memberCount = await _db.ProjectMembers.CountAsync(ct);
-        var authorIds = await _db.ProjectMembers
-            .Where(m => m.ProjectRole == "Author")
-            .Select(m => m.UserId).Distinct().ToListAsync(ct);
-        var coordIds  = await _db.ProjectMembers
-            .Where(m => m.ProjectRole != "Author")
-            .Select(m => m.UserId).Distinct().ToListAsync(ct);
+
+        // Seat counts come from the quota guard rather than being recomputed
+        // here. This block used to inline its own copy of the author/coordinator
+        // split, so the number the customer was shown and the number the invite
+        // gate enforced were two independent implementations that could disagree
+        // — and did, once the guard's definition of a seat changed. One source.
+        var authorSeats = await _quota.CheckCanAddUserAsync("Author", ct);
+        var coordSeats  = await _quota.CheckCanAddUserAsync("Coordinator", ct);
         var projectCount = await _db.Projects.CountAsync(ct);
         var storageBytes = await _db.ProjectModels.AsNoTracking()
             .Where(m => m.DeletedAt == null)
@@ -81,8 +83,8 @@ public class TenantAdminController : ControllerBase
             },
             usage = new
             {
-                authors      = new { current = authorIds.Count,  max = limits.MaxAuthors },
-                coordinators = new { current = coordIds.Count,   max = limits.MaxCoordinators },
+                authors      = new { current = authorSeats.Current, max = authorSeats.Max },
+                coordinators = new { current = coordSeats.Current,  max = coordSeats.Max },
                 projects     = new { current = projectCount,     max = limits.MaxProjects },
                 storage      = new { currentMb = storageBytes / 1024 / 1024, maxMb = limits.StorageMb },
                 memberSeats  = memberCount,
--- a/Planscape.Server/src/Planscape.Infrastructure/Services/QuotaGuardService.cs
+++ b/Planscape.Server/src/Planscape.Infrastructure/Services/QuotaGuardService.cs
@@ -73,12 +73,61 @@ public class QuotaGuardService : IQuotaGuardService
         var current = axis switch
         {
             QuotaAxis.Projects     => await _db.Projects.CountAsync(p => p.TenantId == tid, ct),
-            QuotaAxis.Authors      => await _db.ProjectMembers.Where(m => m.TenantId == tid && m.ProjectRole == "Author").Select(m => m.UserId).Distinct().CountAsync(ct),
-            QuotaAxis.Coordinators => await _db.ProjectMembers.Where(m => m.TenantId == tid && m.ProjectRole != "Author").Select(m => m.UserId).Distinct().CountAsync(ct),
+            QuotaAxis.Authors      => await CountAuthorSeatsAsync(tid, ct),
+            QuotaAxis.Coordinators => await CountUserSeatsAsync(tid, ct) - await CountAuthorSeatsAsync(tid, ct),
             _                      => 0,
         };
         return (limits, current);
     }
+
+    /// <summary>
+    /// A seat is a PERSON in the tenant, so seats are counted over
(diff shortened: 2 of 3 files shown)


StingD85:
Closing in favour of the capability-based approach on `claude/quota-capability-split` (new PR linked below).

Not repointing this branch: the `Iso19650Role` version was proposed and then **rejected on evidence**, and silently swapping a PR's contents would hide that. The history should show why.

**Why it was rejected:** `author = Iso19650Role == "A"` meters the *Appointing Party* — the client — not an author. `ProjectMembersController` GetRoles is explicit:

```csharp
new { Code = "A",  Label = "Appointing Party" },   // not Architect, not Author
new { Code = "BA", Label = "BIM Author" },         // the actual author code
```

That is why the Authors axis always read 0: almost nobody is tagged "A", while "BA" was counted as a coordinator.

Switching to `"BA"` was considered and also rejected. `Iso19650Role` is a functional/discipline taxonomy (Appointing Party, PM, BIM Coordinator, BIM Author, Architect, SE, ME, CE). ISO 19650 defines information-management responsibility, not software seats — so it can never be the standards-correct billing key. Metering it would bill an MEP Engineer who authors models as a coordinator, and change a customer's bill when their discipline is corrected.

The replacement meters **capability** instead, sourced from the same predicate the access gates read, so billing and access cannot drift apart — the failure that produced eleven dead gates in #540.

StingD85:
## ⚠️ DO NOT MERGE — superseded, and the approach here was rejected

This PR meters author seats on `Iso19650Role == "A"`. **That is wrong and should not ship.**

`Iso19650Role` is a functional/discipline taxonomy (`A/M/E/S/H/P/C/I/K/Q/F/W/L/Z`), and ISO 19650 assigns information-management responsibility — not software seats. **`"A"` is the Appointing Party — the client** — not "Author" and not "Architect". Metering on it counts the client as the only author (which is *why* the Authors axis read 0 for everyone else) and drops `"BA"`, BIM Author, onto the other axis entirely. It is the wrong question, not a missing code.

Merging this would auto-deploy a billing rule keyed on the wrong column.

### Where the work went instead

- **#606** — re-keys seats onto **authoring capability** (`ProjectRoles.CanAuthorInformation`), sourced from the same #540 capability layer access control uses, so billing and access cannot drift. Also raises the author cap to each plan’s paid headcount (price-neutral) and fixes an integer overflow found on the way: `MaxUsers` was `MaxAuthors + MaxCoordinators`, which wraps to `-2,147,483,648` once either side is `int.MaxValue`, denying a tenant’s *first* user.
- **#607** (stacked on #606) — makes viewers genuinely free: only author accounts consume a seat, promotion is charged, and seat reassignment is proven to have no lock-in.

Both carry the same RED/GREEN evidence this PR had, against the corrected key.

### What still stands from this PR

The diagnosis underneath it, which #606 keeps: the meter counted `ProjectMembers` while the seat-selling paths write `AppUser` and nothing else, so an accepted invite could never move the number it had just been checked against. That finding is preserved — only the *key* changed.

**Suggested action: close this in favour of #606 + #607.** I have not closed it myself.

PR #306, opened 2026-06-05
Viewer Phase-2 (appearance engine, cost, camera, minimap) + LiveKit A/V + orbit/EF/security hardening

StingD85:
## Summary

This branch lands a large, reviewable batch of Planscape web-viewer + meeting work, in
self-contained, individually-revertable commits. WebGL/viewer **feel** and native mobile A/V
are human/device-verified (they can't be CI-verified); server code is `dotnet build`-clean;
all viewer JS is `node --check`-clean; `assets/viewer ↔ wwwroot` is kept byte-equal
(incl. the vendored libs).

### Workstreams

**Orbit fix (rotation approach).** The model is rendered Y-up via a parent `modelPivot`
(Z-up→Y-up) with stock OrbitControls default up + `rotateSpeed = +0.9`; coordinate readouts
stay true Z-up via a rotation-aware `toWorld`/`toScene` bridge.

**Viewer UX (6 commits).** ACC-style framing + click/dblclick + pivot + issue/selection zoom;
markup lifecycle + exclusive tool state; 3D right-click context menu; section box (6-plane
clip + sliders + TransformControls gizmo + caps); orthographic toggle + ViewCube; collapsible
panels + 3D explode.

**Phase-2 viewer (M0–M7).**
- **M0** verified `mesh→meta` resolver + normalized keys (root-cause of dead toggles).
- **M1** ONE appearance engine (`applyAppearance`, single traverse, most-restrictive
  hide>ghost>colour>show) + discipline derived from category for as-built models.
- **M2** one shared ghost material, live tint/opacity.
- **M3** full Properties palette + per-element **cost** end-to-end (exporter→server→client).
- **M4** colouriser: palette sets, numeric **gradient**, `<No Value>`, interactive legend.
- **M5** `levelCamera` (already live) + human-eye view + Ctrl+Arrow elevation (camera getter).
- **M6** minimap fixed: single-renderer **scissor inset** (the 2nd WebGL context rendered
  empty) + drag-leak stopped.
- **M7** walk wheel relocated adjacent to the nav buttons.

**LiveKit A/V meetings.** Server token endpoint + `ActiveSurface` state + SignalR surface
broadcast; web `livekit-av.js` (cam/mic, tiles, screen-share, surface switching); mobile
`live.tsx`; `MeetingMedia` migration.

---

## Review-critical surfaces

**(a) Orbit reconciliation vs PR #305.** `main` (PR #305) shipped a different orbit hack
[...]

PR #602, opened 2026-08-05
ci: prove the SERVED viewer is current, not just the sources on disk (#309)

### LINKED ISSUE #309: CI: prove the DEPLOYED viewer is current (build container + assert served assets 200)
### Problem
The Docker image serves the **minified npm `dist/`** produced by the viewer build
(`Planscape/assets/viewer/build.mjs`), overlaid onto `wwwroot/`
(`Planscape.Server/docker/Dockerfile`: `npm run build` → `COPY --from=viewer /viewer/dist/ … wwwroot/`).
But CI only runs `node --check` on the **raw** sources and an assets ↔ wwwroot **byte-equal**
diff — neither builds the container or touches the actually-served artifact. So "CI green" does
**not** prove the deployed viewer is current: a stale/cached image (or a new asset that the build
never emits) can ship while every check passes.

This bit us on PR #306 — `livekit-av.js` (added to `wwwroot/` but not part of `build.mjs`'s
output) can 404 from a stale container. Manual mitigation landed as **Step 0** in
`docs/PR306_MANUAL_QA.md` (`docker compose build --no-cache` + `curl …/livekit-av.js` → 200), but
that's human-run, not enforced.

### Proposed CI step
Add a job that builds the container and asserts the deployed viewer is current:
1. `docker build` the API image (or `docker compose build api`) from the branch.
2. Run it; wait for `/health`.
3. `curl -fsS` a set of **known-served viewer assets** and assert **HTTP 200** for each —
   including a deliberately *new* asset that would 404 on a stale build
   (e.g. `livekit-av.js`, plus `viewer.html`, `coordination-viewer.js`, `signalr-shim.js`,
   `vendor/livekit-client.umd.min.js`).
4. Optionally diff a hash/마커 to confirm the served bytes match the freshly-built `dist/` + the
   committed raw-served files.

Also harden the build itself so it can't silently drop a raw-served file:
- Either make `build.mjs` emit/copy the raw-served viewer files (`livekit-av.js`, `meeting-sync.js`)
  into `dist/`, or assert in the Dockerfile that the final `wwwroot/` contains every file the
  drift gate covers.

### Acceptance
- A PR that adds a new viewer asset but forgets to wire it into the build **fails CI** (404),
  instead of green-then-404-in-prod.
- "CI green" implies the deployed viewer is current.

Ref: PR #306; `docs/PR306_MANUAL_QA.md` Step 0.

StingD85:
Closes #309.

# The existing viewer gate proves less than its name implies

This is the most important thing in this PR, and it changes how a rule that gets quoted in
handoffs should be read.

**`coordination-viewer.yml`'s drift check compares the RAW SOURCES on the runner's disk — not
the artifact a browser downloads.**

| | `coordination-viewer-drift.yml` compares | size |
|---|---|---|
| `Planscape/assets/viewer/coordination-viewer.js` | ↔ `wwwroot/coordination-viewer.js` | **431,575 bytes** — raw, unminified |
| what a browser actually GETs from the container | *(nothing checks this)* | **199,617 bytes** — minified `dist/` build |

Both sides of that diff are inputs. Neither is the output. So:

**What the drift gate CAN catch:** someone hand-editing `wwwroot/` instead of the canonical
source, so the two copies of the *source* diverge. That is real and worth keeping.

**What it CANNOT catch — and never could:**
- a **stale or cached image** serving an old bundle (the #306 failure: `livekit-av.js` 404'd
  while every check was green);
- an asset added to `wwwroot/` that `build.mjs` never emits;
- **any discrepancy in the built output at all** — including the one this PR found on its very
  first run (see below), which existed the whole time and was invisible to every previous tick;
- whether the container even starts, or serves anything.

"CI diffs 7 files byte-for-byte" is accurate and is quoted as a hard guarantee — but it is a
guarantee about **source parity**, not about what ships. Anyone reading a green tick as "the
deployed viewer is current" was reading more into it than it says. That gap is what this PR
closes, and the drift gate stays exactly as it is: it is fast, it catches a real mistake, and
it is now correctly scoped rather than over-trusted.

---

# It found a real defect on its first run

Not a hypothetical. Three dist assets differed from a fresh build by exactly 13 bytes each,
while `coordination-viewer.css` and `three.min.js` were byte-identical and every asset served
200.

**The image was never stale. The build was underspecified.**

The 13 bytes are a `"use strict";` prologue. esbuild walks **up** from each entry point looking
for a `tsconfig.json` and applies its `compilerOptions` to `.js` files too.
`Planscape/tsconfig.json` sets `"strict": true`, which implies TypeScript's `alwaysStrict`, so
esbuild emits the directive. The Dockerfile does `WORKDIR /viewer` and copies the sources in
flat — **no `tsconfig.json` above them in the container**, so no directive.

Same esbuild (0.24.2, pinned by the committed lockfile, which is *not* in `.dockerignore`), same
sources, same platform, identical `node_modules`. Different ancestry, different bytes.

Isolated by single-variable test rather than inferred:

| layout | `signalr-shim.js` |
|---|---|
| in-repo checkout | 3121 |
| out-of-repo flat copy | 3108 |
| out-of-repo replica of the exact nesting | 3108 |
| same replica **+ only `Planscape/tsconfig.json`** | **3121** ← isolated |

Eliminated first, by measurement: esbuild version, platform (Windows and the ubuntu runner
agree), source drift (no `import`/`export` anywhere), `node_modules` contents, and an ancestor
`package.json` — adding one changed nothing.

**Fix (`eef6e7e2d`): declare `format: 'esm'` in `build.mjs`.** It reproduces what production
already serves byte-for-byte (3108 / 23475 / 199617), so declaring it changes no deployed
artifact. `'cjs'` would have edited a working deployment by 13 bytes to satisfy a check, which
is backwards; `'iife'` **double-wraps** sources that are already hand-written IIFEs (3132,
matching neither environment) — tested before it was believed. `package.json` here already
declares `"type": "module"`, so `esm` is the declared type, not a convenient one.

With the format declared, all three layouts build identically. **The assertion in the workflow
is unchanged and still strict — this fixes the build, not the check.**

*Hazard recorded in `build.mjs`:* `viewer.html` loads all three with `<script defer src>`, not
`type="module"`. None has a top-level `import`/`export` today (verified: 0 each), so `esm`
output is a classic script. Adding one would silently break the page — and this gate compares
real served bytes, so it would catch it.

---

## The blind spot

Nothing in CI has ever touched the artifact that actually gets **served**.

| Existing check | What it looks at |
|---|---|
| `coordination-viewer-drift.yml` | `Planscape/assets/viewer/*` vs `wwwroot/*`, byte-equal — **on the runner's disk** |
| various | `node --check` on **raw sources** |

Neither builds the container, starts it, or requests a URL. So **"CI green" did not imply
the deployed viewer was current**: a stale or cached image could ship — or an asset could be
added to `wwwroot/` that `build.mjs` never emits — while every check passed.

Not hypothetical. It bit PR #306: `livekit-av.js` was added to `wwwroot/` but is not part of
`build.mjs`'s `dist/` output, and it 404'd from a stale container. The mitigation shipped as
a **human** step — `docs/PR306_MANUAL_QA.md` Step 0, *"`docker compose build --no-cache` then
`curl` the asset"* — which is exactly the kind of check that gets skipped on the busy day it
matters.

## What this adds

`.github/workflows/viewer-served-artifact.yml`:

1. **Builds `dist/` on the runner** with the same `build.mjs` the Dockerfile's viewer stage
   runs. These are the *reference bytes*.
2. **`docker build --no-cache`.** The `--no-cache` is the point of the exercise — a cached
   layer is precisely how a stale bundle ships, so reusing cache would reproduce the blind
   spot this exists to remove.
3. **Runs it** against a throwaway Postgres and waits for `/health/live`.
   `/health/live`, **not** `/health` — the latter is authenticated and answers 403 to an
   anonymous caller, which would read here as "never came up".
4. **Requests every viewer asset** and asserts HTTP 200 **and** byte-equality:

| Class | Asserted against | Why |
|---|---|---|
| dist-emitted | the `dist/` just built | a stale image cannot pass this |
| raw-served | the repo's `wwwroot/` | the #306 class of file — in `wwwroot`, never in `dist` |
| vendored | served + non-empty | third-party, copied wholesale |

**Anti-vacuity:** a 200 carrying a **zero-byte body** is failed explicitly, so an empty
response can never stand in for a working asset.

## The coverage assertion — so the two gates can't drift apart

`coordination-viewer-drift.yml` keeps 7 files in sync between source and `wwwroot`. Syncing a
file is worthless if it is never actually served. This workflow asserts every one of those 7
appears in its served-asset lists — so **adding a file to the drift list without serving it
fails here, not in production.** That is the acceptance criterion from the issue, enforced
rather than described.

## Asset paths verified against source, not guessed

`viewer.html` references `./coordination-viewer.{js,css}`, `./viewer-extras.js`,
`./signalr-shim.js`, `./meeting-sync.js`, `./livekit-av.js`, `./vendor/es-module-shims.js`,
and pulls `three` through an **importmap** at `./vendor/three/three.module.js`.
`livekit-av.js` loads `./vendor/livekit-client.umd.min.js` — the #306 chain. All four vendor
paths confirmed present in `wwwroot`.

`three.min.js` is kept as a **staleness canary** and labelled as one in the file: the page
does *not* load it (three comes from the importmap), but it **is** a `build.mjs` output the
Dockerfile copies, so its bytes drift when the image is stale. Documented so nobody later
deletes a line that looks unused.

## Kept separate from the fast gate, deliberately

Building a container and standing up Postgres costs minutes. `coordination-viewer-drift.yml`
stays a seconds-long byte diff and keeps giving fast feedback on the common mistake; this
runs alongside it on the same paths, plus `Planscape.Server/docker/**`, plus
`workflow_dispatch` for when only the base image has moved.

## What I did NOT verify — read this

**This workflow was not executed locally.** Running it means starting containers, and
containers are shared state in this checkout that I was asked not to touch without
permission.

Verified offline instead:

- the YAML parses;
- every `run` block passes `bash -n`;
- all four vendor assets exist in `wwwroot`;
- the drift-gate coverage loop was run standalone against both the real 7-file list (all
  covered) and a **negative control** (`not-served.js` → correctly flagged).

**The container round-trip itself is unproven until this runs on a PR.** Expect to iterate
once on the first run — most likely on how long the API takes to reach `/health/live` on a
cold image, or on an env var `Program.cs` fail-fasts over that I have not anticipated.

## Not done

The issue's alternative hardening — making `build.mjs` emit the raw-served files into
`dist/`, or asserting inside the Dockerfile that `wwwroot/` is complete — is left alone.
This gate catches the same failures from outside without changing what ships, which is the
smaller and more reversible move. If you would rather fix it at the build instead, this
workflow still tells you whether it worked.

### DIFF
--- a/.github/workflows/viewer-served-artifact.yml
+++ b/.github/workflows/viewer-served-artifact.yml
@@ -1,0 +1,256 @@
+name: Viewer served-artifact gate
+
+# Why this workflow exists (#309)
+# ===============================
+# Nothing in CI has ever touched the artifact that actually gets SERVED.
+#
+#   coordination-viewer-drift.yml  diffs Planscape/assets/viewer/* against wwwroot/*
+#                                  — raw sources, byte-equal, on the runner's disk
+#   (various)                      run `node --check` on raw sources
+#
+# Neither builds the container. Neither starts it. Neither requests a URL. So
+# "CI green" did NOT imply the deployed viewer was current: a stale or cached
+# image could ship — or an asset could be added to wwwroot/ that the viewer build
+# never emits — while every check passed.
+#
+# That is not hypothetical. It bit PR #306: `livekit-av.js` was added to wwwroot/
+# but is not part of build.mjs's dist/ output, and it 404'd from a stale
+# container. The mitigation shipped as a HUMAN step — `docs/PR306_MANUAL_QA.md`
+# Step 0, "docker compose build --no-cache then curl the asset" — which is
+# exactly the kind of check that gets skipped on the busy day it matters.
+#
+# This workflow enforces it: build the image, run it, and assert every viewer
+# asset is served with HTTP 200 AND that its bytes match what the build just
+# produced.
+#
+# Kept SEPARATE from the fast contract gate on purpose. Building a container and
+# standing up Postgres costs minutes; coordination-viewer-drift.yml stays a
+# seconds-long byte diff and keeps giving fast feedback on the common mistake.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'Planscape/assets/viewer/**'
+      - 'Planscape.Server/src/Planscape.API/wwwroot/**'
+      - 'Planscape.Server/docker/**'
+      - '.github/workflows/viewer-served-artifact.yml'
+  pull_request:
+    paths:
+      - 'Planscape/assets/viewer/**'
+      - 'Planscape.Server/src/Planscape.API/wwwroot/**'
+      - 'Planscape.Server/docker/**'
+      - '.github/workflows/viewer-served-artifact.yml'
+  # Buildable on demand — useful when only the base image has moved.
+  workflow_dispatch:
+
+jobs:
+  served-artifact:
+    name: Build the image, serve it, verify what comes back
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # ── 1. Build dist/ on the runner: the reference bytes ─────────────────
+      # Same build.mjs the Dockerfile's viewer stage runs. Whatever the container
+      # serves for a dist-emitted file must equal what this produces — that is
+      # what makes "the deployed viewer is current" a measurement rather than an
+      # assumption.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Build the viewer bundle (reference copy)
+        working-directory: Planscape/assets/viewer
+        run: |
+          set -euo pipefail
+          if [ -f package-lock.json ]; then npm ci --no-audit --no-fund
+          else npm install --no-audit --no-fund; fi
+          npm run build
+          ls -la dist/
+
+      # ── 2. Build the actual image ────────────────────────────────────────
+      # --no-cache on the viewer stage is the point of the exercise: a cached
+      # layer is precisely how a stale bundle ships. #306's asset 404'd from a
+      # cached image, so reusing cache here would reproduce the blind spot this
+      # workflow exists to remove.
+      - name: Build the API image (no cache)
+        run: |
+          set -euo pipefail
+          docker build \
+            --no-cache \
+            --build-arg IFCCONVERT_URL=skip \
+            -f Planscape.Server/docker/Dockerfile \
+            -t planscape-api:ci \
+            .
+
+      # ── 3. Run it ────────────────────────────────────────────────────────
+      - name: Start Postgres
+        run: |
+          set -euo pipefail
+          docker network create planscape-ci
+          docker run -d --name pg --network planscape-ci \
+            -e POSTGRES_DB=planscape \
+            -e POSTGRES_USER=planscape \
+            -e POSTGRES_PASSWORD=CiOnly2026! \
+            postgres:16-alpine
+          for i in $(seq 1 60); do
+            if docker exec pg pg_isready -U planscape -d planscape >/dev/null 2>&1; then
+              echo "postgres ready"; exit 0
+            fi
+            sleep 2
+          done
+          echo "::error::Postgres never became ready"; docker logs pg; exit 1
+
+      - name: Start the API container
+        run: |
+          set -euo pipefail
+          # CI-only JWT key. Never leaves this job, signs nothing that outlives
+          # it, and must still clear Program.cs's guards (32+ chars, 4+ distinct
+          # characters, not in the banned list).
+          docker run -d --name api --network planscape-ci -p 5000:8080 \
+            -e ASPNETCORE_ENVIRONMENT=Development \
+            -e ASPNETCORE_URLS=http://+:8080 \
+            -e 'ConnectionStrings__Default=Host=pg;Port=5432;Database=planscape;Username=planscape;Password=CiOnly2026!' \
+            -e "Jwt__Key=$(openssl rand -base64 48)" \
+            -e Jwt__Issuer=Planscape \
+            -e Jwt__Audience=Planscape.Client \
+            -e RateLimiting__Enabled=false \
+            planscape-api:ci
+
+      - name: Wait for /health/live
+        run: |
+          set -uo pipefail
+          # /health/live, NOT /health. /health is the full diagnostic and is
+          # authenticated — it answers 403 to an anonymous caller, which would
+          # read here as "never came up". /health/live is AllowAnonymous by
+          # design and is the liveness contract (see render.yaml).
+          for i in $(seq 1 90); do
+            if curl -fsS http://localhost:5000/health/live >/dev/null 2>&1; then
+              echo "api is live after ~$((i*2))s"; exit 0
+            fi
+            sleep 2
+          done
+          echo "::error title=API never became live::/health/live did not answer within 180s"
+          docker logs api | tail -100
+          exit 1
+
+      # ── 4. The actual assertion ──────────────────────────────────────────
+      - name: Every viewer asset is SERVED, and matches what was just built
+        run: |
+          set -uo pipefail
+          fail=0
+          src=Planscape/assets/viewer
+          dist=$src/dist
+          www=Planscape.Server/src/Planscape.API/wwwroot
+
+          # Emitted into dist/ by build.mjs, then overlaid onto wwwroot/ by the
+          # Dockerfile. Served bytes must equal the freshly-built dist bytes —
+          # this is the check that a stale image cannot survive.
+          #
+          # three.min.js is included as a STALENESS CANARY, not because
+          # viewer.html loads it — the page pulls three through an importmap
+          # pointing at vendor/three/three.module.js. three.min.js is still a
+          # build.mjs output copied by the Dockerfile, so if the image is stale
+          # its bytes drift, which is exactly the signal wanted. Said plainly so
+          # nobody later "fixes" this by deleting a line that looks unused.
+          DIST_ASSETS="coordination-viewer.js signalr-shim.js viewer-extras.js coordination-viewer.css three.min.js"
+
+          # NOT emitted by build.mjs. They ship from the repo's wwwroot/ and
+          # survive the dist overlay. They are exactly the class of file #306
+          # tripped over, so they are asserted explicitly rather than assumed.
+          RAW_ASSETS="viewer.html meeting-sync.js livekit-av.js"
+
+          # Vendored third-party, copied wholesale. Every one of these is
+          # referenced by viewer.html or by livekit-av.js — verified against the
+          # source, not guessed:
+          #   es-module-shims.js        <script> in viewer.html
+          #   three/three.module.js     importmap target for bare "three"
+          #   livekit-client.umd.min.js loaded by livekit-av.js (the #306 chain)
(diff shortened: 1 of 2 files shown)


StingD85:
## First CI run: the harness works. The assertion found something, and I have not resolved it.

As predicted in the body, the container round-trip was unproven. It now is — **every
infrastructure step passed**:

```
success  Build the viewer bundle (reference copy)
success  Build the API image (no cache)
success  Start Postgres
success  Start the API container
success  Wait for /health/live
failure  Every viewer asset is SERVED, and matches what was just built
```

So the image builds, the container serves, `/health/live` answers, and no timeout tuning is
needed. The red is the **assertion**, which is the part meant to have teeth.

### What it found

```
OK    viewer.html              (101969 bytes)
OK    meeting-sync.js          (53166 bytes)
OK    livekit-av.js            (65976 bytes)
OK    coordination-viewer.css  (65546 bytes, hash-identical to fresh build)
OK    three.min.js             (641540 bytes, hash-identical to fresh build)

ERROR coordination-viewer.js   served 199617  built 199630   (-13)
ERROR signalr-shim.js          served   3108  built   3121   (-13)
ERROR viewer-extras.js         served  23475  built  23488   (-13)
```

Every asset is **served** — no 404s, so the #306 failure class is clean. Three of the five
dist-emitted files differ from a fresh build by **exactly 13 bytes each**.

### What that pattern rules out, and what it does not

**It is almost certainly not a stale image.** A stale bundle would differ by an arbitrary amount,
would not land on the same delta across three files with different names and sizes, and would
very likely drag `coordination-viewer.css` and `three.min.js` with it — both of which are
byte-identical.

It also is not a toolchain drift: the Dockerfile viewer stage and the CI step both run
`npm ci` against the same committed `package-lock.json` (esbuild pinned `0.24.2`), and
`build.mjs` has no banner, no `define`, no timestamp and no git SHA — it is deterministic by
construction. `three.min.js` is built by the same esbuild call with the same `sourcemap: true`
and matches exactly.

The three that differ are the only ones built with `bundle: false`.

**What I have NOT established:** whether this is a genuine served-vs-built discrepancy worth
fixing, or my byte-equality assertion being stricter than esbuild's output is stable across
build environments for non-bundled entry points. Resolving it needs a local container build to
diff the two artifacts directly, and I was asked not to start containers in this checkout.

### Recommendation

Do not merge this on the assumption the red is cosmetic, and do not weaken the assertion to
make it green — either would reintroduce the blind spot the workflow exists to remove. The next
step is one local run:

```bash
docker build --no-cache --build-arg IFCCONVERT_URL=skip \
  -f Planscape.Server/docker/Dockerfile -t planscape-api:probe .
docker create --name probe planscape-api:probe
docker cp probe:/app/wwwroot/signalr-shim.js /tmp/served.js
docker rm probe
cd Planscape/assets/viewer && npm ci && npm run build
diff <(tail -c 200 /tmp/served.js) <(tail -c 200 dist/signalr-shim.js)
```

`signalr-shim.js` is the smallest of the three, so the 13 bytes will be readable directly. If
the difference turns out to be a sourcemap trailer or similar build-environment artefact, the
right fix is to compare *content* excluding that trailer — stated explicitly in the workflow —
rather than to drop the byte check.

StingD85:
## Falsification run: the framing is confirmed, the proposed fix is not. No change applied.

Measured, not argued. No container build was needed — building `dist/` locally and varying one
option was enough.

### Step 1 — which side does a plain local build reproduce?

| file | local build | CI "built" | CI "served" |
|---|---|---|---|
| `coordination-viewer.js` | **199630** | 199630 | 199617 |
| `signalr-shim.js` | **3121** | 3121 | 3108 |
| `viewer-extras.js` | **23488** | 23488 | 23475 |
| `three.min.js` | **641540** | 641540 | 641540 |

My box reproduces the **runner** side exactly. The container is the outlier.

### Step 2 — what the 13 bytes are. CONFIRMED.

Head of the locally-built `signalr-shim.js`:

```
"use strict";(function(){"use strict";if(typeof window>"u"||window.ReactNativeWebView||win…
```

The **outer** `"use strict";` is esbuild's — the inner one is the source's own, at
`signalr-shim.js:21`, inside its hand-written IIFE. `"use strict";` is exactly 13 bytes, and
`3121 − 13 = 3108`, the served size. **The 13 bytes are the prepended directive.**

### Step 3 — is `format: 'iife'` the cause? NO. FALSIFIED.

| block-2 `format` | `signalr-shim.js` | vs served (3108) |
|---|---|---|
| *(none — current)* | 3121 | +13 |
| **`'iife'`** | **3132** | **+24 — further away** |
| `'cjs'` | 3121 | +13 |
| **`'esm'`** | **3108** | **exact match** |

Declaring `iife` makes the files **larger**, matching neither side. The reason is in the source:
these three files are **already hand-written IIFEs** (`(function(){…})()`), so `format: 'iife'`
wraps them a second time. The block comment calling them "the three IIFE viewer files" describes
the *sources*, not a requested output format.

**So the one-line fix as proposed would not have fixed this, and would have changed shipped bytes
in a third direction.** Good thing it was tested first.

### What IS established

The **framing was right**: this is format inference varying across environments, and the block-2
loop is underspecified. The resolution is just different from the hypothesis —

- the **container** infers **ESM** (implicitly strict → no directive) → 3108
- the **CI runner and my machine** infer **CJS** (directive emitted) → 3121

Ruled out along the way, by measurement rather than assumption:

- **esbuild version** — `package-lock.json` is committed, is **not** in `.dockerignore`, and both
  the Dockerfile stage and the CI step run `npm ci`. Same pinned `0.24.2`.
- **platform** — Windows (mine) and ubuntu (runner) both give 3121; only the container differs,
  and it is linux too.
- **source drift** — the source has no `import`/`export`, so nothing in the file forces ESM.

### Two things this settles

**The image was never stale.** Every asset served 200, `three.min.js` and the CSS are
byte-identical, and the three that differ are explained entirely by a directive esbuild chose
not to emit. This was an underspecified build, not a stale artifact — which is the better
outcome, and the opposite of what a bare byte-mismatch would have suggested.

**What the old gate was actually proving.** `coordination-viewer-drift.yml` byte-compares
`Planscape/assets/viewer/*` against `wwwroot/*` — the **raw, unminified** sources on the runner's
disk. Those are the 431575-byte originals, not the ~199KB artifact the browser downloads. So
every previous "assets are current" tick was checking inputs that both builds agree on, and had
no visibility into the output at all. It could not have caught this, and could not have caught a
stale image either. That is the gap this PR closes.

### Why I have applied nothing

The instruction was to apply `format: 'iife'` **only if confirmed**. It is falsified, so nothing
is applied — `build.mjs` is untouched and the assertion is unchanged and still strict.

Declaring a format is still the right fix, but **which** one is now a decision that changes
shipped bytes, not a mechanical edit:

- `'cjs'` keeps today's runner/local output (3121) and adds a directive the browser does not need;
- `'esm'` matches what the container currently ships (3108) but is a misleading label for a file
  loaded via plain `<script src>`, not `type="module"`;
- `'iife'` is semantically the honest description but double-wraps an already-IIFE source (3132).

My inclination is `'cjs'` — smallest behavioural change from what CI and developers already
produce, and the directive is harmless — but that ships 13 bytes that production does not
currently serve, so it is your call, not mine.

PR #606, opened 2026-08-05
fix(billing): meter seats on authoring capability, not on ISO 19650 role

StingD85:
Supersedes #593 (closed — see the rationale there).

## Why the previous approach was wrong

`author = Iso19650Role == "A"` meters the **Appointing Party** — the client — not an author. `ProjectMembersController.GetRoles` is explicit:

```csharp
new { Code = "A",  Label = "Appointing Party" },   // not Architect, not Author
new { Code = "BA", Label = "BIM Author" },         // the actual author code
```

That is why the Authors axis always read 0: almost nobody carries `"A"`, while `"BA"` was counted as a coordinator.

Switching to `"BA"` was considered and rejected too. `Iso19650Role` is a functional/discipline taxonomy; **ISO 19650 defines information-management responsibility, not software seats**, so it can never be the standards-correct billing key. It would bill an MEP Engineer who authors models as a coordinator, and move a customer's bill when their discipline is corrected.

## What this does instead

**Capability split.** `ProjectRoles.CanAuthorInformation` plus an EF-translatable predicate, sitting beside `CanCurate` / `CanApproveSitePhotos`, so **billing and access read one source** and cannot drift apart — the failure mode that produced eleven dead gates in #540. A test asserts the in-memory and SQL forms agree for every declared `UserRole`.

`SecurityOfficer` is excluded from the author axis: it cannot edit projects, members, or BIM-Manager roles by its own definition, so billing it as an author would charge for an audit control.

**The invite sells the seat it meters.** Both paths now offer Author or Viewer and write `UserRole.Contributor` / `UserRole.Viewer`. The API-visible `role="Coordinator"` now creates a read-only account. A renamed test asserts the *author* axis does **not** move — the assertion that would have caught the original mispairing.

**Limits are price-neutral.** `MaxAuthors` = each plan's former total (Trial 1, PluginOnly 1, Studio 6, Practice 12, Network 20); read-only unlimited. `BillingPlanLimitsTests` pins the paid headcount per plan, so a later edit to `MaxAuthors` fails and forces the pricing question to be answered deliberately rather than silently.

## Integer overflow found while applying it — latent today

`tenant.MaxUsers` was `MaxAuthors + MaxCoordinators`. With either side at `int.MaxValue` it wraps negative — measured `1 + int.MaxValue = -2,147,483,648` — and it is consumed as `userCount >= tenant.MaxUsers`, so **a negative cap denies the tenant's first user**. Enterprise already wraps to `-2` today; it is only latent because registration assigns Trial.

Making viewers unlimited would have detonated this on every new tenant. Both call sites now use a saturating `TotalSeats`; grep confirms no raw sum remains.

## Known shortfall, stated deliberately

**"Viewers unlimited" is not yet true in practice.** Returning infinity for `TotalSeats` would let `AdminController` and `ProjectMembersController` add unlimited **authoring** accounts with nothing counting them, since `MaxUsers` is the only cap those two enforce and neither consults the quota guard. So `TotalSeats` reports the paid headcount instead.

The cost is real: a Studio tenant is still capped at **6 accounts total**, so a seventh free viewer is refused by `MaxUsers` even though the read-only axis is unlimited. Erring toward charging correctly beat erring toward giving seats away — but it is not the full intent.

Making viewers genuinely free needs those two paths to check authoring capability instead of `MaxUsers` — the same "meter the writes" move. That changes their enforcement behaviour, so it is deliberately left for its own PR.

## Test plan

- [x] Full suite: **Failed 0, Passed 598, Skipped 9, Total 607**
- [x] In-memory and SQL forms of the capability predicate agree for every declared `UserRole`
- [x] Invite writes the role it sells; author axis asserted not to move for `role="Coordinator"`
- [x] `BillingPlanLimitsTests` pins paid headcount per plan
- [x] Saturating `TotalSeats` at both call sites; no raw sum remains

## Before merging

This is a **pricing position**, not just a fix: `MaxAuthors` of 6 on Studio, with viewers free once the follow-up lands. Confirm the two-tier split is what is being sold — the engineering is sound either way, but that line is a business decision.

Note: `Build & Test` / `CI Gate` fail on a Postgres service container (`role "root" does not exist`) — broken on `main` itself and unrelated.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/AuthController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/AuthController.cs
@@ -460,7 +460,7 @@ public class AuthController : ControllerBase
             Plan          = BillingPlan.Trial,
             Currency      = currency,
             BillingCycle  = BillingCycle.Monthly,
-            MaxUsers      = planLimits.MaxAuthors + planLimits.MaxCoordinators,
+            MaxUsers      = planLimits.TotalSeats,
             MaxProjects   = planLimits.MaxProjects,
             MimEnabled    = false,
             TrialExpiresAt = DateTime.UtcNow.AddDays(30)
@@ -1095,7 +1095,7 @@ public class AuthController : ControllerBase
                     Plan           = BillingPlan.Trial,
                     Currency       = "USD",
                     BillingCycle   = BillingCycle.Monthly,
-                    MaxUsers       = limits.MaxAuthors + limits.MaxCoordinators,
+                    MaxUsers       = limits.TotalSeats,
                     MaxProjects    = limits.MaxProjects,
                     MimEnabled     = false,
                     TrialExpiresAt = DateTime.UtcNow.AddDays(365)
--- a/Planscape.Server/src/Planscape.API/Controllers/OnboardingController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/OnboardingController.cs
@@ -89,7 +89,12 @@ public class OnboardingController : ControllerBase
         foreach (var m in req.Members)
         {
             // Quota check per member — fail mid-list cleanly with a 402.
-            var role = string.Equals(m.Role, "Author", StringComparison.OrdinalIgnoreCase) ? "Author" : "Coordinator";
+            // Two options = the two seat axes: can-author, or read-only. See the
+            // note in TenantAdminController.Invite — Coordinator is an authoring
+            // role, so the old Author-vs-Coordinator pair gated on one axis and
+            // consumed a seat on the other. API-visible: a member sent with
+            // role = "Coordinator" is now created read-only.
+            var role = string.Equals(m.Role, "Author", StringComparison.OrdinalIgnoreCase) ? "Author" : "Viewer";
             var q = await _quota.CheckCanAddUserAsync(role, ct);
             if (!q.Allowed) return StatusCode(StatusCodes.Status402PaymentRequired, new { error = "quota_exceeded", added, denied = m, quota = q });
 
@@ -102,7 +107,9 @@ public class OnboardingController : ControllerBase
                 Email = m.Email.Trim().ToLowerInvariant(),
                 DisplayName = m.DisplayName,
                 PasswordHash = "INVITED",
-                Role = role == "Author" ? UserRole.Contributor : UserRole.Coordinator,
+                // UserRole is the seat: ProjectRoles.CanAuthorInformation reads it.
+                Role = role == "Author" ? UserRole.Contributor : UserRole.Viewer,
+                // ISO code is for deliverables/reporting only, never billing.
                 Iso19650Role = role == "Author" ? "A" : "C",
                 IsActive = false,
             };
--- a/Planscape.Server/src/Planscape.API/Controllers/TenantAdminController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/TenantAdminController.cs
@@ -106,7 +106,15 @@ public class TenantAdminController : ControllerBase
     public async Task<ActionResult> Invite([FromBody] InviteUserRequest req, CancellationToken ct)
     {
         // Quota check — refuses with 402 if cap reached.
-        var role = string.Equals(req.Role, "Author", StringComparison.OrdinalIgnoreCase) ? "Author" : "Coordinator";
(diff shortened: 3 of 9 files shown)


StingD85:
## Safe to close once you are ready — the one non-salvageable fix is now captured in #616

The seat split here is superseded by the licence-based model (entitlement = the StingTools licence; one licence = one author seat). Most of what these PRs carry goes with them, deliberately.

**The exception is the saturating `TotalSeats` fix**, and it is *not* independently salvageable: `TotalSeats` is itself part of the author/coordinator split (`Tenant.cs`, `QuotaGuardService`, `AuthController` ×2, plus 3 test files), so cherry-picking it would re-introduce the very model the licence decision replaces.

**#616** therefore carries the requirement and its test forward, to be re-fixed against the licence model: any total-seat figure written to `Tenant.MaxUsers` must **saturate** rather than sum, because `1 + int.MaxValue = -2,147,483,648` and `MaxUsers` is consumed as `userCount >= MaxUsers` — a negative cap denies the tenant its first user.

One correction recorded in #616 rather than repeated: this is currently **latent, not live**. No tenant has a non-positive cap; `planscape`’s `int.MaxValue` comes from a *direct* assignment in `PlatformTenantSeeder.cs:57`, not from the sum, and a large positive cap permits rather than denies. Both creation paths assign `Trial` (1 + 0). It goes live the moment any plan with an unlimited axis is assigned at creation — which is exactly what these PRs did by setting `MaxCoordinators = int.MaxValue` on every plan, and how it was found.

Also still standing from this stack, unaffected by the seat decision:
- **#615** salvages the last-admin lockout guard onto `main` independently (Failed 2 → Failed 0). **#608 can close once #615 merges.**
- The meter-vs-writes diagnosis — the seat meter counted `ProjectMembers` while the seat-selling paths write `AppUser` and nothing else — remains true and is worth remembering when wiring licences to the server’s view.

Not closing these myself.

StingD85:
Closing — superseded by the licence-based seat model (entitlement = the StingTools licence; one licence = one author seat; billing reads no role).

The one fix worth keeping has landed on its own: **#617** implements the saturating seat total against `main`, green, closing #616. My earlier assessment that it was not independently salvageable was too pessimistic — the arithmetic is separable from the author/coordinator split, and #617 touches only the `Limits` record and the two `MaxUsers` writers, with no `QuotaGuardService` and no change to any plan’s numbers.

Also already salvaged: the last-admin lockout guard, via **#615** (merged), which is why #608 is closed.

What dies with these PRs, deliberately: the author/coordinator axis, the `Iso19650Role`-keyed and capability-keyed meters, the free-viewer enforcement, and the plan-limit changes. All of it is replaced by the licence model.

What survives as knowledge rather than code, and is worth remembering when the plugin licence is wired to the server: **the seat meter counted `ProjectMembers` while the seat-selling paths write `AppUser` and nothing else**, so an accepted invite could never move the number it had just been checked against. Whatever reflects licences server-side must not repeat that shape.

StingD85:
Superseded by #617 (`928da5db4`, *fix(billing): saturate the plan seat total so an unlimited plan cannot deny its first user*), which landed on `main` while this was open and rewrote the same limits table in `Tenant.cs`.

The two are not reconcilable by merge — they are different commercial models:

| Plan | main today (#617) | this PR |
|---|---|---|
| Trial | 1 author + 0 coordinators | 1 + unlimited |
| PluginOnly | 1 + 0 | 1 + unlimited |
| Studio | **1** + 5 | **6** + unlimited |
| Practice | **1** + 11 | **12** + unlimited |
| Network | **1** + 19 | **20** + unlimited |

main sells one author plus N coordinators; this PR sells N authors plus unlimited viewers. Merging would have silently replaced a live pricing structure with a more generous one, so it is being closed rather than resolved in a conflict editor.

**Worth salvaging separately.** The diagnosis behind this PR stands on its own and is independent of the limits table: `QuotaAxis.Authors` counted `ProjectRole == "Author"`, a string nothing assigns, so the axis read 0 for every tenant. `"A"` in ISO 19650 is *Appointing Party* — the client — not *Author*, and `"BA"` (BIM Author) fell to the other axis entirely. If that counting bug is still present on `main` after #617, it deserves its own PR that changes **how seats are counted** without touching **what the caps are**.

Closed as superseded, not as wrong.

PR #607, opened 2026-08-06
fix(billing): only author accounts consume a seat — viewers genuinely free

StingD85:
**Stacked on #606** — base is `claude/quota-capability-split`, so this diff shows only the follow-up. Merge #606 first.

Author seats are the only paid unit. Read-only accounts are unlimited and free, including the Appointing Party. **Plan prices and author counts unchanged** (Trial 1, PluginOnly 1, Studio 6, Practice 12, Network 20). Projects-axis quota untouched.

## What a tenant AT its author cap can do

| | before | after |
|---|---|---|
| Add a viewer | ❌ refused | ✅ unlimited, free |
| Add an author | ❌ refused | ❌ refused |
| Promote a viewer to author | ✅ **allowed — unmetered** | ❌ refused |
| Demote an author → free the seat | ✅ | ✅ immediately reusable |

`AdminController` and `ProjectMembersController` compared **total active accounts** against `tenant.MaxUsers`, so a Studio tenant was capped at 6 accounts of *any* kind — the seventh person was refused even when read-only. Both now ask `IQuotaGuardService` for the author axis, so the cap that refuses a user is the same number the tenant dashboard shows them.

## Grepped, not assumed

Every `MaxUsers` reader in `src/`:

- **Enforcement — exactly two, both converted:** `AdminController:78`, `ProjectMembersController:256`.
- **Writes:** `AuthController` (registration + D1 handoff), `SeedData`, `DemoSandboxJob`, `PlatformTenantSeeder`.
- **Display:** `AdminController:44` — and **no client reads it** (verified across `Planscape/src`, `planscape-web`, `Planscape.Desktop/src`).
- **Dead:** `TierLimits.MaxUsers(tier)` has **zero callers**. Never an enforcement path.

## Promotion is a purchase

`UpdateUser` now charges a non-author becoming an author. Without it, "viewers are free" is just a bypass — mint unlimited free viewers, then promote them all. Demotion is always allowed and frees the seat instantly.

## Seat reassignment: already possible, no lock-in

`PUT /api/admin/users/{id}` with a new role is the mechanism, so **no new endpoint was needed**. I grepped for cooldown / lock-in / billing-period / minimum-term gating on role changes — **none exists**. `Freeing_a_seat_makes_it_immediately_reusable` proves it: fill Studio to 6 authors, confirm the 7th is refused, demote one, and the next request succeeds. No cooling-off, no period boundary.

## TotalSeats re-checked, as asked

With viewers genuinely uncapped there is no total-account cap, so `TotalSeats` is now honestly `int.MaxValue` on every plan — safe **only because it saturates**. Written as `MaxAuthors + MaxCoordinators` it is `6 + int.MaxValue = -2147483643` for Studio, and `MaxUsers` is consumed as `userCount >= MaxUsers`, which would deny the tenant’s **first** user. That overflow was found precisely because this change was attempted, and it still holds.

#606 returned `MaxAuthors` from `TotalSeats` purely because these two paths had not yet been converted; that reason is now gone. The **paid** number is `MaxAuthors`, and the tests read it from there.

## RED before GREEN, both paths

| | result |
|---|---|
| `SeatEnforcementTests` before | **Failed 4, Passed 1** |
| `SeatEnforcementTests` after | **Failed 0, Passed 5** |
| Full suite | **Failed 0, Passed 603**, Skipped 9, Total 612 |

The single pass in the RED run is the control — at the cap, today, *everything* is refused, so "author refused" was already true. Each test pins the tenant’s `MaxUsers` to the current headcount so the OLD rule refuses outright; a success can then only come from the new author-seat rule, never from leftover headroom.

Injecting `IQuotaGuardService` into two controllers is a DI change the compiler cannot check — both are exercised over real HTTP here, so a mis-registration fails the suite rather than production.

Each test class gets its own database (`PlanscapeTest_{Guid}`), so mutating plan state cannot leak. Fixtures are `zz-fixture-*`.

## Not fixed — reported

`AdminController.UpdateUser` has **no last-Owner guard**: an admin can demote the final Owner to Viewer and lock the tenant out of administration. Pre-existing, but more reachable now that demotion is the seat-management tool. Out of scope — it changes behaviour on a path this PR only extends.

CI’s Build & Test / CI Gate is broken on main itself (Postgres service container, "role root does not exist"), so verification is local. Not merged, per the standing instruction.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/AdminController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/AdminController.cs
@@ -17,13 +17,39 @@ public class AdminController : ControllerBase
 {
     private readonly PlanscapeDbContext _db;
     private readonly Planscape.Infrastructure.Authorization.IPermissionRevocationStore _revocations;
+    private readonly Planscape.Infrastructure.Services.IQuotaGuardService _quota;
 
     public AdminController(
         PlanscapeDbContext db,
-        Planscape.Infrastructure.Authorization.IPermissionRevocationStore revocations)
+        Planscape.Infrastructure.Authorization.IPermissionRevocationStore revocations,
+        Planscape.Infrastructure.Services.IQuotaGuardService quota)
     {
         _db = db;
         _revocations = revocations;
+        _quota = quota;
+    }
+
+    /// <summary>
+    /// Seats are AUTHOR seats. Read-only accounts are unlimited and free, so
+    /// only an account that can author information is checked against the cap.
+    ///
+    /// <para>Asks <see cref="Planscape.Infrastructure.Services.IQuotaGuardService"/>
+    /// rather than counting here: it is the same implementation the invite paths
+    /// and the tenant dashboard use, so the cap a user is refused by is the cap
+    /// they were shown. This controller previously compared total active
+    /// accounts against <c>tenant.MaxUsers</c>, which capped a Studio tenant at
+    /// 6 accounts of ANY kind — a seventh read-only account was refused.</para>
+    /// </summary>
+    private async Task<string?> AuthorSeatRefusalAsync(UserRole role, CancellationToken ct = default)
+    {
+        if (!ProjectRoles.CanAuthorInformation(role)) return null;   // viewers are free
+
+        var seats = await _quota.CheckCanAddUserAsync("Author", ct);
+        return seats.Allowed
+            ? null
+            : $"Author seat limit reached ({seats.Current} of {seats.Max}). "
+            + "Read-only accounts are unlimited — create this user as a Viewer, "
+            + "or free a seat by changing an existing author to Viewer.";
     }
 
     // ── Organization Management ──
@@ -74,9 +100,10 @@ public class AdminController : ControllerBase
         var tenant = await _db.Tenants.FindAsync(tenantId);
         if (tenant == null) return NotFound("Tenant not found");
 
-        var userCount = await _db.Users.CountAsync(u => u.TenantId == tenantId && u.IsActive);
-        if (userCount >= tenant.MaxUsers)
-            return BadRequest($"User limit ({tenant.MaxUsers}) reached for {tenant.Tier} tier");
+        var newRole = Enum.TryParse<UserRole>(req.Role, true, out var r) ? r : UserRole.Contributor;
+
+        // Only authoring accounts consume a seat; viewers are unlimited.
+        if (await AuthorSeatRefusalAsync(newRole) is { } refusal) return BadRequest(refusal);
 
         if (await _db.Users.AnyAsync(u => u.Email == req.Email))
             return Conflict($"Email {req.Email} already exists");
@@ -87,7 +114,7 @@ public class AdminController : ControllerBase
             Email = req.Email,
             DisplayName = req.DisplayName,
             PasswordHash = HashPassword(req.Password),
-            Role = Enum.TryParse<UserRole>(req.Role, true, out var r) ? r : UserRole.Contributor,
+            Role = newRole,
             Iso19650Role = req.Iso19650Role ?? "M"
         };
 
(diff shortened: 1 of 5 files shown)


StingD85:
## Safe to close once you are ready — the one non-salvageable fix is now captured in #616

The seat split here is superseded by the licence-based model (entitlement = the StingTools licence; one licence = one author seat). Most of what these PRs carry goes with them, deliberately.

**The exception is the saturating `TotalSeats` fix**, and it is *not* independently salvageable: `TotalSeats` is itself part of the author/coordinator split (`Tenant.cs`, `QuotaGuardService`, `AuthController` ×2, plus 3 test files), so cherry-picking it would re-introduce the very model the licence decision replaces.

**#616** therefore carries the requirement and its test forward, to be re-fixed against the licence model: any total-seat figure written to `Tenant.MaxUsers` must **saturate** rather than sum, because `1 + int.MaxValue = -2,147,483,648` and `MaxUsers` is consumed as `userCount >= MaxUsers` — a negative cap denies the tenant its first user.

One correction recorded in #616 rather than repeated: this is currently **latent, not live**. No tenant has a non-positive cap; `planscape`’s `int.MaxValue` comes from a *direct* assignment in `PlatformTenantSeeder.cs:57`, not from the sum, and a large positive cap permits rather than denies. Both creation paths assign `Trial` (1 + 0). It goes live the moment any plan with an unlimited axis is assigned at creation — which is exactly what these PRs did by setting `MaxCoordinators = int.MaxValue` on every plan, and how it was found.

Also still standing from this stack, unaffected by the seat decision:
- **#615** salvages the last-admin lockout guard onto `main` independently (Failed 2 → Failed 0). **#608 can close once #615 merges.**
- The meter-vs-writes diagnosis — the seat meter counted `ProjectMembers` while the seat-selling paths write `AppUser` and nothing else — remains true and is worth remembering when wiring licences to the server’s view.

Not closing these myself.

StingD85:
Closing — superseded by the licence-based seat model (entitlement = the StingTools licence; one licence = one author seat; billing reads no role).

The one fix worth keeping has landed on its own: **#617** implements the saturating seat total against `main`, green, closing #616. My earlier assessment that it was not independently salvageable was too pessimistic — the arithmetic is separable from the author/coordinator split, and #617 touches only the `Limits` record and the two `MaxUsers` writers, with no `QuotaGuardService` and no change to any plan’s numbers.

Also already salvaged: the last-admin lockout guard, via **#615** (merged), which is why #608 is closed.

What dies with these PRs, deliberately: the author/coordinator axis, the `Iso19650Role`-keyed and capability-keyed meters, the free-viewer enforcement, and the plan-limit changes. All of it is replaced by the licence model.

What survives as knowledge rather than code, and is worth remembering when the plugin licence is wired to the server: **the seat meter counted `ProjectMembers` while the seat-selling paths write `AppUser` and nothing else**, so an accepted invite could never move the number it had just been checked against. Whatever reflects licences server-side must not repeat that shape.

StingD85:
Superseded by #617 (`928da5db4`), along with its base #606 — see the detail there.

This PR sat on `claude/quota-capability-split` (#606) and depended on that PR's limits table, where a seat is an author and viewers are unlimited. #617 landed a different model on `main`: one author plus N coordinators, with `MaxCoordinators` finite per plan. Under main's table a Studio tenant is 1 author + 5 coordinators; under this stack it was 6 authors + unlimited viewers. Merging would have replaced a live pricing structure rather than fixing a bug.

**Independently worth keeping in mind.** Two findings here do not depend on which pricing model wins:

1. Seats were counted over `ProjectMembers` — a table the seat-selling operations never write. `POST /api/tenant/invite` creates an `AppUser`, so the meter read a different table from the one the metered operation touched.
2. `AdminController` compared *total active accounts* against `tenant.MaxUsers`, so a Studio tenant was capped at 6 accounts of **any** kind and a seventh read-only account was refused.

If either is still true on `main` after #617, each is worth its own PR that changes how seats are counted without changing what the caps are.

Closed as superseded, not as wrong.

PR #608, opened 2026-08-06
fix(admin): stop a tenant locking itself out of administration

StingD85:
**Stacked on #607** (which is stacked on #606). Merge order: #606 → #607 → this.

## The bug

Every admin surface is `[Authorize(Roles = "Admin,Owner")]`. Demote or deactivate the **last** account holding one of those roles and nobody can administer the tenant again — **including undoing the change that caused it**. There is no self-service recovery; it takes a manual database edit by the platform operator.

`PUT /api/admin/users/{id}` could do this in a single request, by either of two routes:

| route | before | after |
|---|---|---|
| `{ "role": "Viewer" }` on the last admin | ✅ succeeds → locked out | ❌ refused |
| `{ "isActive": false }` on the last admin | ✅ succeeds → locked out | ❌ refused |
| either, while another admin remains | ✅ succeeds | ✅ still succeeds |

## Why it is a separate PR

I found this while building #607 and reported it there rather than folding it in, because it changes behaviour on a path that PR only extended.

It is not academic. #607 makes demotion the seat-management tool — *"free a seat by changing an existing author to Viewer"* — so the obvious way to release a seat was also the way to brick the tenant. The two changes are related enough to sequence and separate enough to review independently.

`TenantAdminController.RemoveUser` already refuses to delete a `UserRole.Owner`, so the delete route was partly covered; demote and deactivate were not. `AdminController` has no delete endpoint.

## RED before GREEN

| | result |
|---|---|
| before | **Failed 2, Passed 1** |
| after | **Failed 0, Passed 3** |
| full suite | **Failed 0, Passed 606**, Skipped 9, Total 615 |

The passing test in the RED run is the **control** — demoting an administrator while another remains already worked and must keep working. A guard that refused *every* demotion would satisfy both lockout tests while making role management impossible; the control is what rules that out. The lockout tests also assert the administrator count is unchanged, not merely that the response was non-2xx.

### One detail worth knowing about the tests

The two lockout tests restore the seeded administrator through the `DbContext` in a `finally`. That is not tidiness. They are written against *intended* behaviour, so in the RED run the calls **succeed** and the tenant genuinely becomes unadministrable — which then breaks authentication for every later test in the class (my first RED run showed 3 failures for exactly this reason, not 2). Recovering through the API is impossible by definition, which is the bug, so recovery has to bypass it. Once the guard exists the restores are no-ops.

## Scope

One guard, one private helper (`IsAdministrator`, mirroring the `[Authorize]` roles; an inactive account cannot sign in, so it does not count). No response shapes, routes, prices, caps or quota axes touched.

CI’s Build & Test / CI Gate is broken on main itself (Postgres service container, "role root does not exist"), so verification is local. Not merged, per the standing instruction.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/AdminController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/AdminController.cs
@@ -40,6 +40,14 @@ public class AdminController : ControllerBase
     /// accounts against <c>tenant.MaxUsers</c>, which capped a Studio tenant at
     /// 6 accounts of ANY kind — a seventh read-only account was refused.</para>
     /// </summary>
+    /// <summary>
+    /// Can this account reach the admin surfaces? Mirrors
+    /// <c>[Authorize(Roles = "Admin,Owner")]</c>, and an inactive account cannot
+    /// sign in, so it does not count.
+    /// </summary>
+    private static bool IsAdministrator(UserRole role, bool isActive)
+        => isActive && (role == UserRole.Owner || role == UserRole.Admin);
+
     private async Task<string?> AuthorSeatRefusalAsync(UserRole role, CancellationToken ct = default)
     {
         if (!ProjectRoles.CanAuthorInformation(role)) return null;   // viewers are free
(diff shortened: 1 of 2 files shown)


StingD85:
Superseded by #615, now merged to `main`.

#615 carries this same last-administrator guard, salvaged onto `main` without the seat-metering stack underneath it. RED/GREEN was re-proven independently on that branch rather than inherited: guard removed → Failed 2, Passed 1; guard in place → Failed 0, Passed 3, including the control that demotion still succeeds while another admin remains.

The `AuthorSeatRefusalAsync` / `IQuotaGuardService` code that surrounded the guard here was deliberately left behind — it belongs to the role-based seat axis that the licence-based entitlement decision supersedes.

Closing this leaves #606/#607 as the remaining stack members. They are still blocked: the saturating `TotalSeats` overflow requirement must be captured somewhere it cannot be lost before they close. The PR intended to carry it (referred to as #616) was described but never opened.

StingD85:
Superseded by #615, which merged at 2026-08-06T02:18:28Z as `9a05f918d`.

Verified rather than assumed: this PR and #615 touch the same two files, and `LastAdministratorGuardTests.cs` is byte-identical between this branch and `main`. The remaining +54/-9 in `AdminController.cs` is **#607's** billing work (`IQuotaGuardService` injection, `AuthorSeatRefusalAsync`) inherited through the stack — this PR sits on `claude/viewers-free-author-seats` — not anything unique to the last-admin guard.

Closing so the queue reflects reality. The guard itself is on `main`; nothing is lost. #606 and #607 remain open on their own merits.

PR #614, opened 2026-08-06
docs(roles): propose four role fields → three, and the sheet that resolves them by hand

StingD85:
**Proposal only — nothing here runs.** The one executable artefact, `tools/role-reconciliation-sheet.sql`, is read-only.

## No revenue risk — stated up front

Entitlement is the **StingTools licence** (signed, offline, machine-bound). Billing reads `ProjectRole`, `Iso19650Role` and `AppUser.Role` **neither before nor after** this change, so it cannot mis-bill anyone. The blast radius is **permissions only** — which is exactly why it is being done by hand.

## The root cause, named

`ProjectMember.Iso19650Role` holds values from **two unrelated vocabularies**:

| Vocabulary | Source | Codes |
|---|---|---|
| ISO 19650 **role** | `ProjectMembersController.GetRoles()` | `A PM BC BA AR SE ME CE QS CA CT SC FM OM CL M V Z` |
| STING **discipline** | `ASS_DISCIPLINE_COD_TXT`, `TAG_CONFIG_v5_0_DISC_SYS_FUNC.csv` | `A E FP H LV M MG P RP S` |

They overlap on **exactly two codes**, and those two are the most common values in the data: **`A`** = Appointing Party *or* Architectural; **`M`** = Model Author *or* Mechanical.

This also retires an earlier wrong call of mine: I reported `S` and `EL` as invalid role codes. **`S` is Structural** — a correct discipline code that only looked wrong because it sat in a column being read as a role. Giving discipline its own field fixes the collision at the root rather than papering over it.

## Why no auto-migration

Measured on dev, 34 member rows:

| Confidence | Rows |
|---|---|
| **AMBIGUOUS** (`A` or `M`) | **20 (59%)** |
| HIGH | 13 |
| REVIEW (`EL`) | 1 |

Any automatic rule silently mis-resolves the **majority**, and capability derives from role, so a wrong resolution is a wrong permission. A cleverer rule would not help either — `Contributor + A` on *"Architectural Lead"* is plainly the discipline, while `Manager + A` on *"BIM Coordinator"* is genuinely arguable. Only someone who knows these people decides that.

## Target shape — four fields to three, not two

| Field | Purpose | Change |
|---|---|---|
| `AppUser.Role` | Authorization | **unchanged** |
| `ProjectMember.Iso19650Role` | ISO role: responsibility + capability | single role field |
| `ProjectMember.Discipline` | Trade | **new** |
| `ProjectMember.ProjectRole` | — | retired after backfill |

`AppUser.Role` stays because every admin surface is `[Authorize(Roles = "Admin,Owner")]`; making it an ISO code means rewriting authorization across the API — separate change, separate risk. `AppUser.Iso19650Role` carries the same ambiguity and is a follow-up.

## Sequence

Additive schema (plus a `RoleMigrationBackup` column holding the pre-migration pair, so it is reversible) → dual-write → generate the sheet **against production** → backfill from the **approved** columns only, refusing to run on a half-approved sheet → drop `ProjectRole` in a later migration.

The capability map is explicit and testable, with a test asserting every code `GetRoles()` serves has exactly one entry — so adding a role to the API without deciding its capability fails the build rather than defaulting to allow or deny.

## Honest limits

**Production has not been measured.** Every count here is the local dev database; the sheet must be regenerated against production before anything is approved. `EL` is deliberately **not** auto-corrected to `E` — a plausible typo fix is still a guess about a permission, so it goes to the sheet as REVIEW.

Not merged, per the standing instruction.

### DIFF
--- a/Planscape.Server/tools/role-reconciliation-sheet.sql
+++ b/Planscape.Server/tools/role-reconciliation-sheet.sql
@@ -1,0 +1,104 @@
+-- role-reconciliation-sheet.sql — one row per project member, for hand-resolution.
+--
+-- READ-ONLY. Produces the sheet the product owner fills in before any role
+-- migration runs. Nothing here writes.
+--
+--   docker exec -i docker-postgres-1 psql -U planscape -d planscape \
+--     -f - --csv < tools/role-reconciliation-sheet.sql > role-reconciliation.csv
+--
+-- ── WHY A SHEET AND NOT AN AUTOMATIC RULE ───────────────────────────────────
+-- ProjectMember.Iso19650Role holds values from TWO different vocabularies:
+--
+--   ISO 19650 ROLE (ProjectMembersController.GetRoles)
+--     A PM BC BA AR SE ME CE QS CA CT SC FM OM CL M V Z
+--   STING DISCIPLINE (ASS_DISCIPLINE_COD_TXT, TAG_CONFIG_v5_0_DISC_SYS_FUNC.csv)
+--     A E FP H LV M MG P RP S
+--
+-- They overlap on exactly TWO codes, and those two are the most common values
+-- in the data:
+--     "A" -> Appointing Party (role)  OR  Architectural (discipline)
+--     "M" -> Model Author    (role)  OR  Mechanical    (discipline)
+--
+-- A stored "M" cannot be resolved by looking at it. Capability is derived from
+-- the role, so guessing wrong grants or denies real permissions. Hence: propose,
+-- flag confidence, and let someone who knows these people by name decide.
+--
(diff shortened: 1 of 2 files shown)


StingD85:
Closing: the content of this PR is now on `main` via #782, which landed `docs/ROLE_MODEL_MIGRATION_PROPOSAL.md` and `Planscape.Server/tools/role-reconciliation-sheet.sql` as part of a sweep of stale-but-clean branches.

That sweep should have excluded branches with an open PR, and did not — this one was landed without going through its own review. Apologies for the detour; the proposal itself is unchanged from what was proposed here.

Verified: merging this branch into `main` now produces a tree byte-identical to `main`, so there is nothing left in it.

PR #618, opened 2026-08-06
fix(viewer): recompute near/far every frame so the model survives zooming out

StingD85:
## Summary

`near` and `far` were set **once**, by `fitCamera`, from the framed bounds. Dollying out then pushed the camera past that static far plane and **the model vanished after a modest scroll-out**. It was worst with `zoomToCursor`, which retreats along the pointer ray and can leave the model off-axis, so the far plane went wrong sooner than a straight dolly would suggest.

`updateClipPlanes()` now derives both planes from the camera's **current** distance to the model bounds, called from the animate loop and from `fitCamera`:

- `far = max(dist + radius*3, radius*8)` — clears the **back** of the model from wherever the eye is, with slack for pins, section box and the sky/ground helpers.
- `near` stays in front of the **front** of the model (`dist - radius` is the closest geometry can be) while remaining small enough to walk inside a building.

The renderer uses a logarithmic depth buffer (verified still set in this file), so the wide `near..far` span costs no z-precision. Ortho spans symmetrically (`near = -far`) so geometry behind the eye plane still draws. The projection matrix is rebuilt only on a meaningful change (>1% far, >25% near) to avoid per-frame churn.

## Provenance — worth knowing

This was **recovered from an uncommitted working tree** during a worktree audit. `git log --all -S "ZOOMFIX"` found no commit anywhere: the only copy was a dirty checkout that a prune would have destroyed. It's committed here so it can't be lost again.

The branch was 49 commits behind `main`, and `viewer.html` had changed there since (including the Phase 195 rollback/restore). `main` is merged in — **no conflicts** — and the fix re-verified afterwards: `ZOOMFIX` and `updateClipPlanes` present in both copies, the two `viewer.html` files byte-equal for the CI gate, and every symbol it leans on (`fitCamera`, `modelBounds`, `updateOrthoFrustum`, `logarithmicDepthBuffer`) still present.

## ⚠️ Not verified in a browser

Nobody has watched this behave. Merging on the author's explicit instruction, with that stated plainly rather than implied.

What to check once deployed:
1. Frame a model, then scroll **out** well past the initial framing — the model must stay visible instead of disappearing.
2. Scroll out with the cursor off to one side (`zoomToCursor` path) — the original worst case.
3. Walk **inside** the geometry — the near plane must still clip sanely, not slice the room away.
4. Switch to an orthographic view and repeat 1.

## Test plan

- [x] Merged `main` cleanly; no conflicts
- [x] `ZOOMFIX` + `updateClipPlanes` present in both `viewer.html` copies after the merge
- [x] Source ↔ wwwroot **byte-equal** (the CI gate)
- [x] All referenced symbols still exist in the post-merge file
- [ ] **Browser behaviour — unverified** (steps above)


StingD85:
**Verified in a browser — 2026-08-06.** Owner confirms zoom behaves correctly on the live deploy. The "browser behaviour — unverified" box in this PR's test plan is now closed.

The regression this prevents: `near`/`far` were set once by `fitCamera`, so dollying out pushed the camera past a static far plane and the model disappeared. Recomputing both planes per frame from the camera's current distance fixes it.

Deployment confirmed the same day — the served `viewer.html` carries 2 `ZOOMFIX` markers and 3 `updateClipPlanes` references.

Worth remembering how close this came to being lost: it was recovered from an uncommitted working tree during a worktree audit, existed in no commit anywhere, and a prune would have destroyed it permanently.

Issue #554, opened 2026-08-02
test: 2 CsiMasterFormatTests failing on main since 2026-06-20 — NormalizeSection changed from collapse- to remove-whitespace, tests never updated


StingD85:
## Summary

Two tests in `StingTools.Tags.Tests.CsiMasterFormatTests` have been **failing on `main` since 2026-06-20** — roughly six weeks.

They are not flaky and they were not born red. They passed when written, and a later deliberate behaviour change to `CsiMasterFormat.NormalizeSection` broke them without the tests being updated to match. The interesting part is that **the implementation change looks correct and the tests look stale** — but the change also altered user-facing output in a way that may not have been intended, so this should not be closed by simply rewriting the assertions.

Filing only. Not fixed here.

## The failures

```
StingTools.Tags.Tests.CsiMasterFormatTests.NormalizeSection_collapses_whitespace_and_uppercases
  Assert.Equal() Failure: Strings differ
  Expected: "23 31 00"
  Actual:   "233100"
  CsiMasterFormatTests.cs:72

StingTools.Tags.Tests.CsiMasterFormatTests.Reconcile_finds_gaps_overspec_and_title_mismatches
  Assert.Equal() Failure: Strings differ
  Expected: "22 40 00"
  Actual:   "224000"
  CsiMasterFormatTests.cs:95
```

One root cause, surfacing twice — `Reconcile` keys its result on the normalised section, so it inherits the change.

## When it broke — measured

| Commit | Date | Result |
|---|---|---|
| `1aea18cda` — *Phase 192 C2: SpecLink / CSI MasterFormat cross-reference* | 2026-06-12 | **8 passed, 0 failed** |
| `2c64c8351` — *ACC + CSI integration hardening (#332)* | 2026-06-20 | **6 passed, 2 failed** |

`1aea18cda` added `CsiMasterFormat.cs` and `CsiMasterFormatTests.cs` in the same commit, and they were green. `2c64c8351` changed the normaliser. The test file has had **no commits since it was created**, so it still asserts the original contract.

## What changed

`StingTools/Core/Classification/CsiMasterFormat.cs:113-121`:

```csharp
/// <summary>Canonical key for a CSI section number. Removes ALL whitespace (and
/// upper-cases) so spaced "23 05 00" and unspaced "230500" reconcile to the same
/// key — SpecLink exports spaced, models often store unspaced. Dots are preserved,
[...]

PR #623, opened 2026-08-06
feat(kut): one LOD ladder (Deliverable D = LOD 500), the missing CSI divisions, Owner defaults

StingD85:
KUT project-readiness pass. Our role is **information management, coordination and verification — not authoring**, and the Owner's environment is **CSI MasterFormat + RIB SpecLink + Fohlio**. Nothing in the contract requires ISO 19650 naming on Owner deliverables, COBie, or IFC, and none was added.

---

## The decision that needs to be on the record: Deliverable D is LOD 500

The client's **A1 Design Scope of Services states LOD 400 for Deliverable D. We are treating that as an error in the client document** and verifying Deliverable D at **LOD 500**.

| Deliverable | LOD | Milestone id |
|---|---|---|
| A — schematic | 200 | `deliverable-a` |
| B — 50% documents | 300 | `deliverable-b` |
| C — 100% documents | 350 | `deliverable-c` |
| Conformed set | 350 | `conformed-set` |
| Construction supervision (Work Program 3.1) | 400 | `construction` |
| **D — record / as-built** | **500** | `deliverable-d` |

LOD 400 is *fabrication and installation* maturity — the state a model is in while the work is being built. LOD 500 is *field-verified as-built*, which is by definition what a record/handover model is. A record model held at 400 would not have to carry the installed serial numbers, installation dates or maintenance data that make it useful to the Owner's FM team — the entire point of the deliverable.

**LOD 400 was not discarded.** It is now reachable as the `construction` milestone, covering Work Program item *3.1 Supervise the Building Construction Contract*. Nothing that was verifiable before became unverifiable.

**Action for the project team:** raise this at the next BEP review so the contract record and the verification gate agree. If the Owner confirms 400 is intended, set `deliverable-d` back to `"lod": 400` in `_BIM_COORD/lod_matrix.json` — the project overlay wins over the corporate baseline, so it is a one-line project decision needing no code change.

---

### 1. One LOD system, and a complete ladder

Two competing LOD systems existed. The deliverable-keyed `STING_LOD_MATRIX.json` + `LodVerificationEngine` was kept; the RIBA-keyed `LOD_REQUIREMENTS.json` + `LODValidationCommand` scored the same elements by its own heuristic and gave a second answer.

`LODValidationCommand` turned out **not** to be unreachable — `StingDockPanel.xaml:2573` has a live "LOD Check" button — so per the brief it was **not deleted**. It now delegates to `LodVerificationEngine`, keeping its `StingResultPanel` output and the one thing the matrix does not cover: the `STING_LOD_*_VISIBLE` family-switch audit (a visibility concern, not a maturity one). `LOD_REQUIREMENTS.json` had no other reader and was deleted.

The ladder is now the full standard **100/200/300/350/400/500** across all 20 `categoryRules`. This was structural, not cosmetic: `Resolve` returns null for a missing rung and `Verify` then does `continue`, so a milestone at an undefined rung would have **silently skipped every element** instead of failing loudly.

**Parameter names were verified, not assumed.** The brief's suggested `ASS_SERIAL_NUM_TXT` and `ASS_INSTALL_DATE` **do not exist**, and `ASS_INSTALL_DATE_TXT` is explicitly marked `DEPRECATED` in `MR_PARAMETERS.txt`. LOD 500 uses `ASS_INSTALLATION_DATE_TXT` + `ASS_SERIAL_NR_TXT` + `MNT_TYPE_TXT`; all nine parameters referenced by the matrix are declared with a GUID and bound `<ALL>` in `RESOLVED_BINDINGS.csv`.

**Fewer false positives at LOD 300.** `placeholderFamilyPatterns` contained `(?i)generic` and `(?i)\bdefault\b`, which match Revit's own stock naming ("Generic - 200mm", "Default") — the first Deliverable-B run would have drowned in noise. Both removed; `requireTypeNotGeneric` is untouched, so a "Generic - 200mm" wall still fails LOD 300 once, for the right reason, instead of twice.

### 2. CSI map — six missing divisions, and the reason half of it never fired

Added **02 · 05 · 14 · 31 · 32 · 33** (86 → 148 rules, 20 divisions). Without them `CSI_Assign` left every civil, structural-steel and elevator element unresolved.

While testing, a **pre-existing defect** surfaced that would have made much of this dead on arrival: `GetFamilyName`/`GetFamilySymbolName` return `""` for anything that is not a `FamilyInstance`, and `CsiRule.Score` treats empty as no-match. Every family/type discriminator on a **system** category was unmatchable — the shipped `Walls (?i)masonry|block|brick → 04 20 00` and `(?i)concrete → 03 30 00` rows had **never once fired**, and every wall silently took the 09 29 00 Gypsum Board default. Fixed with `CsiMap.TypeName` (a no-op for loadable families); affected discriminators, including those two shipped rows, moved to the `TypeRegex` column.

### 3. Owner defaults

- KUT classification **Uniclass → CSI**.
- **IPC plumbing code** is a Revit Project Information value no repo file can set, so it is documented as a week-1 step. `DrainageSizer.ResolveCode` and `VentDesigner` route any `IPC*` value to `IPCSiAdapter`; **anything else — including blank — silently falls back to BS EN 12056**, producing UK drainage and vent sizes that look valid and are wrong.

### 4. Two orphaned Fohlio commands surfaced

`Fohlio_ExportFinishes` / `Fohlio_ImportFinishes` were wired in the handler and `WorkflowEngine` but had no button and no workflow — unreachable, despite A1 Deliverables B and C both requiring room finishes in Fohlio. Added beside the existing Fohlio trio, with tooltips noting they match by **Room Number, not by tag**.

---

## Not done — already fixed upstream

The brief asked for a `CsiMasterFormat.FormatSection` display formatter and a fix for 2 failing `CsiMasterFormatTests`. **Both had already landed on `main` in PR #554.** `FormatSection` exists, `Reconcile` emits it, the XLSX writer and TaskDialog consume it, and the tests were already reconciled. Baseline measured **243 passed / 0 failed**, not the 239/2 the brief expected. No change was made.

---

## Verification

- `dotnet build -t:Rebuild` → **0 errors / 0 warnings**
- `StingTools.Tags.Tests` → **243 passed / 0 failed**
- LOD matrix validated field-by-field against the `LodCheck`/`LodMatrix` C# classes; its inheritance and `"+param"` merge simulated to confirm **every milestone resolves for every category with no silent skips**; placeholder patterns checked against 7 Revit stock names (all pass through) and 6 real placeholders (all caught)
- CSI map replayed through a port of `ParseCsvLines`/`Score`/`Resolve` over **44 cases**, including system elements passed with an empty family exactly as the real code does — 0 failures, no malformed section numbers, no dropped rows
- Workflow wiring diffed against `origin/main`: **66 unresolved `commandTag`s and 59 malformed steps exist on both** — pre-existing, now logged as ROADMAP KUT-1/KUT-2. All 5 `WORKFLOW_KUT_*` presets are clean on both.

**Not verified in Revit.** Nothing here was exercised in a live Revit session — no runtime verification was performed.

## Flagged, not fixed

Ten items added to `docs/ROADMAP.md` as **KUT-1 … KUT-10**, including the two dead-workflow findings above, the fact that project-scope `LOD_Verify` never sees categories without an explicit rule (Roofs, Ceilings, Stairs, Foundations…), the CFM/GPM unit-system gap, the hardcoded BS 7671 electrical path, and the `ProjectStandardsManager` disconnect.


StingD85:
## Amendments — A1 (scope hole), A2 (Division 02), A3 (rung 500)

Pushed as `00cfeab22`. Build `-t:Rebuild` **0/0**, tests **243/243**.

---

### A1 — the LOD gate was silently narrowing its own scope (closes ROADMAP KUT-3)

`LodScope.Collect` built project scope from `ExplicitCategories`, which filters out `"*"` — so the fallback the matrix defines precisely so nothing escapes **never applied outside a manual selection**. Never scanned: Roofs, Ceilings, Stairs, Railings, Ramps, Furniture, Furniture Systems, Structural Foundations, Electrical Fixtures, Duct/Pipe Fittings, Curtain Panels/Mullions, Rooms.

The gate then reported a confident per-category percentage **without disclosing what it had declined to look at**. Ceilings, stair finishes, railings and millwork are what the Owner's panel scrutinises hardest; Furniture and Furniture Systems are two of the six Fohlio FF&E categories, so Deliverable D could show a green LOD gate over a furniture register nothing had verified.

**Three fixes:**

1. **`"*"` is reachable.** `Resolve` already falls through to `"*"` for an unmatched category (verified by reading it — no engine change needed), so only the collector needed widening. Blast radius constrained three ways: `CategoryType.Model` (drops annotation/view content), membership of `TagConfig.DiscMap` (the tagging pipeline's own taggable-category definition, so LOD scope and tag scope agree), and a documented `NonPhysicalCategories` set (Materials, RVT Links, Model Groups, Analytical\*, Loads, Parts, Areas, Zones, Property Lines). The project's `TagConfig.CategorySkipList` is honoured. A category with an explicit rule is **always** collected. If the matrix defines no `"*"` rule, the old explicit-only filter is kept.
2. **14 new explicit category rules** so those categories get a real ladder instead of leaning on `"*"`. Ceilings/Railings/Ramps get a lighter ladder (the retired `LOD_REQUIREMENTS.json` held them at 300 vs 350 for Roofs/Stairs — that distinction was carried across). Furniture and Furniture Systems follow the FF&E pattern so they line up with Fohlio. Rooms gets a spatial ladder with no manufacturer/model requirement, which would be nonsense on a room.
3. **Scope disclosure in all three output forms** — `LodScopeReport.DisclosureLines()` feeds the TaskDialog, the CSV (`#` header lines before the data) and the JSON gate report (a `scope` object). Each states elements verified, categories riding `"*"`, and **NOT SCANNED** categories with counts.

**Before/after** on a synthetic 3,314-element model:

| | in scope |
|---|---|
| BEFORE (explicit categories only) | **879** |
| AFTER (taggable model categories) | **2,789**  (+1,910) |

All 14 previously-invisible categories now covered by explicit rules; 525 non-physical elements **disclosed** as not scanned rather than dropped; 1,420 annotation elements correctly never collected.

---

### A2 — Division 02 **withdrawn** (amends ROADMAP KUT-5)

I took the **withdraw** option, not phase awareness. Both reasons were real:

- **Reason 1 (mechanical):** the discriminator sat in `FamilyRegex`, and `GetFamilyName` returns `""` for non-`FamilyInstance` — a `TopographySurface` could never match.
- **Reason 2 (conceptual):** `Resolve` is never handed `Phase Demolished`, and nobody names a toposolid "demolition". Moving the regex to `TypeRegex` makes the row *capable of firing* without making it *fire*.

So all four Division 02 rows are gone and the CSV records why. Phase awareness is the real fix, but it widens every row from 6 to 7 fields and `ParseCsvLines` silently drops short rows — **every project's existing `_BIM_COORD/csi_map.csv` overlay would stop loading**. That needs its own change with a backward-compatible parser and its own verification pass. KUT-5 is amended with that design note.

**Reason 1, general fix:** took the **TypeRegex** branch, not a new `CsiMap.FamilyName`. For a system element that would return the *system-family* name ("Basic Wall", "Pipe Types") and re-score **every** system-category rule in the map for no current benefit. Instead the constraint is documented in the CSV header.

**Audit of all 144 rows — asked for explicitly, including "nothing else":** **zero** FamilyRegex rows remain on a system category. Three sit on `Structural Foundations`, which is genuinely mixed as you noted (isolated footings `FamilyInstance`, wall footings `ContinuousFooting`); they target piles, which *are* `FamilyInstance`, and the bare category default catches continuous footings. `Structural Framing`, `Structural Columns`, `Site`, `Parking`, `Planting` confirmed `FamilyInstance` and fine.

Map is now **144 rules across 19 divisions**. Resolution harness re-run over **50 cases, all passing**, including system elements passed with an empty family exactly as the real code does.

---

### A3 — rung 500 was unsatisfiable on building fabric

`+ASS_INSTALLATION_DATE_TXT` was required at 500 on **every** category. Rung 500 now splits:

- **Serialised, serviceable plant** — Mechanical/Electrical Equipment, Plumbing/Lighting Fixtures, Air Terminals, Sprinklers, Fire Alarm Devices, Specialty Equipment: keeps `+ASS_SERIAL_NR_TXT` + `+ASS_INSTALLATION_DATE_TXT`.
- **Fabric and distribution** — Walls, Floors, Doors, Windows, Ducts, Pipes, Conduits, Cable Trays, Structural Framing/Columns/Foundations, Casework, plus all new A1 categories: `{ "inherit": "400" }`, no added params. The rung still carries 400's manufacturer/model requirements.
- **`"*"` follows fabric** — which matters more now that A1 makes the fallback apply to far more elements.

**Doors and Windows → fabric.** They carry manufacturer and model at 400 and are schedule-tracked, but they are not serialised plant and a contractor's as-built does not deliver a serial per door.

Both `STING_LOD_MATRIX.json` and the KUT overlay have their `description` rewritten to explain the split — that string is the only place a future reader learns *why* rung 500 differs by category.

---

### Verification

- Build `-t:Rebuild` **0 errors / 0 warnings**; `StingTools.Tags.Tests` **243/243**
- Both LOD JSONs parse; **34 categoryRules**, full 6-rung ladder each; all **9** parameters declared with a GUID in `MR_PARAMETERS.txt` and bound `<ALL>` in `RESOLVED_BINDINGS.csv`
- Inheritance simulated against the real `ResolveCheck`/`MergeList` logic: **every milestone resolves for every category, no silent skips**, and rung 500 confirmed as bare-inherit on fabric vs serial+install on plant
- CSI harness: **50 cases, 0 failures**; row audit for the FamilyRegex defect
- CHANGELOG Phase 228 entry **updated in place** (not duplicated); ROADMAP **KUT-3 closed**, **KUT-5 amended**, **KUT-11 added** for the deliberately-unfixed family-side fallback

**Nothing was exercised in a live Revit session** — the scope widening, the disclosure output and CSI assignment against real elements remain unproven at runtime.

Part B (KUT-1/KUT-2, the 11 presets that execute nothing) is a separate branch and PR, as briefed.

Issue #616, opened 2026-08-06
Carry forward from #606: seat-total arithmetic must saturate, or a tenant on an unlimited plan is created with a negative user cap


StingD85:
Captured so it survives the closure of #606/#607, whose author/coordinator seat split is superseded by the licence-based model. **The fix is not independently salvageable** — `TotalSeats` is itself part of that split (`Tenant.cs`, `QuotaGuardService`, `AuthController` ×2, plus 3 test files), so extracting it would re-introduce the model the licence decision replaces. It must be **re-fixed against the licence-based model** when that is built.

## The defect

`Tenant.MaxUsers` is derived at tenant creation as:

```csharp
MaxUsers = planLimits.MaxAuthors + planLimits.MaxCoordinators;   // AuthController:463 and :1098
```

C# arithmetic is unchecked by default, so once **either** side is `int.MaxValue` the sum wraps **negative**:

```
          1 + int.MaxValue  =  -2,147,483,648
          6 + int.MaxValue  =  -2,147,483,643
int.MaxValue + int.MaxValue =  -2
```

`MaxUsers` is then consumed as `userCount >= tenant.MaxUsers` (`AdminController:78`, `ProjectMembersController:256`). Against a negative cap that is true at `userCount = 0`, so **the tenant is refused its very first user** — with the message `User limit (-2) reached`, which points at nothing.

## Exposure today — latent, not live

Stating this precisely, because an issue that overstates its own severity gets dismissed when someone checks it.

**Measured on the dev database — no tenant currently has a broken cap:**

| Slug | Plan | MaxUsers | cap ≤ 0 |
|---|---|---|---|
| demo | 5 | 50 | no |
| e2e-aad60d3818d5 | 0 | 20 | no |
| e2e-e8582c5c35c4 | 0 | 20 | no |
| planscape | 5 | 2147483647 | no |

Two things follow:

1. **`planscape`’s `int.MaxValue` is not this bug.** `PlatformTenantSeeder.cs:57` assigns `MaxUsers = int.MaxValue` **directly** — a literal, not the sum, so no overflow occurs. And an enormous positive cap *permits*; it is a negative cap that denies.
2. **Both creation paths assign `BillingPlan.Trial`** (`AuthController:460` registration, `:1095` D1 handoff), and Trial is `(MaxAuthors 1, MaxCoordinators 0)` → `1 + 0 = 1`. No overflow.

[...]

PR #626, opened 2026-08-06
refactor(quota): retire the per-user quota axes — D1 owns seat entitlement

### LINKED ISSUE #619: QuotaGuardService: the Authors seat axis counts a role string nothing writes, so it is always 0
Found while building licence presentation (D1 model). Live on `main` today, and **untouched by the closed #606/#607** — those PRs died around this, not on it.

## The defect

`Planscape.Server/src/Planscape.Infrastructure/Services/QuotaGuardService.cs:76`

```csharp
QuotaAxis.Authors => await _db.ProjectMembers
    .Where(m => m.TenantId == tid && m.ProjectRole == "Author")
    .Select(m => m.UserId).Distinct().CountAsync(ct),
```

Nothing in the server ever writes `ProjectRole == "Author"`. Every assignment in `Planscape.Server/src`:

| Site | Value written |
|---|---|
| `SeedData.cs:275` | `"Coordinator"` |
| `SeedData.cs:304` | `"Contributor"` |
| `AuthController.cs:1231` | `"Owner"` |
| `ProjectsController.cs:172` | `"Manager"` |

`ProjectMember.ProjectRole` is documented as Viewer/Contributor/Coordinator/Manager (`Planscape.Core/Entities/ProjectMember.cs:15-16`).

So `QuotaAxis.Authors` evaluates to **0 forever**, and since `BillingPlanLimits` gives `MaxAuthors = 1` for every plan except Enterprise (`Tenant.cs:177-183`), the author cap can never be reached.

## Why it can't work as written

The paths that *check* the author quota create an **`AppUser`**, never a `ProjectMember`:

- `OnboardingController.cs:93` calls `CheckCanAddUserAsync("Author", …)`, then creates an `AppUser` at `:99-109`
- `TenantAdminController.cs:108` does the same, creating an `AppUser` at `:125`

The counter reads one table; the writer writes another. They cannot converge by tuning the predicate — the fix has to make the counted thing and the written thing the same thing.

## Also affected

`QuotaAxis.Coordinators` (`:77`) is `ProjectRole != "Author"`, which — since nothing is ever `"Author"` — counts **every** project member including Owners and Managers.

The same logic is duplicated display-only at `TenantAdminController.cs:49-55`.

## Not covered by tests

`grep Quota` over `Planscape.Server/tests/Planscape.Tests` returns nothing. There is no `QuotaGuardService` test at all.

## Suggested shape of a fix

Whatever ends up being counted must be written by the path that sells the seat — the lesson from #606/#607. Count `AppUser` rows (which the invite paths actually create) filtered by `Iso19650Role`, or drop the ProjectMembers axis entirely. Add a test that asserts an accepted invite *moves the number the cap is checked against*, not that the endpoint returns 200.

For reference, the equivalent on the D1 side is now a single shared function used by both the spending and reporting paths: `marketing-site/functions/api/license/_lib/seats.ts`.

StingD85:
Retires the .NET per-user quota axes. Seat entitlement is the StingTools licence, recorded in D1 by `issue.ts` and counted by `_lib/seats.ts` (#621, merged). **Do not merge — for review.**

## Behaviour change

**Invites are no longer refused with 402 by the .NET server.** Both `POST /api/tenant/invite` and `POST /api/onboarding/team` previously returned `402 quota_exceeded`; they no longer can. Author entitlement is the D1 licence count.

`tenant.MaxUsers` is enforced separately and is **not** part of this retirement — see below.

## Why removed rather than repaired

`QuotaAxis.Authors` counted `ProjectMembers` with `ProjectRole == "Author"` — a literal none of the four writers in this codebase produce (`SeedData.cs:275/304` → Coordinator/Contributor, `AuthController.cs:1231` → Owner, `ProjectsController.cs:172` → Manager). `MaxAuthors` is 1 on every plan but Enterprise and was **never enforced**. `Coordinators`, its mirror (`!= "Author"`), counted every member including Owners and Managers.

### Measured, not inferred (RED)

A temporary characterisation test seeded four `ProjectMembers` using only the role strings the codebase actually writes, then asserted the defect against unmodified code:

| Assertion | Result |
|---|---|
| Control: members visible | `4` |
| `CheckCanAddUserAsync("Author")` | `Current=0, Max=1, Allowed=true` |
| `CheckCanAddUserAsync("Coordinator")` | `Current=4` (counts Owner + Manager) |

The control assertion is load-bearing: `PlanscapeDbContext`'s tenant filter falls back to `Guid.Empty`, so without first proving the four rows are visible, `Authors == 0` would have been indistinguishable from an empty result set. **Passed** — the defect is real.

That test is deleted in this PR. Its subject no longer exists, so it cannot survive the change it justified.

## Removed

- `CheckCanAddUserAsync` — `IQuotaGuardService` declaration and `QuotaGuardService` implementation
- The `Authors` / `Coordinators` arms of `CountAsync`
- The `Authors` / `Coordinators` arms of `QuotaAttribute`, plus the now-dead `projectRoleHint` constructor parameter
- The invite quota gate in `OnboardingController` and `TenantAdminController` (the `role` local is **kept** in both — it drives `UserRole` / `Iso19650Role` and the response rows)

## Kept

`QuotaAxis.Projects`, `QuotaAxis.Storage`, and the entire storage path. `[Quota(QuotaAxis.Projects)]` on `ProjectsController:116` still returns 402.

## Enum decision: DELETED, not left unused

`QuotaAxis` is now `{ Projects, Storage }`. A stray `[Quota(QuotaAxis.Authors)]` is a **compile error** — strictly stronger than the safe-default alternative, where an unused member would fall to the filter's `_ =>` arm and silently ALLOW, which is indistinguishable from a working cap until someone relies on it.

`Storage` is the only axis reaching the default arm, and allowing it there is correct rather than a fallback: the byte count isn't known until the body is read, so upload controllers call `CheckCanUploadBytesAsync` in the action body. Commented in place.

## Two things beyond the mapped blast radius

Both were required for the change to be honest, and both are flagged for review:

1. **`TenantAdminController`'s dashboard duplicated the same broken predicate** (`ProjectRole == "Author"`) for display at lines 48-55. Left in place, #619 could not be closed as "the broken predicate ceases to exist", and a billing page would keep showing a firm of four people "Authors 0/1". Removed from the `usage` payload; `limits` still advertises `MaxAuthors`/`MaxCoordinators` as plan metadata (they still feed `TotalSeats`).

2. **That is an API contract change with a live consumer.** `planscape-web/app/settings/team/page.tsx` renders `data.usage.authors` and `data.usage.coordinators` through a `Quota` component that reads `axis.max` directly — `undefined` would **throw on render**, not degrade. The two tiles and the two `lib/types.ts` fields are removed with it.

`inviteMessage`'s `402` branch in that page is left alone deliberately — it is defensive, costs nothing, and removing it is not required for correctness.

## `tenant.MaxUsers` — stays, but it is now mis-shaped

Not part of this retirement. It has **two** enforcement sites, not one:

- `ProjectMembersController:310` — invite-by-email, when creating a pending account
- `AdminController:86` — `POST /api/admin/users`

Both are `userCount >= tenant.MaxUsers` over active `AppUser` rows: a **total-account cap**, not a seat cap. It is derived from the same `BillingPlanLimits` this PR partly retires — `AuthController:463` and `:1098` set `MaxUsers = planLimits.TotalSeats`, which saturates (see `Tenant.cs:176-190`, #616).

**Recommendation: it stays for now, tracked separately — it should become an abuse ceiling decoupled from `TotalSeats`, not licence-derived.**

- It cannot be **licence-derived** today: that needs a .NET → D1 link that does not exist, and the D1 count meters *machines*, not accounts. Viewers hold no licence, so it could never bound them.
- It should not simply **go**: the invite endpoints create `AppUser` rows, and with the 402 gate gone `MaxUsers` is the last thing bounding account creation at all.
- But leaving it tracking `TotalSeats` keeps "viewers are free" half-true, exactly as you said. Studio is `MaxAuthors(1) + MaxCoordinators(5) = 6`, so a Studio firm is refused its 7th account — including a free viewer.

I have **not** filed this as an issue; say the word and I will.

## Verification

| | Result |
|---|---|
| RED (measurement on unmodified code) | 1 passed — defect confirmed |
| Baseline `dotnet test` on `origin/main` | **617 passed / 9 skipped / 626 total / 0 failed** |
| GREEN after removal | **617 passed / 9 skipped / 626 total / 0 failed** |
| `Planscape.API` build | 0 errors, 7 warnings (pre-existing NU1603 ×4 + CS0618 ×3) |
| `planscape-web` `tsc --noEmit` | 68 errors before **and** after — all pre-existing `Badge tone` / `Button variant` mismatches, none in the edited files |

No test referenced the removed paths, so the identical counts are expected rather than reassuring. The only quota test — `QuotaGuard_ProjectCap_*` in `SecurityCriticalPathTests` — exercises `CheckCanAddProjectAsync` / `QuotaAxis.Projects`, both kept, and still passes.

Known-broken CI (Build & Test / CI Gate failing on a Postgres service container, "role root does not exist") is broken on `main` itself and unrelated.

Closes #619.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/OnboardingController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/OnboardingController.cs
@@ -88,10 +88,10 @@ public class OnboardingController : ControllerBase
         var added = new List<object>();
         foreach (var m in req.Members)
         {
-            // Quota check per member — fail mid-list cleanly with a 402.
+            // Seat entitlement is the StingTools licence, counted in D1 (#621) —
+            // this endpoint no longer refuses with 402. `role` still drives the
+            // UserRole / Iso19650Role written below and the response rows.
             var role = string.Equals(m.Role, "Author", StringComparison.OrdinalIgnoreCase) ? "Author" : "Coordinator";
-            var q = await _quota.CheckCanAddUserAsync(role, ct);
-            if (!q.Allowed) return StatusCode(StatusCodes.Status402PaymentRequired, new { error = "quota_exceeded", added, denied = m, quota = q });
 
             var existing = await _db.Users.FirstOrDefaultAsync(u => u.Email == m.Email.ToLowerInvariant(), ct);
             if (existing != null) { added.Add(new { existing.Id, existing.Email, role, status = "exists" }); continue; }
--- a/Planscape.Server/src/Planscape.API/Controllers/TenantAdminController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/TenantAdminController.cs
@@ -4,7 +4,6 @@ using Microsoft.EntityFrameworkCore;
 using Planscape.Core.Entities;
 using Planscape.Core.Interfaces;
 using Planscape.Infrastructure.Data;
-using Planscape.Infrastructure.Services;
 
 namespace Planscape.API.Controllers;
 
@@ -13,9 +12,13 @@ namespace Planscape.API.Controllers;
 /// numbers a buyer needs to feel in control of their subscription:
 ///
 ///   • Plan, currency, trial expiry, billing-cycle
-///   • Live usage vs limits (authors / coordinators / projects / storage)
+///   • Live usage vs limits (projects / storage)
 ///   • Last-month revenue (if any payments) + next charge date
 ///   • Recent admin events (member invited, project created, plan changed)
+///
+/// Author / coordinator seat usage is NOT reported here. Seat entitlement is
+/// the StingTools licence, counted in D1 (#621); the counts that used to appear
+/// were derived from a ProjectRole predicate nothing writes (#619).
 ///
 /// All endpoints scoped to the resolved tenant by the global query filter
 /// from S1.1 — there is no path-parameter for tenant id so an admin can't
@@ -28,13 +31,11 @@ public class TenantAdminController : ControllerBase
 {
     private readonly PlanscapeDbContext _db;
     private readonly ITenantContext _tenantContext;
-    private readonly IQuotaGuardService _quota;
 
-    public TenantAdminController(PlanscapeDbContext db, ITenantContext tenantContext, IQuotaGuardService quota)
+    public TenantAdminController(PlanscapeDbContext db, ITenantContext tenantContext)
     {
         _db = db;
         _tenantContext = tenantContext;
-        _quota = quota;
     }
 
     /// <summary>Single dashboard payload — plan, usage, members, billing.</summary>
@@ -47,12 +48,6 @@ public class TenantAdminController : ControllerBase
         var limits = BillingPlanLimits.For(tenant.Plan);
 
         var memberCount = await _db.ProjectMembers.CountAsync(ct);
-        var authorIds = await _db.ProjectMembers
-            .Where(m => m.ProjectRole == "Author")
-            .Select(m => m.UserId).Distinct().ToListAsync(ct);
-        var coordIds  = await _db.ProjectMembers
-            .Where(m => m.ProjectRole != "Author")
-            .Select(m => m.UserId).Distinct().ToListAsync(ct);
         var projectCount = await _db.Projects.CountAsync(ct);
         var storageBytes = await _db.ProjectModels.AsNoTracking()
             .Where(m => m.DeletedAt == null)
@@ -81,8 +76,14 @@ public class TenantAdminController : ControllerBase
             },
             usage = new
             {
-                authors      = new { current = authorIds.Count,  max = limits.MaxAuthors },
-                coordinators = new { current = coordIds.Count,   max = limits.MaxCoordinators },
+                // `authors` and `coordinators` are gone. They were counted with
+                // the same ProjectRole == "Author" predicate as the retired
+                // quota axes — a literal none of this codebase's four writers
+                // produce — so `authors` always displayed 0 and `coordinators`
+                // displayed every member including Owners and Managers (#619).
+                // Seat usage now lives where seats are sold: the D1 licence
+                // count (#621). `limits` below still advertises what the plan
+                // includes; we simply no longer claim to measure it here.
                 projects     = new { current = projectCount,     max = limits.MaxProjects },
                 storage      = new { currentMb = storageBytes / 1024 / 1024, maxMb = limits.StorageMb },
                 memberSeats  = memberCount,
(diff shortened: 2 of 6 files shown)


StingD85:
## Do not merge until presentation is confirmed live

This PR removes the `402 quota_exceeded` gate. The replacement — licence presentation to D1 (#621) — is **not yet working in production**, so merging now leaves a window with neither enforcement.

Measured on production (2026-08-07):

```
POST /api/license/issue      -> 401  {"error":"Missing or malformed Authorization header"}
POST /api/license/present    -> 500  {"error":"Licensing is not configured."}
POST /api/does-not-exist-xyz -> 405
```

`present.ts` is deployed (it answers from its own code, distinct from the static 405 of a nonexistent route) but **`LICENSE_PRIVATE_KEY` is not set in Pages production**. Licence *issuing* is equally broken — `issue.ts` runs `requireAuth` before its key guard, so its clean 401 masks the same missing secret.

## Sequence to unblock

**1. Verify the key before uploading it.** The plugin carries the public half compiled in at `StingTools/Core/Licensing/LicensePublicKey.cs:7` and verifies with RSASSA-PKCS1-v1_5 / SHA-256 (`LicenseCrypto.cs:36`). So the test is whether the candidate private key pairs with the public key every deployed plugin already trusts:

```bash
openssl rsa -in /path/to/private.pem -pubout
```

Compare against the PEM literal in `LicensePublicKey.cs`.

- Identical → correct key, upload it.
- Different → **do not upload**. Licences it issues would be rejected by every installed plugin, trading a "licensing not configured" outage for an "invalid licence" one, which is worse and harder to diagnose.

**2. Set the secret** (runtime binding, no redeploy needed):

```bash
wrangler pages secret put LICENSE_PRIVATE_KEY --project-name planscape-marketing
```

**3. Re-probe.** `present` must leave 500, **and** test `issue.ts` with a real Authorization header — issuing is a separate outage the 401 has been masking.

**4. Confirm presentation lands.** Build and deploy the plugin to whatever the installed `.addin` actually points at (grep the manifest — do not trust the documented path), Revit closed and the Planscape Companion tray app stopped. Start Revit once, then:

```bash
npx wrangler d1 execute planscape-waitlist --remote --command "SELECT machine_code, last_seen_at, last_seen_plugin_version FROM licenses WHERE revoked_at IS NULL ORDER BY last_seen_at DESC LIMIT 5;"
```

All three `last_seen_*` columns were absent until the migration on 2026-08-07, so presentation cannot have worked before now — **any populated `last_seen_at` is unambiguously from that run.** No stale-value risk.

**5. Only then merge this PR.**

## Note on marketing-site deploys

`marketing-site` has **no git-connected Pages build** (#651) — merging to `main` deploys nothing. #621 sat undeployed for six days and Pages' static fallback made it indistinguishable from a nonexistent route. Re-run the three-way probe above rather than assuming a merge shipped.

## Follow-up once this merges

`tenant.MaxUsers` becomes the only bound on account creation and still caps free viewers — see #653.

PR #649, opened 2026-08-10
Parameter governance, ISO sheet numbering, and door/window type marks (K-11 · K-12 · K-13 · G-20)

StingD85:
Closes the parameter-governance workstream and the Kibale (KNP26) blockers. **Nothing here needs Revit to review.**

## The root cause, and the gate that stops it recurring

Seven separate defects this pass were one shape: **a consumer reads a parameter name the binder does not ship**, so the read returns null forever and nothing says a word.

`WARN_VISIBLE_BOOL` (missing prefix) · `PRJ_ORG_PROJECT_CODE` (missing `_TXT`) · three `PRJ_TB_SHOW_*` underscore pairs · the originator split · `TAG_3_BOLD_BLUE_BOOL` · `TAG_2_5_NOM_BLACK_BOOL`

Nothing validated the two sides against each other. `tools/validate_param_readership.py` now does, wired into CI as a **ratchet** — it fails on an *increase*, not on a non-zero count.

## What this fixes for the operator

- **41 `PRJ_*` were declared and read but never bound.** 36 now bound to Project Information. Until this, `PRJ_PROJECT_COD_TXT` did not appear in Manage → Project Information and **no correct sheet number could be produced**.
- **Sheet numbers fail loudly** instead of dropping a segment. `{project}`/`{originator}` are now *omitted* when unresolved (the `{seq}` rule), so Revit rejects the number rather than accepting `-PLN-COT01-…` on an issued drawing. **No placeholder is substituted** — a fabricated code produces a sheet that looks correct and is not.
- **Sheets carry their ISO segments individually**, so a title-block label can show any subset.
- **Door/window type marks generate** — monotonic, never reused, existing marks adopted not overwritten, PROD resolved through the tag pipeline so the mark can never disagree with the ISO tag.
- **Type schedules exist.** `IsItemized` was never set `false` anywhere in the plugin.

## Corrections to my own earlier findings

- Retiring `PRJ_ORIGINATOR_COD_TXT` was **wrong** and is reversed — a `.rfa` label is a consumer the C# scan cannot see.
- The MatchLine readers were **not** broken; all three already fell back to `sheet.SheetNumber`.
- "No sheet exists to mislead anyone" is **false** — the sheet is created before the number is assigned.

## Known open, deliberately

- **G-25** — `param_binding_resolver --apply` would orphan 10 live `SHT_*`. Run inspected and **reverted**; the tool is now `--check`-by-default so the drift is visible instead of arriving as a side effect.
- **3B.4** — 6 of 13 `pres-*` types carry `autoTag=True`. Not fixed: `STING_DRAWING_TYPES.json` is checksum-locked.

## Verification

```
build 0 errors / 0 warnings · schema self-test 8/8 · readership self-test 4/4
readership 377 / ceiling 384 PASS · path discipline Tier1 0 Tier2 0
content manifest 207/207 · binding sim 3238 bound / 163 skipped / 0 conflicts
```

See `docs/KNP26_READINESS.md` — ~11 minutes of setup blocks modelling; the Revit session does not.


StingD85:
Closing this in favour of three slices that are now on `main`. **The branch is deliberately not deleted** — it is the only copy of the pre-slice state.

## Why it could not be merged

Measured on the **merged tree**, not on `git diff main..branch`:

| | `main` | merged |
|---:|---:|---:|
| shared parameters | 3,593 | 3,600 |
| **native-typed (non-`TEXT`)** | **763** | **616** |

Merging takes **149 parameters back to `TEXT`** and drops their display mirrors. The branch predates Phase 188 and the conflicted regions span effectively the whole parameter file, so resolving by picking a side could not avoid it.

**Every check in CI stays green while it happens.** Nothing in CI reads the shipped parameter file's *types*. It surfaces only in Revit, on live models, as tag labels reading nothing and "Inconsistent Units" errors returning.

(The brief estimated 185 parameters; the measured figure is 149. Still catastrophic, still invisible.)

## What replaced it

| PR | What |
|---|---|
| **#843** | Define and bind the five parameters `ClassificationReader` reads |
| **#844** | The code, resolved against current `main` |
| **#845** | The engineering record — gap register, audits, and their tooling |

**#843 is the one worth reading.** `ClassificationReader.Read()` reads `UNICLASS_PR_TXT`, `UNICLASS_SS_TXT`, `UNICLASS_EF_TXT`, `NBS_CODE_TXT` and `ASSET_RFI_URL_TXT`, and **none of the five existed on `main`** — not in `MR_PARAMETERS.txt`, the `.csv` mirror, `CATEGORY_BINDINGS.csv` or `RESOLVED_BINDINGS.csv`. Every read returned `""`. The whole Uniclass / NBS / RFI axis was dead, silently, and "Uniclass 2015" is the **default** in the classification-standard picker — an order whose leading three rungs could never match. This branch had authored those rows; they were re-landed on `main`'s Phase-188 data through the generators.

## Absorption, by content

Ancestry is meaningless here — the slices were rebuilt onto `main`, so none of this branch's 115 commits is an ancestor of anything. By content, of the 509 files this branch changed:

- **114 byte-identical** on `main`
- **247 present but different** — 200 of them tag-family `.rfa`
- **5 absent**

## Deliberately not carried, with reasons

**The four parameter and binding files** — the Phase 188 revert above.

**206 tag-family `.rfa` binaries**, plus one new one (`STING_Tag_Universal.rfa`). A month old, unreadable and untestable from a terminal, and exactly the artefact class whose stale labels produce the "Inconsistent Units" error this repository has already chased once. A human with Revit can re-save them against current `main` if they are wanted; the branch keeps them until then.

**`FORMULAS_WITH_DEPENDENCIES.csv`, and with it `BlockCountG34Tests.cs` and `MasonryMortarG15Tests.cs`.** Both sides changed `CST_S_MAS_MORTAR_VOLUME_CU_M`: this branch **deletes** it (G-15/G-34, moving masonry mortar to a BOQ line), `main` **rewrites** it to use the brick-bond lookup and is actively using it in the material schedule. `main`'s is newer and in use. This is a genuine design disagreement, not a merge artefact, and it needs a decision rather than a resolution.

**`STING_CONTENT_MANIFEST.json`** — every `tagFamilies` entry loses its `buildSpec` field. **`LABEL_DEFINITIONS.json`** — older copy.

**6 files under `Planscape.Server` / `planscape-web`** (quota, tenant admin). Unrelated to the BIM work, and `main` has moved 146 files and ~14,700 lines in that area since.

**`PR_BODY_kibale_integration.md`** and **`g8_binding_audit.txt`** — working files left at the repository root.

## Two things found on the way that were worse than expected

**Git auto-merged a duplicate gate with no conflict marker.** Both sides had independently implemented *"Gate 7 — measured lines with no resolvable quantity"* in `BOQPrepForExportCommand`. Git merged them as two non-overlapping insertions: no marker, just a duplicate gate, a redeclared local and a tuple-arity mismatch. Had the arity matched it would have **compiled** and reported every gate twice. A conflict asks; this only failed to build.

**Two commands were dead from the dock panel.** `StingAutoTagger` and `StandardsEngine` read `commandData.Application` directly, which is null there — so both buttons logged start and done, showed nothing, and threw nothing. Caught by a gate `main` added after this branch was cut, and fixed in #844.

## One correction to my own analysis

I initially reported `cost_rates_5d.csv` as losing 7 rows. It does not. The branch **adds a `PROD` column**, so my key was comparing `MAT_CODE` against `PROD`. The 7 are renames to real Revit category names (`Duct Systems` → `Ducts`, `Pipe Systems` → `Pipes`), same codes, **zero rate changes** — a fix, since those names could never match a category. It was carried.

## Verified across the three slices

Build **0 errors / 0 warnings**; **1113** Boq tests pass; `check_kut_documents` 80 assertions, `check_path_discipline`, `check_workflow_wiring` and `check_command_doc_acquisition` all green. **Phase 188 typing asserted on `main` after each merge: 763 native-typed parameters, unchanged.**

**Nothing was run in Revit.** No runtime path in any of the three slices was exercised.

PR #676, opened 2026-08-16
Security: gate the federation controllers on project membership, not just tenant

StingD85:
## The bug (live)

Cross-**tenant** isolation is sound — the global `ITenantScoped` filter fails closed. But tenant is not the unit of authorization for federation data. Five controllers carried only `[Authorize]` and filtered by tenant, so **a user invited to any one project in a tenant could read and modify every other project in that tenant**:

| Surface | Exposure |
|---|---|
| `ModelTransformController` | `PUT/DELETE .../transform` — overwrite or delete another project's coordinate transform |
| `CoordinateSystemController` | `POST/PUT/DELETE .../coordinate-system` — redefine the CRS every model aligns against |
| `AlignmentController` | `POST .../auto-align` (writes a transform), `POST .../coherence` |
| `SceneNodesController` | `GET /projects/{id}/scene`, `GET /scene-nodes/{nodeId}/file` — download another project's geometry |
| `ModelDiffController` | `GET .../diff` — another project's element-level change history |

`ModelsController` and `IfcIngestController` already applied the documented gate; these five were missed.

## The fix

Mirrors `IfcIngestController` exactly — two layers, different questions, different codes:

- **`[ProjectAccess]`** (read gate → **404**) at controller level. 404 not 403, so a denial doesn't confirm the project exists.
- **`RequireProjectMemberAsync`** (write gate → **403**) as the first statement of every mutating action.

Two endpoints needed bespoke handling:
- `GetChunkFile` is routed by `{nodeId}`, so the attribute can't resolve a project from route data — the owning project is resolved off the node and checked explicitly, returning 404 to match.
- `Ingest` is `[AllowAnonymous]` (converter bearer) with no `{projectId}` — the attribute falls through by design.

**A2 — `FederatedModelHub.JoinProject`** had no check at all; the SignalR group name was the only key, so any authenticated user could subscribe to any project's `ModelUpdated` stream, across tenants. Now goes through `ProjectMembershipGuard`.

> It reads its tenant from the connection's JWT rather than the ambient filter. `ITenantContext` resolves off `IHttpContextAccessor`, which is **not reliably populated inside a hub method**; an unresolved tenant makes `CurrentTenantId` empty, which the filter treats as "matches no rows" — fail-closed, but it would have denied every legitimate member too. `NotificationHub` carries the same latent fragility; left alone as out of scope.

**A3 — `AutoAlignService.ComputeAsync`** persisted with a bare "overwrite if exists", ignoring `IsConfirmed`, silently destroying a coordinator's hand-set alignment. (The IFC-ingest path has always respected the flag.) It now refuses — and returns the transform it *would* have applied, since this path is an explicit user action, not an upload side effect.

## Verification

**26 new tests.** Every attacker is an active, authenticated member of the **same tenant** as the victim project, just a member of a *different* project. That's the distinguishing detail: a cross-tenant test passes against the unfixed code, which is exactly why the existing tenant-isolation suite never caught this.

Ran against the **unfixed** source (changes stashed): **13 failed, 8 passed** — every denial and gate test failed, while the happy-path tests passed. So the tests detect the defect rather than restate the fix, and the fixture isn't simply denying everything.

- With the fix: **26/26 pass**
- Full suite: **659 passed / 0 failed / 11 skipped**
- Build: **0 errors**, same 7 pre-existing warnings as baseline

Covered: the 403 in-action gate (controllers constructed directly against real SQLite with real `ProjectMember` rows), the 404 `[ProjectAccess]` gate (filter driven directly), and a reflection test pinning the attribute onto all five controllers so removing it fails a test.

**Not covered:** end-to-end MVC wiring — no host is booted (a `WebApplicationFactory` can't run against SQLite, and a second factory is unreliable here per ROADMAP DEP-7). Stated so a green run isn't read as proof of the pipeline.

Track A of a three-track federation-hardening effort; B (placement) and C (reliability) follow separately. Docs: `docs/COORDINATION_AUDIT_FINDINGS.md` PART IV.

---

## Follow-up — two identity controllers the first pass missed

A re-audit found the **same within-tenant hole on two more federation controllers** that also carried `[Authorize]` + tenant-only filtering. Both write the cross-host element **identity** table (the IFC-GlobalId ↔ host-element-id mapping that clash, issues and BCF resolve through), so a member of any one project could rewrite another project's identity in the same tenant.

| Surface | Exposure (before) |
|---|---|
| `IfcController` | `POST .../ifc/data`, `POST .../ifc/iot-binding` — ingest/overwrite another project's cross-host mappings + IoT bindings |
| `GlobalIdRegistryController` | `POST/PATCH/DELETE .../global-id-registry`, `POST .../auto-match` — create/edit/delete another project's canonical identity mappings |

Same fix, same two layers: `[ProjectAccess]` (read → 404) at controller level + `RequireProjectMemberAsync` (write → 403) as the first statement of every mutating action. `GetMappings` / `Resolve` / `List` / `Get` are reads covered by the class attribute.

**+9 tests** in `FederationAuthorizationTests`: a same-tenant non-member is 403'd on IFC ingest, IoT-binding, and registry create/update/delete/auto-match (spies prove the denial fires **before** the write); a real member still succeeds; and both controllers are added to the reflection theory that pins `[ProjectAccess]`. **27/27 in the class pass**, API builds **0 errors**.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/AlignmentController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/AlignmentController.cs
@@ -4,15 +4,29 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.EntityFrameworkCore;
+using Planscape.API.Authorization;
+using Planscape.API.Services;
 using Planscape.Core.Entities;
 using Planscape.Core.Interfaces;
 using Planscape.Infrastructure.Data;
 using Planscape.Infrastructure.Services;
 using Planscape.Infrastructure.SignalR;
 
+/// <remarks>
+/// <c>[ProjectAccess]</c> gates reads (404). The two POSTs both mutate or run
+/// expensive work against the project's alignment state, so each also takes the
+/// member gate (403) — the auto-align POST in particular writes a
+/// ProjectModelTransform.
+///
+/// The auto-align action overrides the class route with an absolute
+/// <c>~/api/projects/{projectId:guid}/...</c> template; that template still
+/// carries <c>{projectId}</c>, so the attribute resolves it from route data as
+/// normal.
+/// </remarks>
 [ApiController]
 [Route("api/projects/{projectId:guid}/alignment")]
 [Authorize]
+[ProjectAccess]
 public class AlignmentController : ControllerBase
 {
     private readonly PlanscapeDbContext _db;
@@ -57,6 +71,8 @@ public class AlignmentController : ControllerBase
         [FromServices] IFederatedCoherenceJob coherenceJob,
         CancellationToken ct)
     {
+        if (await this.RequireProjectMemberAsync(_db, projectId, ct) is { } denied) return denied;
+
         var report = await coherenceJob.RunAsync(projectId, _tenant.TenantId, ct);
         return Ok(report);
     }
@@ -72,6 +88,8 @@ public class AlignmentController : ControllerBase
         [FromServices] IHubContext<NotificationHub> notificationHub,
         CancellationToken ct)
     {
+        if (await this.RequireProjectMemberAsync(_db, projectId, ct) is { } denied) return denied;
+
         // #12 — pass both hubs so a successful auto-align broadcasts ModelUpdated:
         // FederatedModelHub for any /hubs/model client + NotificationHub
         // (project-{id}) which is where the dashboard + Revit plugin actually
--- a/Planscape.Server/src/Planscape.API/Controllers/CoordinateSystemController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/CoordinateSystemController.cs
@@ -3,6 +3,8 @@ namespace Planscape.API.Controllers;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
+using Planscape.API.Authorization;
+using Planscape.API.Services;
 using Planscape.Core.Entities;
 using Planscape.Core.Interfaces;
 using Planscape.Infrastructure.Data;
@@ -15,9 +17,17 @@ using Planscape.Infrastructure.Data;
 /// this definition.
 /// Route: /api/projects/{projectId}/coordinate-system
 /// </summary>
+/// <remarks>
+/// The CRS is the frame every model in the project is aligned against, so an
+/// unauthorised edit silently mis-places the whole federation. Gated like
+/// <see cref="IfcIngestController"/>: <c>[ProjectAccess]</c> for read (404) and
+/// <see cref="ControllerProjectMembershipExtensions.RequireProjectMemberAsync"/>
+/// on every mutation (403).
+/// </remarks>
 [ApiController]
 [Route("api/projects/{projectId:guid}/coordinate-system")]
 [Authorize]
+[ProjectAccess]
 public class CoordinateSystemController : ControllerBase
 {
     private readonly PlanscapeDbContext _db;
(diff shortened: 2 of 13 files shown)


StingD85:
## Federation hardening — consolidated review

Posting here since #676 is the root of the 12-PR stack (#676 → #678 → … → #688). This summarises a review of the whole chain and the fixes applied on top of it.

### Verdict

Strong work across security, placement, and reliability. Three blockers were found and fixed; the Track-C reliability items were reviewed and judged safe. The stack is now rebased onto latest `main`, internally coherent, and green end-to-end.

### Blockers found & resolved

**F1 — two identity controllers missed the project-membership gate _(fixed in this PR)_.**
The first pass gated five controllers + the SignalR hub, but `IfcController` and `GlobalIdRegistryController` — which write the cross-host element **identity** table (the IFC-GlobalId ↔ host-element-id mapping that clash, issues, and BCF all resolve through) — still carried `[Authorize]` + tenant-only filtering. So any member of any one project could read and overwrite **another** project's identity mappings within the same tenant. Cross-tenant was already closed by the global `ITenantScoped` filter; this was the within-tenant, cross-project gap. Added `[ProjectAccess]` (read → 404) + `RequireProjectMemberAsync` (write → 403) to both, mirroring `ModelsController` / `IfcIngestController`. **+9 tests** (non-member 403 on ingest / IoT-binding / registry create-update-delete-auto-match, member still succeeds, plus the reflection pin).

**B1 — viewer edits in the generated copy, not the source (#678, #680).**
The `applyModelTransform` auto-apply gate (#678) and the mesh-unit rescale (#680) were committed only to `Planscape.Server/src/Planscape.API/wwwroot/viewer.html`. The canonical source is `Planscape/assets/viewer/viewer.html`: the `SyncCoordinationViewer` MSBuild target copies source → wwwroot on every API build (so the edits would be reverted), and the `coordination-viewer-drift` CI gate fails on the divergence. Ported both edits into the source — all 7 synced files byte-equal again.

**B2 — rotated-frame translation off the frame axes (#682).**
#682 correctly made the stored rotation frame-relative (`θ_model − θ_frame`) but left the translation on raw CRS-grid axes. Under a **rotated** project frame the geometry (spun by `−θ_frame`) and its translation disagree, so two models with distinct survey origins stop overlaying by `(R(−θ_frame) − I)·(originB − originA)` — ~133 m at a 30° frame with blocks 500 m/400 m apart. Grid-aligned frames (`θ_frame = 0`) were unaffected, which is why it slipped through every existing test. Fixed by rotating the grid displacement by `−θ_frame` in `ModelGeorefWriter`; added `A_shared_point_coincides_even_when_the_frame_is_rotated` (verified **RED** before, **GREEN** after — the coincidence assertion also pins the sign).

### Track C — reviewed, no change needed

Delete-propagation (C3) and idempotency (C6) are wired but dormant; C3 cannot false-delete because the server acts only on explicit, host-attributed `RemovedGlobalIds`. Assessed as safe, not blockers.

### State

- Whole stack rebased onto latest `main`; all 12 base→head links properly stacked.
- Full `Planscape.Tests` at the stack tip: **768 passed / 0 failed / 16 skipped**.
- Viewer `source == wwwroot` ✓ across all 7 synced files.

One note for reviewers: during the restack, a fork-point slip briefly pointed #687/#688 at a commit already in their base, which GitHub auto-closed. Both were recovered (their feature commits were intact) and reopened with the correct heads (`4616511bb`, `9a3d6fab4`).

StingD85:
**Restack update**

- Added the F1 fix that makes this PR the complete cross-project authz change: `[ProjectAccess]` + `RequireProjectMemberAsync` on `IfcController` and `GlobalIdRegistryController` (the two identity controllers the first pass missed), with +9 tests.
- Rebased onto **latest `main`** (`3b33566ee`). Clean — every intervening `main` commit is in the Revit plugin / marketing site, disjoint from `Planscape.Server`.

Head is now `22485a27a`. `FederationAuthorizationTests`: 27/27.

PR #678, opened 2026-08-16
Placement P1: apply high-confidence transforms automatically, keep the manual override

StingD85:
> **Stacked on #676** (Track A). Base is `claude/federation-authz`; retarget to `main` once that merges.

## The bug

The alignment machinery was sound and **not wired to the screen**. Both automatic writers — the IFC ingest path and `AutoAlignService` — store `IsConfirmed = false`, and the viewer's gate was:

```js
const confirmed = (t.isConfirmed !== false);
if (!confirmed || isIdentity) return;
```

So a correct, survey-derived alignment was computed, persisted, sent to the viewer, and then **discarded by that one line**, every time, until a human confirmed it by hand. That's the whole "same-site models from different tools don't line up on their own" symptom, and it lived in four tokens of JavaScript.

## The fix

`IsAutoComputed` records *how* the numbers were obtained; it couldn't also record whether they're **live**. `ProjectModelTransform` gains:

| Column | Meaning |
|---|---|
| `AppliedAutomatically` | applied on its own, because the georeferencing graded HIGH |
| `Confidence` | `HIGH` / `LOW` / `NONE` |
| `Source` | `manual` / `ifc-map-conversion` / `auto-align` / `revit-georef` |

Viewer now applies when `isConfirmed === true` **OR** `appliedAutomatically === true`.

**Only HIGH-confidence transforms move a model.** `TransformConfidencePolicy` is the single shared rule: survey origin from IfcMapConversion + a CRS that either declares itself (IfcProjectedCRS) or matches the project's own + a non-FAIL report. Anything less is stored as a *suggestion*.

A model with **no** usable georeference stays at the origin — deliberately not "apply a best guess". An un-placed model at the origin is obviously un-placed and gets fixed in seconds; a model flung 500 km by a guess *looks placed* and costs an investigation.

The policy is shared and Revit-free because three pipelines produce automatic transforms (the Revit GLB path joins in P2). If each carried its own notion of "good enough", the same model would render in different places depending on which touched it last.

**Precedence is enforced at write time, not render time.** Automatic writers refuse to touch a confirmed row, so a confirmed row always holds the coordinator's numbers — nothing to reconcile when drawing. A manual PUT also clears `AppliedAutomatically` and stamps `Source="manual"`, making the rule total. (A manual transform saved with `IsConfirmed=false` stays a draft and doesn't render — unchanged.)

## Verification

- **19 policy unit tests** — exhaustive over a pure function. Includes the load-bearing case that `CrsEquivalent(null, null)` is **false**: if unknown matched unknown, every model without a declared CRS in a project without one would grade HIGH and be moved on *no evidence at all*.
- **3 write-path tests** on `AutoAlignService` — HIGH is marked auto-applied; unanchored and FAILed evidence are stored but not applied.
- **12 viewer-gate assertions** (`tests/web-harness/transform-gate.mjs`) — extracts the shipped `applyModelTransform` from `viewer.html` **on disk** and runs it in a `vm` sandbox against a fake three.js root (the `align-audit.mjs` precedent). Re-introducing the old gate was tried: **3 of 12 fail**, including the headline one.
- **Schema patch proven on real Postgres 16** — against a DB recreated in its *pre-change* shape, the three `ADD COLUMN IF NOT EXISTS` statements applied cleanly, were idempotent on a second pass, and left a pre-existing row at `AppliedAutomatically = false`. **No existing project starts rendering differently on deploy.**
- Full suite **691 passed / 0 failed / 11 skipped**; build **0 errors**.

Schema follows ADR 0001 (EnsureCreated + idempotent patcher). `dotnet ef migrations add` remains unsafe here — the model snapshot is stale and a new migration would try to re-create existing tables.

## Not in this PR

P2 (Revit GLB georef source), P3 (unit reconciliation — the dead `scaleFactor`), P4 (StingBridge georef), P5 (translation conventions + AABB freshness). This PR is the one that makes an *already-correct* transform actually render.

---

**Follow-up (review):** the `applyModelTransform` edit landed only in the generated `wwwroot/viewer.html`. The canonical source is `Planscape/assets/viewer/viewer.html` (the `SyncCoordinationViewer` MSBuild target copies source→wwwroot each build) and the `coordination-viewer-drift` CI gate fails on the divergence. Ported the identical change to the source so the two are byte-equal again (all 7 synced files verified).

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
@@ -4,6 +4,7 @@ using Microsoft.AspNetCore.SignalR;
 using Microsoft.EntityFrameworkCore;
 using Planscape.API.Authorization;
 using Planscape.API.Services;
+using Planscape.Core.Coordinates;
 using Planscape.Core.Entities;
 using Planscape.Core.Interfaces;
 using Planscape.Infrastructure.Data;
@@ -191,10 +192,7 @@ public class IfcIngestController : ControllerBase
                 if (ar.HasMapConversion && (ar.SurveyEasting.HasValue || ar.SurveyNorthing.HasValue))
                 {
                     await UpsertProjectModelTransformAsync(
-                        projectId, effectiveModelId, tenantId,
-                        ar.SurveyEasting ?? 0, ar.SurveyNorthing ?? 0, ar.SurveyElevation ?? 0,
-                        ar.MapConversionRotationDeg ?? 0, result.UnitScaleToMm,
-                        ct);
+                        projectId, effectiveModelId, tenantId, ar, result.UnitScaleToMm, ct);
                 }
             }
             catch (Exception aex)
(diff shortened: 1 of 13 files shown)


StingD85:
**Restack update**

- Ported this PR's `applyModelTransform` auto-apply edit from the generated `wwwroot/viewer.html` into the canonical source `Planscape/assets/viewer/viewer.html` (B1). The `SyncCoordinationViewer` build target copies source → wwwroot and the `coordination-viewer-drift` gate requires byte-equality, so the edit had to live in the source, not the generated copy.
- Rebased onto the updated `authz` base (now on latest `main`).

Head is now `7e173582a`. All 7 synced viewer files byte-equal.

PR #679, opened 2026-08-16
Placement P2: give the Revit GLB path a georef source, on one shared writer

StingD85:
> **Stacked on #678** (P1) → #676 (Track A). Retarget as those merge.

## The bug

Revit exports geometry about the **project internal origin** (`ClashExportContext` uses `Transform.Identity`), so a published GLB carried no survey position and **every Revit model landed at 0,0,0** regardless of where the building is.

The coordinate sidecar meant to carry that position had three faults:

1. Its easting/northing were **always null**, behind a comment blaming `BasePoint.GetCoordinateSystem()` being absent from the Revit 2025 API. That's true and beside the point — `ProjectLocation.GetProjectPosition(XYZ.Zero)` has always returned `EastWest` / `NorthSouth` / `Elevation` / `Angle`, the method was **already being called**, and the code read `Angle` and `Elevation` while dropping the two coordinates that place the building. (`ProjectSetupCommand.cs` uses `EastWest`/`NorthSouth` a few files away.)
2. `exportMode` was hardcoded with a `TODO`.
3. It was **never uploaded** — nothing read the file.

## The fix

The survey position travels as **metadata beside the geometry, never baked into the mesh**. Not a shortcut — a site at easting 432,000 m would put every vertex ~432 km from the origin, where 32-bit float loses millimetre precision and surfaces z-fight. It's also exactly what the IFC path does with `IfcMapConversion`.

- `RevitGeoref` (plugin) reads `ProjectPosition` **once**, used by both the sidecar and the upload, so they can't disagree.
- `UploadModelRequest` gains an optional georef block (E/N/elevation in metres, true north, CRS, length unit, export mode).
- `ModelsController.Upload` writes the transform via a shared writer.

**`ModelGeorefWriter` is now the ONE place** that turns georeferencing into a stored transform, and `IfcIngestController` was collapsed onto it. Before: one copy for IFC, none for Revit. Adding a *second* copy is how a building ends up in a different place depending on which pipeline last touched it — the most expensive class of bug here and the hardest to attribute. The writer was built around the IFC path's existing convention (`t = -origin`, m→mm) **deliberately**, so this changes no existing IFC behaviour.

Two "do nothing" cases matter as much as the placement:
- **No survey origin → no transform row at all.** Not an identity, and not a guess.
- **`SharedCoordinates` export → no transform.** That geometry is already in the survey frame; applying the origin again would double-count it. The plugin only emits `ProjectInternal` today, but the mode is reported as *fact* rather than assumed, so a future shared-coordinates export can't silently break.

## Honest limitation — the CRS

Revit has **no native CRS concept**. Without an anchor the transform grades LOW and is *stored* rather than applied, so a Revit model would still need one confirmation. The plugin reads an optional `PRJ_CRS_EPSG_TXT` project parameter — set it once per project and every future publish auto-places.

That's a **project-level declaration, not a per-model transform entry**, so the Track B definition of done still holds. But it *is* a setup step, and the plugin logs explicitly when it's missing.

## Verification

- **9 tests** on `ModelGeorefWriter`, including `The_ifc_source_label_produces_the_same_numbers_as_the_revit_one` — two hosts reporting the same survey origin must land in the same place. That assertion is the whole reason the writers were collapsed into one.
- **Revit plugin builds against the Revit 2025 API: 0 errors, 0 warnings.**
- Server build 0 errors; full suite **700 passed / 0 failed / 11 skipped**.

**NOT verified:** `RevitGeoref.Read` itself — it needs a live Revit `Document`, so no unit test reaches it. The numbers it produces are asserted only from the server side against hand-written inputs. Reading a real document's survey point remains a Revit-session check.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
@@ -8,6 +8,7 @@ using Planscape.Core.Coordinates;
 using Planscape.Core.Entities;
 using Planscape.Core.Interfaces;
 using Planscape.Infrastructure.Data;
+using Planscape.Infrastructure.Services;
 using Planscape.Infrastructure.SignalR;
 using System.Text.Json;
 using System.Text.Json.Serialization;
@@ -66,6 +67,7 @@ public class IfcIngestController : ControllerBase
     private readonly Planscape.Core.Interfaces.IIfcAlignmentValidator _alignmentValidator;
     private readonly IHubContext<FederatedModelHub> _modelHub;
     private readonly IHubContext<NotificationHub> _notificationHub;
+    private readonly Planscape.Infrastructure.Services.IModelGeorefWriter _georefWriter;
     private readonly ILogger<IfcIngestController> _logger;
 
     public IfcIngestController(
@@ -76,6 +78,7 @@ public class IfcIngestController : ControllerBase
         Planscape.Core.Interfaces.IIfcAlignmentValidator alignmentValidator,
         IHubContext<FederatedModelHub> modelHub,
         IHubContext<NotificationHub> notificationHub,
+        Planscape.Infrastructure.Services.IModelGeorefWriter georefWriter,
         ILogger<IfcIngestController> logger)
     {
         _db = db;
@@ -85,6 +88,7 @@ public class IfcIngestController : ControllerBase
         _alignmentValidator = alignmentValidator;
         _modelHub = modelHub;
         _notificationHub = notificationHub;
+        _georefWriter = georefWriter;
         _logger = logger;
     }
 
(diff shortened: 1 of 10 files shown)


StingD85:
**Restack update**

Rebased onto the updated `autoapply` base; the stack is now rooted on latest `main`. No change to this PR's own commit — carried forward unchanged (new SHA from the rebase only).

Head is now `e6fc393d4`.

StingD85:
## Pre-review finding — one-line doc fix: stale coordinate comment

`StingTools/BIMManager/RevitGeoref.cs` (~line 108) documents the survey origin as:

> "…precisely the number the server needs to **negate**."

But the current `ModelGeorefWriter` deliberately does **not** negate — it uses `t = +origin − frameOrigin`. The negation was exactly the mirrored-translation bug this stack fixed (Placement P5). The comment now describes the superseded sign convention, on load-bearing coordinate code where a maintainer who trusts it could re-introduce a negation and mirror every Revit model about the frame origin.

**Suggested fix (comment only, no runtime change):** describe the shipped convention — the survey coordinates of the project internal origin, which the server places with `t = +origin − frameOrigin` (see `ModelGeorefWriter` remarks) — rather than "the number the server needs to negate."

I left this as a flag rather than fix it in place, because the file sits at the bottom of the stack and editing it would force a rebase of every branch above for a one-line comment. Happy to do that cascade if you'd prefer it fixed directly.

StingD85:
## Resolved — F5 fixed in place

The stale comment flagged above is fixed in `a8e0e84a0`. `RevitGeoref.Read` now describes the shipped convention — the server places the model with `t = +origin − frameOrigin` and does **not** negate (see `ModelGeorefWriter` remarks) — and warns against reintroducing a sign flip. Comment-only change; the fix was cascaded up the stack, and the full server suite is green (769 passed / 0 failed / 15 skipped) at the tip.

PR #680, opened 2026-08-16
Placement P3: one mesh-unit convention — the metre — across both GLB writers

StingD85:
> **Stacked on #679** (P2) → #678 (P1) → #676 (Track A). Retarget as those merge.

## The bug

**Two GLB writers in this repo disagreed by a factor of 1000.**

| Writer | Conversion | Result |
|---|---|---|
| `GlbSerializer` | feet × `0.3048` | metres |
| `RevitGltfExporter` | feet × `304.8` | **millimetres** |

Both upload to the same endpoint and are rendered by the same viewer — which assumes metres. glTF 2.0: *"The units for all linear distances are meters."* `applyModelTransform` divides the stored millimetre translation by 1000 precisely because the geometry it positions is metres.

It hid because **a model viewed alone looks correct at any uniform scale** — the camera fits to whatever bounds it finds, so a building rendered 1000× too large looks perfectly normal. The mismatch only surfaces once a Revit model is federated with one from another tool, and then it reads as *"the models don't line up"* — a coordinate problem, sending you looking in entirely the wrong place.

Alongside it: `ProjectModel.Units` was written on every upload and **read by nothing**, and the ingest path computed a `scaleFactor` whose two branches both returned `1.0`, silently discarding any declared `IfcMapConversion` scale.

## The fix

**One convention: the metre.** `RevitGltfExporter` writes metre vertices, and the published bounds moved with it — same geometry, so leaving bounds in mm would swap a *visible* render bug for an *invisible* bounds one.

`ProjectModel.Units` now means something: it rides on the transform payload the viewer already fetches (`meshUnitScale`), applied **unconditionally** — a mm model needs rescaling whether or not it's georeferenced, so it must not sit behind the "is this transform live" gate.

**Mesh unit and georef scale stay separate and multiply.** `ScaleFactor` is a *survey correction* declared by an `IfcMapConversion`; the mesh unit is how the vertex data happens to be written. Conflating them makes a unit fix look like a survey error. *(Deliberate deviation from the runner's "compute scaleFactor from model unit" — the rendered result is identical because both multiply.)* The dead `scaleFactor` is now real: `ModelGeoref` carries `MapConversionScale`, the writer inverts it, matching `AutoAlignService`.

## Two traps found on the way

1. **The IFC→GLB job copied the SOURCE IFC's unit onto the GLB derivative.** Harmless while nothing read the field — the moment `Units` drives scaling it would have shrunk every converted model by 1000. The converter emits glTF, so the derivative is metres by construction. Fixed, plus a **targeted idempotent backfill** scoped by `UploadedBy` + `Format = 0`, so the source IFC row and every hand-published model are untouched.
2. **The server defaulted an undeclared unit to `"mm"`.** Any uploader omitting the field would be scaled by 1/1000. Now the canonical metre — and an unknown unit also reads as metres, because the fail-safe direction for "I don't recognise this" is **change nothing**; a wrong guess silently rescales a whole building.

## Verification

- **16 unit tests** on `MeshUnits`, including that unknown/blank reads as metres rather than as a guess.
- **10 harness assertions** against the **shipped sources** — the viewer scales a mm mesh with *no* transform; mesh unit and georef scale multiply rather than replace; an absent `meshUnitScale` changes nothing; plus a grep-level guard that the two Revit writers still agree. Crude, but it's the only thing that fails when someone edits one and not the other.
- **Backfill SQL proven on real Postgres 16** against a 4-row fixture: mislabelled converted GLB corrected; already-correct one untouched; **source IFC row and a genuinely-mm hand-published Revit model both left alone**. Idempotent on a second pass.
- Server 0 errors; **Revit plugin builds against the Revit 2025 API: 0 errors, 0 warnings**; full suite **719 passed / 0 failed / 12 skipped** (all skips pre-existing environment-gated).

## Behaviour change, stated plainly

Revit models published *before* this change are stored with mm meshes and `Units = "mm"`. They now render at the **correct** size because the viewer honours that unit. A row predating the `Units` column entirely would read as metres. Re-publishing puts any model on the canonical footing.

---

**Follow-up (review):** same as one level down — this PR's mesh-unit edit was only in `wwwroot/viewer.html`. Ported it into the canonical `Planscape/assets/viewer/viewer.html` so the build target does not revert it and the source↔wwwroot byte-equal drift gate passes.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
@@ -638,7 +638,11 @@ public class IfcIngestController : ControllerBase
                 CrsCode       : report.CrsName,
                 HasDeclaredCrs: report.HasProjectedCrs,
                 LengthUnit    : report.LengthUnit,
-                SourceLabel   : "ifc-map-conversion");
+                SourceLabel   : "ifc-map-conversion",
+                // Previously dropped: the ingest computed a scaleFactor whose
+                // two branches both returned 1.0, so a declared map-conversion
+                // scale never reached the transform.
+                MapConversionScale: report.MapConversionScale);
 
             var confidence = await _georefWriter.WriteAsync(
                 projectId, projectModelId, tenantId, georef, report.Verdict, ct);
--- a/Planscape.Server/src/Planscape.API/Controllers/ModelTransformController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ModelTransformController.cs
@@ -59,12 +59,28 @@ public class ModelTransformController : ControllerBase
                   && t.TenantId       == _tenant.TenantId,
                 ct);
 
+        // P3 — the mesh's own length unit travels with the transform because
+        // this is the one payload the viewer already fetches per model, and the
+        // two are multiplied at render time anyway. It is reported even when
+        // there is NO transform: a model whose mesh is in millimetres needs
+        // rescaling whether or not it is georeferenced, and before this the
+        // viewer never read ProjectModel.Units at all, so such a model rendered
+        // 1000x too large. Alone that is invisible (the camera fits to bounds);
+        // federated with a metre model it is a thousand-fold mismatch.
+        var meshUnits = await _db.ProjectModels.AsNoTracking()
+            .Where(m => m.Id == modelId && m.ProjectId == projectId && m.TenantId == _tenant.TenantId)
+            .Select(m => m.Units)
+            .FirstOrDefaultAsync(ct);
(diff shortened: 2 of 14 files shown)


StingD85:
**Restack update**

- Ported this PR's mesh-unit rescale edit from `wwwroot/viewer.html` into the canonical source `Planscape/assets/viewer/viewer.html` (B1), same reason as #678.
- During the rebase a `viewer.html` conflict (this source edit vs. the auto-apply block inherited from #678) was resolved to the invariant `source := wwwroot`, so the source now carries both the auto-apply and mesh-unit blocks and matches the shipped copy byte-for-byte.
- Rebased onto the updated `revit-georef` base (now on latest `main`).

Head is now `8812704a2`.

PR #681, opened 2026-08-16
Placement P4: read a full georeference from IFC and send it on upload

StingD85:
> **Stacked on #680** (P3) → #679 (P2) → #678 (P1) → #676 (Track A). Retarget as those merge.

## The bug

`GeorefDescriptor` existed with **zero consumers**. Nothing built a complete one and nothing sent one anywhere — so an ArchiCAD or Tekla IFC pushed through the bridge arrived with no survey position, landed at the project origin, and had to be placed by hand. That's the manual step this whole track exists to remove.

`IfcFileHostAdapter.georef_descriptor` read only `IfcMapConversion`'s eastings/northings/height/scale. Three fields were unset or wrong:

| Field | Before | Why it mattered |
|---|---|---|
| `crs_epsg` | never read | The server grades an *unanchored* survey origin LOW, stores the transform as a suggestion, and leaves the model at the origin until someone confirms it. Reading `IfcProjectedCRS.Name` is what makes a model auto-place. |
| `true_north_deg` | never read | A model's rotation was silently dropped. A building at the right E/N but rotated 20° off is arguably *worse* than one at the origin — because it looks placed. |
| `length_unit` | defaulted `"mm"` | Eastings/northings are **metres** by IFC definition, so any consumer trusting the field was off by 1000. |

## The fix

- **True north** from `atan2(XAxisOrdinate, XAxisAbscissa)` — the **same formula the server uses** in `IfcAlignmentValidator`. It has to be: the same file described here and ingested there must not disagree about which way the building faces.
- **CRS** from `IfcProjectedCRS.Name`, which also raises the LoGeoRef tier (50 with a named CRS, 30 without — coordinates whose system is unstated are a weaker claim).
- **Length unit** read from the model via `calculate_unit_scale`, falling back to metres.
- Descriptor default `length_unit` `"mm"` → `"m"`, so an un-populated descriptor means *change nothing* rather than *rescale by 1000*. One existing test pinned the old default and was updated **with the reason**.
- `PlanscapeClient.upload_model` accepts a descriptor and flattens it onto the multipart upload as the same `Georef*` fields the Revit path uses.

## Verification

- **14 tests** on the extraction, driven through a stand-in model so no `ifcopenshell` install is needed — the defects were all in extraction and arithmetic, which is exactly what these pin.
- **8 tests** on the wire contract. These matter because a field-name typo is **not an error at either end**: ASP.NET model binding leaves the property null, the server writes no transform, and the model quietly stays at the origin. *Nothing fails; the building is just in the wrong place.*
  Also pinned: a model with no survey origin sends **no** georef block at all, and a zero true north is *sent* rather than dropped by a falsy check — "no rotation" ≠ "rotation unknown".
- Full Python suites green: **118 core**, **179 StingBridge**.

**NOT verified:** extraction against a real `.ifc`. `ifcopenshell` isn't a core dependency (that's the point of core), so the unit-scale branch and `by_type` against a real file remain an environment-gated check.

### DIFF
--- a/StingBridge/planscape/client.py
+++ b/StingBridge/planscape/client.py
@@ -395,12 +395,28 @@ class PlanscapeClient:
             log.debug("get_element_timestamps failed (non-fatal): %s", e)
             return {}
 
-    def upload_model(self, project_id: str, glb_path) -> dict:
-        """Upload a GLB file to Planscape /api/projects/{id}/models."""
+    def upload_model(self, project_id: str, glb_path, georef=None) -> dict:
+        """Upload a GLB file to Planscape /api/projects/{id}/models.
+
+        ``georef`` is an optional
+        :class:`stingtools_core.hosts.adapter.GeorefDescriptor`. When supplied
+        with a survey origin, the server turns it into a coordinate transform
+        and — if the coordinates are anchored to a named CRS — applies it
+        automatically, so the model lands in the federation without anyone
+        typing a transform.
+
+        It is passed as METADATA rather than baked into the geometry, matching
+        the Revit path: a site at easting 432,000 m would put every vertex
+        ~432 km from the origin, where 32-bit float mesh coordinates lose
+        millimetre precision.
+        """
         import mimetypes
         from pathlib import Path
         path = Path(glb_path)
         mime = mimetypes.guess_type(str(path))[0] or "model/gltf-binary"
+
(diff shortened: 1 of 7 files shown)


StingD85:
**Restack update**

Rebased onto the updated `units` base; stack now rooted on latest `main`. No change to this PR's own commit.

Head is now `054f2148b`.

PR #682, opened 2026-08-16
Placement P5: correct the mirrored translation, and keep chunk bounds fresh

StingD85:
> **Stacked on #681** (P4) → #680 (P3) → #679 (P2) → #678 (P1) → #676 (Track A). Retarget as those merge.

## The bug: the translation was mirrored

Both automatic writers computed the **negation** of the correct translation. IFC ingest used `t = -modelOrigin`; `AutoAlignService` used `t = referenceOrigin - modelOrigin`.

A model's geometry is authored about its own internal origin, and its georeferencing states where that origin sits in the survey CRS. So a physical point `S` sits at local coordinate `S - A` in model A, and `world = t + local`:

```
t = +A  →  world = A + (S - A) = S         ✓ both models agree on S
t = -A  →  world = -A + (S - A) = S - 2A   ✗ and B lands at S - 2B
```

Two models came out **mirrored about the origin** — an east-west pair swaps sides. Correct: `t = modelSurveyOrigin - projectFrameOrigin`.

### Why the existing "overlay proof" missed it

`ModelTransformMathTests` inverts a transform and re-applies **the same** transform, proving `Apply(Inverse(w)) == w` — true for *any* transform. It proves the math is self-consistent, never that the transform was **derived** correctly from survey data. The new tests derive it, and assert the **sign** explicitly, because a magnitude-only check passes against the bug.

## One frame, resolved once

`ModelGeorefWriter` resolves the project frame in one place — declared `ProjectCoordinateSystem` benchmark → nominated reference model → zero. The **same** frame is subtracted from every model, so relative placement is identical whichever is chosen; a site-local frame only buys precision.

`AutoAlignService` no longer does its own arithmetic — it delegates to the writer, so the two cannot drift. Its old frame ("the most recently validated sibling") is **dropped** and now reporting-only: that origin moved every time another model was uploaded, so transforms computed on different days silently disagreed.

The writer returns what it computed **even when it refuses to write**. A refusal that can't say what the survey data implies is useless at exactly the moment the answer matters — and having the caller recompute it is how the two implementations drifted apart in the first place.

## The bug: chunk bounds went stale — and couldn't simply be refreshed

The AABB recompute lived inline in the manual PUT, with two problems that compounded:

1. **It ran only there.** Both automatic writers left the manifest describing where chunks *used to be*. The viewer culls against those bounds → geometry vanishes when the camera looks straight at it.
2. **It transformed an already-transformed box.** So the obvious fix — "call it after every write" — would have made it **worse**, compounding on each call.

Fixing (2) had to come first. `SceneNode` gains a nullable **local** AABB, so the world box is a pure function of (local, transform) and is recomputable any number of times. The refresher is now called from the manual PUT, the manual DELETE (resetting to identity moves the model too) and the shared writer — covering both automatic paths at once.

## Verification

- **7 frame tests** — relative offset preserved; **not mirrored** (asserted by sign); a point shared by two models maps to one world coordinate; benchmark and nominated reference each become the frame; rotation is frame-relative.
- **7 refresher tests** — refreshing twice equals once (*the* regression); local box never mutated; removing the transform restores the local position; a rotated box encloses **all eight** corners (a two-corner shortcut yields an inverted, too-small box that culls visible geometry); scale applies before translation.
- Existing sign expectations in two files were updated **with the reason recorded in each test** — they pinned the mirrored behaviour.
- Schema patch verified idempotent on **real Postgres 16**.
- Full suite **734 passed / 0 failed / 11 skipped**; both web harnesses green; build 0 errors.

## Behaviour change, stated plainly

**Every automatically-placed model moves.** That is the point — they were mirrored. Coordinator-confirmed transforms are untouched; a re-publish or an auto-align run re-derives the rest.

---

**Follow-up (review):** this PR made the stored rotation frame-relative but left the translation on raw CRS-grid axes, so under a rotated project frame two models with different survey origins stop overlaying by `(R(−θ_frame) − I)·(originB − originA)` (~133 m at 30°). Added the missing `R(−θ_frame)` to the translation + a rotated-frame shared-point test (verified RED before, GREEN after; grid-aligned frames unchanged). Full suite 735/0/11.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/IfcIngestController.cs
@@ -644,7 +644,7 @@ public class IfcIngestController : ControllerBase
                 // scale never reached the transform.
                 MapConversionScale: report.MapConversionScale);
 
-            var confidence = await _georefWriter.WriteAsync(
+            var write = await _georefWriter.WriteAsync(
                 projectId, projectModelId, tenantId, georef, report.Verdict, ct);
 
             // Gap 14: Write an audit log entry for the coordinate correction.
@@ -660,8 +660,12 @@ public class IfcIngestController : ControllerBase
                         autoComputed = true,
                         // B1 — a model that MOVES on its own must say so in the
                         // audit trail, and say what it was trusting when it did.
-                        confidence = TransformConfidencePolicy.ToStorageString(confidence),
-                        appliedAutomatically = TransformConfidencePolicy.ShouldAutoApply(confidence),
+                        confidence = TransformConfidencePolicy.ToStorageString(write.Confidence),
+                        appliedAutomatically = write.Written
+                                               && TransformConfidencePolicy.ShouldAutoApply(write.Confidence),
+                        translationXMm = write.TranslationXMm,
+                        translationYMm = write.TranslationYMm,
+                        frameSource = write.FrameSource,
                     }));
             }
             catch { /* audit failure is non-fatal */ }
--- a/Planscape.Server/src/Planscape.API/Controllers/ModelTransformController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ModelTransformController.cs
@@ -33,17 +33,20 @@ public class ModelTransformController : ControllerBase
     private readonly PlanscapeDbContext _db;
     private readonly ITenantContext _tenant;
     private readonly IIfcDeltaService _delta;
+    private readonly ISceneNodeAabbRefresher _aabb;
     private readonly ILogger<ModelTransformController> _logger;
 
     public ModelTransformController(
         PlanscapeDbContext db,
         ITenantContext tenant,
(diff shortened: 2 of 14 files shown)


StingD85:
**Restack update**

- Added the B2 fix: the stored rotation was frame-relative (θ_model − θ_frame) but the translation was left on raw CRS-grid axes, so a **rotated** project frame threw models with distinct survey origins out of registration by `(R(−θ_frame) − I)·(originB − originA)` — ~133 m at a 30° frame. Now rotates the grid displacement by −θ_frame in `ModelGeorefWriter`; grid-aligned frames (θ_frame = 0) are unchanged.
- Added `A_shared_point_coincides_even_when_the_frame_is_rotated` (verified RED before, GREEN after; the coincidence assertion also pins the sign).
- Rebased onto the updated `bridge-georef` base (now on latest `main`).

Head is now `69fab0ace`. Full `Planscape.Tests` green.

PR #683, opened 2026-08-16
Placement P6 + payoff: prove the placement chain end to end on PostgreSQL

StingD85:
> **Stacked on #682** (P5) → #681 → #680 → #679 → #678 → #676. Retarget as those merge. Completes Track B.

## The payoff

A coordinator uploads two models of the **same site** from **different tools**, with different survey origins and a shared CRS, types **no** transform — and both are placed automatically and overlay.

`PostgresFederationPlacementTests` — four `[SkippableFact]`s on **real PostgreSQL**, each in a rolled-back transaction:

1. **Two models of one site are placed automatically and overlay.** Both grade HIGH and are marked auto-applied; neither is marked *confirmed* (that word is reserved for a human); their true relative offset survives (60 m east, 45 m north); a column both models contain maps to **one** world coordinate; and the scene-chunk world AABBs land where the transforms put them — including that the two chunks do **not** overlap in X, because *a test that only asserted "the boxes moved" would pass with both models stacked on top of each other.*
2. **Relative placement doesn't depend on whether a project frame is declared.** The frame shifts the whole federation; if it changed relative placement, declaring a benchmark mid-project would silently move buildings apart.
3. **A model without georeferencing gets no transform row at all** — not a guessed one.
4. **A coordinator's confirmed transform survives an automatic pass**, and the caller is still told what the survey data implies.

## P6 — Tekla

Tekla needs no connector and has none. Its route is a **Tekla-authored IFC through the generic ingest**, which already recognises it:

```csharp
string source = hasAcPsets ? "archicad" : hasTeklaPsets ? "tekla" : "ifc";
```

…and `tekla` is a first-class host constant (`MappingHosts.Tekla`) used by the GlobalId registry and cross-host DTOs. Placement is **host-agnostic by construction** after P1–P5: everything downstream of `IfcMapConversion` runs through `ModelGeorefWriter`, which doesn't know which tool wrote the file. A native Tekla plugin remains **out of scope**.

## What this does NOT cover

**IFC parsing.** There are no `.ifc` fixtures in this repo and xbim is a heavy dependency to stand up for a placement test, so the georeferencing is supplied as the values an `IfcMapConversion` + `IfcProjectedCRS` would yield. The placement chain is under test; the parser is not. Stated so a green run isn't read as proof that a real Tekla export parses.

## Fixture trap worth recording

The first run failed with *"expected 2 transforms, found 0"*. The context had been built from the options-only constructor, so `CurrentTenantId` was `Guid.Empty` and the global filter matched nothing — rows were written that the very next read couldn't see. Worse, the writer's own "is there an existing transform?" lookup was equally blind, so the **not-overwritten** assertion would have passed for entirely the wrong reason. The tenant id is now minted before the context and supplied to it. (Same trap `MaterialSyncAuthorizationTests` documents — easy to fall into because the failure mode is *silence*.)

## Verification

With `PLANSCAPE_TEST_PG` pointed at a throwaway PostgreSQL 16:

> **749 passed / 0 failed / 0 skipped** — nothing gated out, including every previously environment-skipped Postgres test.

Without it: 734 passed / 0 failed / 11 skipped.

### DIFF
--- a/Planscape.Server/tests/Planscape.Tests/PostgresFederationPlacementTests.cs
+++ b/Planscape.Server/tests/Planscape.Tests/PostgresFederationPlacementTests.cs
@@ -1,0 +1,335 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging.Abstractions;
+using Planscape.Core.Coordinates;
+using Planscape.Core.Entities;
+using Planscape.Core.Interfaces;
+using Planscape.Infrastructure.Data;
+using Planscape.Infrastructure.Services;
+using Xunit;
+
+namespace Planscape.Tests;
+
+/// <summary>
+/// TRACK B PAYOFF — the whole point of the placement work, on the production
+/// database engine.
+///
+/// THE SCENARIO
+/// ------------
+/// A coordinator uploads two models of the SAME site from DIFFERENT tools, with
+/// different survey origins and a shared CRS, and does not type a transform.
+/// Both must be placed automatically, and the parts of the building they share
+/// must end up in the same world position.
+///
+/// WHAT THIS EXERCISES, AND WHAT IT DOES NOT
+/// -----------------------------------------
+/// It drives the server-side placement chain end to end — frame resolution,
+/// translation, confidence grading, the auto-apply decision, the transform row,
+/// and the scene-chunk world AABBs — on real PostgreSQL, so the DDL, the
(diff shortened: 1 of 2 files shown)


StingD85:
**Restack update**

Rebased onto the updated `frame` base (which now carries the B2 translation fix); stack rooted on latest `main`. No change to this PR's own commit.

Head is now `abee09343`.

Issue #596, opened 2026-08-05
StingTools.Clash: the legacy ClashDetectionCommands.cs is back, with duplicate LiveClashUpdater + ClashSession


StingD85:
Surfaced by restoring the 97 dead tests in #553. `StingTools.Clash.Tests` compiles again, and the
first thing it reports is a regression guard that has been unable to fire since May.

## The failing test

`StingTools.Clash.Tests/ClashWireupRegressionTests.cs:98`

```
DuplicateLiveClashUpdater_FileIsRemoved [FAIL]
  Legacy ClashDetectionCommands.cs must remain deleted; it shipped duplicate stubs
  that hijacked LiveClashUpdater.Register.
```

## Measured

The file exists — `StingTools/Clash/ClashDetectionCommands.cs`, 65,711 bytes.

`git log --diff-filter=A` puts its return at **`2c20e6724` — "Phase 106: recreate
ClashDetectionCommands.cs after Phase 105 merge loss"**, and it has been touched four times
since. So it came back as a deliberate merge-loss recovery, not by accident — but the guard
that would have flagged the duplication was already offline.

There really are two of each type, in different namespaces:

| Type | Legacy | Real |
|---|---|---|
| `LiveClashUpdater` | `StingTools.Clash` — `ClashDetectionCommands.cs:172` | `StingTools.Core.Clash` — `LiveClashUpdater.cs:12` |
| `ClashSession` | `StingTools.Clash` — `ClashDetectionCommands.cs:55` | `StingTools.Core.Clash` — `ClashSession.cs:17` |

The plugin compiles 0/0 precisely *because* they sit in different namespaces — which is the
hazard the test's own comment describes, not an absence of one. Which type a call site binds
to is decided by whichever `using` happens to be in scope.

## What is NOT established

**I have not determined which `LiveClashUpdater` is actually registered at Revit startup,
or whether live clash detection is presently misbehaving.** The duplication is measured; the
runtime consequence is not. That check needs Revit and is the first thing to do here.

Deliberately not fixed as part of #553 — deleting or renaming a 65k file is a real change
that deserves its own reasoning, and the restore PR's job was to make the suite audible
again, not to act on the first thing it said.

## Baseline

Excluded by name in the `stingtools-unit-tests` CI job with this issue number inline, so it
does not mask any *new* Clash failure. Remove the exclusion when this is resolved.

Related: #553.

PR #684, opened 2026-08-16
Reliability C1+C2: stop losing geometry deltas, and sync every model category

StingD85:
> **Stacked on #683** (Track B) → … → #676. Retarget as those merge. First of Track C.

Two defects in the Revit geometry-delta pipeline. Both fail **silently** and present as *"the server's model just isn't the same as mine."*

## C1 — a failed delta was permanently lost

`GeometrySyncHandler` drained the change queue and then fired the upload as a **discard-result** `Task.Run`.

Draining is destructive: the ids leave the queue and the only other copy is a local list. `PostGeometryDeltaAsync` returns `false` immediately when `CurrentProjectId` is unset, and also on a non-2xx or an idle-expired token — and **the result was thrown away**, so all three looked exactly like success. The elements were never considered again; the Revit model and the server model diverged permanently, and nothing said so. The next edit to a *different* element would sync fine, which makes the gap look like it never happened.

Now: the project link is checked **before** draining; the upload's result is read; on failure the ids go back for the next save.

Re-queued rather than retried in a loop **on purpose** — the trigger is a document save, so the natural retry is the next one; a tight loop would hammer an unreachable server from a background thread with no backoff and no way to stop it.

**Only what was actually sendable is retried.** An element whose geometry can't be extracted (deleted since the edit, no solid, view-specific) would fail extraction again every save — converting one lost delta into an *infinite* retry that drags the genuinely retryable ids along forever. **Deletions are always retried:** a tombstone needs no geometry, and a lost one is the worst case — the element stays visible in the federated model forever.

## C2 — only nine categories ever synced

`LiveClashUpdater` was the sole producer for `GeometrySyncQueue`, and its trigger covers the nine categories **clash** cares about. Doors, windows, mechanical equipment, plumbing and lighting fixtures, furniture, specialty equipment, generic models, stairs, railings, roofs, curtain panels → **never reached the server automatically.** A coordinator could move every door in the building and the federated model would not change.

Worse: the clash trigger is **opt-out**, so a project with `LIVE_CLASH_TRIGGERS_ENABLED=false` silently turned off *all* geometry sync while tag sync kept flowing — the model looked connected and was not.

Geometry sync now has its **own** updater over every model element instance — expressed as **filters** (not an `ElementType`, not owned by a view) rather than a category list, *because a list is exactly how the clash trigger ended up missing doors*. It registers unconditionally, so turning off clash detection can't stop the federated model updating. `LiveClashUpdater` no longer double-produces.

## Testing — and its honest limit

`GeometrySyncHandler` is an `IExternalEventHandler` needing a `UIApplication` and a live `Document`, so it's unreachable from the pure-logic test projects. The two decisions that actually govern whether a change reaches the server — the queue's **sign encoding** and the **retry rule** — were extracted into `GeometrySyncPlan` and pinned by **8 tests**, including:

- the retry set **round-trips** through the encoding (otherwise a retry silently turns a deletion into an edit);
- a negative "changed" id can **never** reach the retry set (it would *erase* the element on the next successful sync).

**NOT covered:** the handler's ordering (check the link before draining), the re-queue call itself, and the new updater's trigger scope — all need Revit. A green run is **not** proof that a delta survives a failed upload in a real session.

## Build / test

- Plugin against the Revit 2025 API: **0 errors, 0 warnings**
- `StingTools.Clash.Tests`: **71 passed, 1 failed** — `DuplicateLiveClashUpdater_FileIsRemoved`, which is **pre-existing on `origin/main`** (CLAUDE.md tracks it as #596) and untouched here. Verified with `git cat-file -e origin/main:…`.

### DIFF
--- a/StingTools.Clash.Tests/GeometrySyncPlanTests.cs
+++ b/StingTools.Clash.Tests/GeometrySyncPlanTests.cs
@@ -1,0 +1,130 @@
+using System.Linq;
+using StingTools.Core.Clash;
+using Xunit;
+
+namespace StingTools.Clash.Tests;
+
+/// <summary>
+/// TRACK C1/C2 — the queue encoding and the retry rule for Planscape geometry
+/// sync.
+///
+/// WHY THESE TWO THINGS AND NOT THE HANDLER
+/// ----------------------------------------
+/// GeometrySyncHandler is an IExternalEventHandler that needs a UIApplication
+/// and a live Document, so nothing in it is reachable from a pure-logic test
+/// project (this one deliberately does not reference the Revit API). The two
+/// decisions that actually govern whether a change reaches the server are plain
+/// arithmetic over ints, so they were extracted into GeometrySyncPlan and are
+/// pinned here.
+///
+/// NOT COVERED: the handler's ordering (check the project link BEFORE draining),
+/// the re-queue call itself, and the new GeometrySyncUpdater's trigger scope.
+/// Those need Revit. Stated so a green run is not read as proof that a delta
+/// survives a failed upload in a real session.
+///
+/// THE DEFECT THIS GUARDS
+/// ----------------------
+/// The queue packs a deletion as a NEGATED element id. The sign is therefore the
+/// semantics: read it backwards and every deletion becomes an attempt to
+/// tessellate an element that no longer exists, while every edit becomes a
+/// tombstone that erases live geometry from the server.
+/// </summary>
+public class GeometrySyncPlanTests
+{
+    // ── the sign convention ─────────────────────────────────────────────────
+
+    [Fact]
+    public void Positive_ids_are_changes_and_negative_ids_are_deletions()
+    {
+        var (changed, deleted) = GeometrySyncPlan.Partition(new[] { 10, -20, 30, -40 });
+
+        Assert.Equal(new[] { 10, 30 }, changed);
+        // Deletions come back POSITIVE — the sign was encoding, not data.
+        Assert.Equal(new[] { 20, 40 }, deleted);
+    }
+
+    [Fact]
+    public void A_zero_id_is_dropped_rather_than_guessed()
+    {
+        // 0 is not a valid Revit element id. Treating it as "changed" would try
+        // to tessellate nothing; treating it as "deleted" would tombstone
+        // element 0. Neither is right, so it goes nowhere.
+        var (changed, deleted) = GeometrySyncPlan.Partition(new[] { 0, 7 });
+
+        Assert.Equal(new[] { 7 }, changed);
+        Assert.Empty(deleted);
+    }
+
+    [Fact]
+    public void An_empty_or_null_drain_is_empty()
+    {
+        var (c1, d1) = GeometrySyncPlan.Partition(new int[0]);
+        Assert.Empty(c1);
+        Assert.Empty(d1);
+
+        var (c2, d2) = GeometrySyncPlan.Partition(null);
+        Assert.Empty(c2);
+        Assert.Empty(d2);
+    }
+
+    // ── the retry rule ──────────────────────────────────────────────────────
+
+    [Fact]
+    public void The_retry_set_round_trips_through_the_queue_encoding()
+    {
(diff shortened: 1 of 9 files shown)


StingD85:
**Restack update**

Rebased onto the updated `payoff` base; stack now rooted on latest `main`. No change to this PR's own commits.

Head is now `617652f7c`.

StingD85:
## Pre-review finding — deferred (needs a dedicated change): 64-bit `ElementId` truncation in geometry sync

Found during a pre-review pass. Flagging rather than fixing here because the correct fix is cross-cutting.

**The bug.** `GeometrySyncUpdater.Execute` enqueues `(docGuid, (int)id.Value)` for adds/modifies and `-(int)id.Value` for deletes; `GeometrySyncPlan.Partition` reads the **sign** as the semantics (positive = add/modify, negative = delete). Two failure modes:

1. **Truncation** — Revit 2024+ `ElementId.Value` is 64-bit and can exceed `Int32.MaxValue` in large / long-lived models; `(int)id.Value` wraps.
2. **Sign-as-semantics** — a wrapped-negative id for a *modified* element is read as a **tombstone**, so the server soft-deletes live geometry the user just edited. A deletion of such an id flips positive and the server tries to tessellate an element that no longer exists. This updater applies across **all** model categories, widening exposure that previously only touched clash's nine.

**Why it's not a one-line fix.** The `int` element-id type threads through the whole delta path:
`LiveClashUpdater.GeometrySyncQueue` (tuple type) → `DrainGeometrySyncIds` → `GeometrySyncPlan.Partition` / `BuildRetrySet` → `GeometrySyncHandler` → `PlanscapeServerClient.PostGeometryDeltaAsync(IList<int>)`, and on the server `FederatedModelController.delta` reads `deletedIds` as a JSON array of `int` (documented as such) and matches `FederatedElement.ElementId` (also `int`, so a DB migration is likely).

**Correct fix:** widen element ids `int → long` across the plugin delta path **and** the server contract + `FederatedElement.ElementId` (+ migration), with a plugin↔server round-trip test. A partial widening (plugin-only) would give false confidence while the server half still truncates, so this deserves its own focused change rather than a patch cascaded through the reliability branches.

StingD85:
## Resolved (core) — F1 element-id widening landed

The data-loss half of F1 is fixed and tested across the plugin and server:

- **Plugin (`a87277f3a`, this branch):** widened the whole geometry-delta id path `int → long` — `GeometrySyncQueue`, its drain, `GeometrySyncPlan.Partition`/`BuildRetrySet`, the handler's changed/deleted/extracted lists + `TryExtractElement`, `RequeueForRetry`, and `PlanscapeServerClient.PostGeometryDeltaAsync(IList<long>)`. A large modified/added id no longer wraps negative and gets mis-split as a tombstone, so the server can't soft-delete live geometry the user just edited.
- **Server (`e43a20209`, on `federation-delta-atomic` above):** the delta endpoint now deserializes `deletedIds` as `List<long>` (it previously overflow-threw on a 64-bit id and silently dropped the tombstone). `FederatedElement.ElementId` and the GLB node DTO were already `long`.
- **Tests:** `GeometrySyncPlanTests.A_large_modified_id_stays_a_change_not_a_tombstone` (plugin) and `A_deleted_id_beyond_int_range_round_trips_through_the_wire` (server). Plugin builds 0 errors; plan tests 9/9; full server suite 770 passed / 0 failed / 15 skipped.

### Remaining — separate, clash-wide (not in this fix)

The **changed-element identity** still truncates: `TryExtractElement` builds `ClashElementKey((int)el.Id.Value)`, which becomes the GLB node's `elementId`. So a `>int.MaxValue` element stores a truncated id on add and a later 64-bit delete won't match it. Fixing this means widening `ClashElementKey.ElementId`, which cascades through the entire clash engine — `ClashSession`'s `Dictionary<int,…>` mesh/OBB/neighbour maps, `ClashHistory`, `ClashGrouper`, and the clash `ClashNotifications`/`ClashRunCommand` DTOs and **their server contract** (all `int`). That's a clash-wide id migration, deliberately out of scope here and worth its own change.


> **The remaining clash-wide residual is now tracked as #722.**

PR #685, opened 2026-08-16
Reliability C5+C6: make the geometry delta atomic and idempotent, and honour the upload caps

StingD85:
> **Stacked on #684** (C1+C2) → … → #676. Retarget as those merge.

## C5 — two endpoints advertised caps their own parser would reject

`RequestSizeLimit` caps the raw HTTP body; the **multipart parser** has a separate ceiling, and `Program.cs` sets a global **200 MB** `FormOptions.MultipartBodyLengthLimit`.

| Endpoint | Advertised | `[RequestFormLimits]`? | Reality |
|---|---|---|---|
| `IfcIngestController.Ingest` | 2 GB | ❌ | died at 200 MB |
| `FederatedModelController.PostDelta` | 256 MB | ❌ | died at 200 MB |
| `ModelsController.Upload` | 500 MB | ✅ | fine |

Anything between 200 MB and the advertised cap died with a bare parser **HTTP 400 raised before the action ran** — so the ingest controller's own `file_too_large` branch never fired and the caller got no usable reason. Both now match their documented contract.

## C6 — the delta apply was neither atomic nor idempotent, and swallowed bad input

**Non-atomic.** Deletes committed immediately via `ExecuteUpdateAsync`; the GLB store and `SaveChangesAsync` came afterwards, no transaction. A failure in between left **deletions applied and additions lost** — the worst available partial state, because the server model is then *missing* geometry the source still has, and **no retry can recover it**: the deletions are already marked, so the next delta's `!e.IsDeleted` filter excludes them. Now one transaction, opened only on a relational provider (EF InMemory raises `TransactionIgnoredWarning` and escalates it to an exception).

**Non-idempotent.** A delta is a replayable write — the plugin re-queues and re-sends after a failure (C1), and a client that times out mid-upload can't tell whether the server applied it. Replaying re-applied tombstones and re-wrote rows; a replay whose GLB had been superseded would **resurrect stale geometry**. Now honours the platform's existing `X-Idempotency-Key` mechanism (reused, not reinvented), recorded **inside** the transaction so a replay can't be marked handled by a delta that then rolled back. No key = no protection = previous behaviour, so older plugins keep working.

**Two bare `catch { }` became logged failures.** A malformed `deletedIds` payload dropped every tombstone while the GLB half still applied — HTTP 200, and the elements the caller asked to remove stayed visible *forever*. A malformed/truncated GLB produced zero nodes, indistinguishable from "a delta that legitimately changed nothing"; the bytes are still stored, so it's recoverable — but only if someone knows it happened.

## Verification

**7 tests on SQLite** — a *real* relational provider, so `BeginTransactionAsync` is a real transaction and the rollback assertion means something. Includes the mirror case that a successful delta applies **both** halves, since a rollback firing unconditionally would satisfy the atomicity test while breaking the feature.

The atomicity test was confirmed against the unfixed code by removing the transaction:

```
Failed  A_failure_after_the_deletes_rolls_the_deletes_back
  the element stayed tombstoned after the delta failed — deletes were not rolled back
```

…with the other six still passing.

Full suite against **real PostgreSQL 16: 756 passed / 0 failed / 0 skipped.**

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/FederatedModelController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/FederatedModelController.cs
@@ -36,23 +36,34 @@ public class FederatedModelController : ControllerBase
     private readonly IHubContext<FederatedModelHub>        _hub;
     private readonly IHubContext<NotificationHub>          _notificationHub;
     private readonly Planscape.Core.Interfaces.IFileStorageService _storage;
+    private readonly ILogger<FederatedModelController> _logger;
 
     public FederatedModelController(
         PlanscapeDbContext db,
         IHubContext<FederatedModelHub> hub,
         IHubContext<NotificationHub> notificationHub,
-        Planscape.Core.Interfaces.IFileStorageService storage)
+        Planscape.Core.Interfaces.IFileStorageService storage,
+        ILogger<FederatedModelController> logger)
     {
         _db      = db;
         _hub     = hub;
         _notificationHub = notificationHub;
         _storage = storage;
+        _logger  = logger;
     }
 
     // ── POST delta ───────────────────────────────────────────────────────────
 
+    private const long MaxDeltaBytes = 256L * 1024 * 1024;
+
     [HttpPost("delta")]
-    [RequestSizeLimit(256 * 1024 * 1024)] // 256 MB cap per delta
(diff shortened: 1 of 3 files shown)


StingD85:
**Restack update**

Rebased onto the updated `delta-reliability` base; stack now rooted on latest `main`. No change to this PR's own commits.

Head is now `7d9b49ade`.

PR #686, opened 2026-08-16
Reliability C4: cascade model deletion, purge after the promised grace, and honour Force

StingD85:
> **Stacked on #685** (C5+C6) → #684 → … → #676. Retarget as those merge.

Three defects in the model lifecycle. Each makes the product **say one thing and do another.**

## Soft-delete did not cascade

`ModelsController.Delete` set `DeletedAt` and nothing else. `GetScene` and the federation manifest filtered on `SceneNode.DeletedAt` — **a column nothing ever set** — so a deleted model kept streaming into the viewer and kept answering element queries.

The model was "deleted" everywhere except where a user would notice. *Deleting something and still seeing it is the kind of bug that makes people stop trusting the tool.*

Both read paths now exclude chunks whose **model** is deleted, and the delete retires chunks + federated elements with it.

## The promised purge job did not exist

`ProjectModel.DeletedAt`'s own doc comment:

> *"Soft-delete — kept for audit; file is purged by a Hangfire job after 30 days."*

There was no such job. Every soft-deleted model's GLB, element map and thumbnail stayed in object storage **indefinitely** — on a project that re-publishes weekly, that's the entire history of the model, billed per GB/month, with nothing in the product that could ever remove it.

*A documented retention promise that nothing implements is worse than no promise — it's the reason nobody goes looking.*

`ModelPurgeJob` now runs nightly on the heavy queue, deleting **bytes before the row**: if byte deletion fails the row is retained so the next run retries, because a row-less blob is unfindable and therefore unrecoverable cost.

## `Force` was never read

The flag has been on `UploadModelRequest`, documented, and sent by the plugin's *"Publish as a new revision (forced)"* mode since it shipped. The dedup branch ran regardless — so a coordinator who deliberately chose to cut a new revision got a **metadata refresh on the old row** and no new revision. And it reported success.

It couldn't simply insert a second row: the unique filtered index on `(TenantId, ProjectId, ContentHash) WHERE DeletedAt IS NULL` rejects identical bytes. `Force` now **retires** the previous row (freeing the index) and links it forward via a new `SupersededByModelId`, so history stays queryable and *"why did this model disappear?"* has an answer pointing at its replacement.

## Verification

**5 tests** on the purge job, including the two mirror cases that stop it being a data-loss machine:
- a **recently**-deleted model is *kept* for the grace period — a purge ignoring the cutoff would pass the happy-path test and destroy the audit trail the soft-delete exists for;
- a **failed byte deletion keeps the row**.

Also: the job works with **no tenant context**, since Hangfire runs it that way and the global filter would otherwise match nothing and purge zero rows forever — the exact failure mode it exists to fix.

Full suite against **real PostgreSQL 16: 761 passed / 0 failed / 0 skipped.**

### Worth recording

The first Postgres run after adding `SupersededByModelId` failed four placement tests: `EnsureCreated()` short-circuits on an existing database and the test DB predated the column. That's **ADR-0001 in miniature** — the startup patcher exists precisely for this. The patch statement was verified separately against a pre-existing table: applied cleanly, idempotent on a second pass, legacy row untouched.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/ModelsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ModelsController.cs
@@ -141,6 +141,30 @@ public class ModelsController : ControllerBase
         }
         var existing = await _db.ProjectModels
             .FirstOrDefaultAsync(m => m.TenantId == project.TenantId && m.ProjectId == projectId && m.ContentHash == hash && m.DeletedAt == null, ct);
+
+        // C4 — honour Force.
+        //
+        // The flag has been on UploadModelRequest, documented, and sent by the
+        // plugin's "Publish as a new revision (forced)" mode since it shipped —
+        // and read by nothing. The dedup branch below ran regardless, so a
+        // coordinator who deliberately chose to cut a new revision got a
+        // metadata refresh on the OLD row and no new revision at all. The
+        // operation reported success, which is why it went unnoticed.
+        //
+        // It could not simply insert a second row: the unique filtered index on
+        // (TenantId, ProjectId, ContentHash) WHERE DeletedAt IS NULL rejects
+        // identical bytes. Retiring the previous row frees the index AND makes
+        // the supersede explicit, so the history stays queryable.
+        if (existing != null && req.Force)
+        {
+            existing.DeletedAt = DateTime.UtcNow;
+            await _db.SaveChangesAsync(ct);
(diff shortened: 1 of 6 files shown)


StingD85:
**Restack update**

Rebased onto the updated `delta-atomic` base; stack now rooted on latest `main`. No change to this PR's own commits.

Head is now `5bf58c96c`.

PR #687, opened 2026-08-16
Reliability C3: propagate deletions from hosts that push full exports

StingD85:
> **Stacked on #686** (C4) → #685 → #684 → … → #676. Retarget as those merge.

## The bug

The IFC ingest is an **upsert** over the elements it carries, so an element that disappears from the source is simply *not mentioned* — and "not mentioned" is indistinguishable from "unchanged, and this was a partial push".

A wall deleted in ArchiCAD therefore stayed on the server **forever**: visible in the viewer, answering clash and compliance queries, counted in every metric.

**Absence cannot mean deletion.**

## Server

`IfcIngestRequest` gains `RemovedGlobalIds`, and `/ifc/data` accepts a **removals-only** push — a save whose only change was deleting elements has nothing to upsert, and rejecting it as `Elements is empty` would drop exactly the deletions this exists to deliver.

Matching `TaggedElement`s are **soft**-deleted: the entity carries tag history, issues and clash references, so hard-deleting would break those and make an accidental deletion in the authoring tool unrecoverable. The global query filter already hides tombstoned rows from every read path — which is the observable behaviour that was missing.

### The part that matters most

Removals are resolved through `ExternalElementMapping` and **limited to the host document that reports them.**

Two hosts contribute to one project, so a full-export diff from the ArchiCAD file lists every Revit-contributed GlobalId as "absent" — and *if absence were enough, that push would wipe the Revit model.* Without the scope, this feature converts a missing-delete bug into a **data-loss** one. `A_host_cannot_remove_another_hosts_elements` is the test that keeps that honest.

## Client

`SyncedIdStore` records the GlobalIds each document last pushed, keyed on `(project, host document)` for the same reason. `diff_removals` turns a full export into an explicit removal list; `ingest_ifc_data` carries it.

`should_send_removals` adds a guard the server cannot provide. The diff is only as trustworthy as the export it's diffing — a crashed or filtered export yielding 3 elements instead of 30,000 looks, at this layer, **exactly** like a coordinator deleting almost the whole model. One reading is catastrophic and not recoverable by retrying; the other costs one sync cycle. So a removal set covering >50% of known elements is **refused and reported**, with the numbers in the message so an operator can judge it.

## Verification

- **16 core tests** — document/project isolation; a first-ever push removes nothing (*no baseline means no knowledge, not "everything is gone"*); removals sorted so two runs over an unchanged file produce byte-identical payloads; and the exactly-half boundary, because an off-by-one there silently changes a product decision.
- **8 server tests** — the load-bearing one being that an ArchiCAD push cannot tombstone a Revit-contributed element.
- **769 C# passed / 0 failed / 0 skipped** against real PostgreSQL 16; **313 Python passed.**

## Noted, not fixed

`IfcController` carries `[Authorize]` and a tenant check but **no** `[ProjectAccess]` / membership gate — the same defect class Track A fixed on five sibling controllers. Out of the stated scope, and adding it changes the plugin's push path for any service account that isn't a project member, which needs verification rather than assumption. Logged in `docs/ROADMAP.md` along with the rest of the follow-ups.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/IfcController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/IfcController.cs
@@ -47,12 +47,18 @@ public class IfcController : ControllerBase
     private readonly PlanscapeDbContext _db;
     private readonly IIdentityResolverService _identity;
     private readonly IIfcIngestService _ingest;
+    private readonly ILogger<IfcController> _logger;
 
-    public IfcController(PlanscapeDbContext db, IIdentityResolverService identity, IIfcIngestService ingest)
+    public IfcController(
+        PlanscapeDbContext db,
+        IIdentityResolverService identity,
+        IIfcIngestService ingest,
+        ILogger<IfcController> logger)
     {
         _db = db;
         _identity = identity;
         _ingest = ingest;
+        _logger = logger;
     }
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


StingD85:
**Restack update**

- Added a **test-only** adaptation so `Planscape.Tests` compiles and passes now that this branch sits above the F1 gate (from #676) and carries its own `ILogger<IfcController>` addition:
  - `FederationAuthorizationTests.NewIfcController` now passes `NullLogger<IfcController>.Instance` (the ctor param this branch adds).
  - `IfcRemovalPropagationTests` now authenticates the host push as a real project member (`AppUser` + active `ProjectMember`). `IngestData` is gated by `RequireProjectMemberAsync`, and a removals push mutates project identity, so it requires membership. Two host-scoping tests had been passing only because the 403 also produced "nothing removed"; they now exercise the real removal path.
- Rebased onto the updated `model-lifecycle` base (now on latest `main`).
- **Note:** during the restack a fork-point slip briefly pointed this branch at a commit already in its base, which GitHub auto-closed the PR over. Recovered (the feature commits were intact) and reopened with the correct head.

Head is now `4616511bb`.

PR #688, opened 2026-08-16
Reliability C7: record IFC conversion status, and stop two orphans claiming they ship

StingD85:
> **Stacked on #687** (C3) → #686 → … → #676. Final Track C PR; also carries the PART V write-up.

## Conversion status

`IfcToGlbConversionJob` is best-effort **by design**: a sidecar error is logged and never thrown, so a failed convert leaves the IFC stored and re-uploadable. That's the right call. But nothing recorded the failure anywhere the **product** could see — while the upload endpoint had already answered `202` with:

> *"a renderable GLB derivative is being generated and will appear as a separate model shortly."*

For a conversion that failed, **that sentence never stops being false.** The coordinator waits, refreshes, waits again, and concludes the platform is broken — a fair reading of a promise that's never withdrawn. The job's log line isn't an answer: it's on the worker, and the person waiting is in a browser.

`ProjectModel` gains `ConversionStatus` + `ConversionError`. The job marks `Converting` on entry and a **terminal** state on every one of its **five** early returns — each of which previously left the model advertising itself as converting forever.

An unconfigured converter is deliberately **not** marked failed: the upload endpoint only promises a derivative when a converter is configured, so reporting a broken conversion that was never attempted would be its own false statement.

Status is stored as **text, not an enum**, so a value written by a newer worker can't fail to deserialise on an older API instance mid-deploy. Writing it is best-effort by the same logic as the job — it must never turn a *successful* conversion into an exception.

## Two orphans that claimed to ship

`FederationLinkedWalker` and `AsBuiltReconciler` have **zero callers**. The former's header claimed it was *"used by Clash triage (S6.1), As-built reconciliation (S6.5), and any BCC dashboard metric that needs 'X across all linked models'"*. None of those consume it, and **cross-tool clash does not exist at all.**

Both are **kept, not deleted** — the code is sound and the walk `FederationLinkedWalker` implements is exactly what a federated metric needs, so rewriting it later costs more than carrying it. Their headers now say **NOT WIRED**.

*An orphan is a cost; an orphan advertising itself as shipped is a trap* — someone auditing whether federated clash works finds a docstring asserting it does, and stops looking.

## Verification

**5 tests.** The load-bearing one asserts that **no terminal path leaves a model reading "Converting"** — that's the state a coordinator waits on forever, and it matters more than any individual branch. Also: the failure *reason* travels (not just "Failed", which sends the coordinator to support when the cause is usually actionable by them), and an unconfigured converter leaves the status alone.

- Full suite against **real PostgreSQL 16: 774 passed / 0 failed / 0 skipped**
- Revit plugin: **0 errors, 0 warnings**

### Note on the DTO

`ModelMetaDto` is a **positional record**, so the new fields are appended with defaults rather than inserted — adding a parameter mid-list silently re-maps every argument after it. My first attempt did exactly that, and the compiler caught it only because a later argument happened to be named.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/ModelsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ModelsController.cs
@@ -289,6 +289,10 @@ public class ModelsController : ControllerBase
             BoundsMaxX = req.BoundsMaxX,
             BoundsMaxY = req.BoundsMaxY,
             BoundsMaxZ = req.BoundsMaxZ,
+            // C7 — an IFC is not renderable until the sidecar converts it. Say so
+            // in the data, not only in the 202 response body, so the viewer and
+            // the dashboard can tell "still working" from "gave up".
+            ConversionStatus = willConvertIfc ? "Pending" : null,
             UploadedBy = User.FindFirst("display_name")?.Value ?? User.Identity?.Name ?? "",
             UploadedByUserId = CurrentUserId(),
             UploadedAt = DateTime.UtcNow,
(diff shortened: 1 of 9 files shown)


StingD85:
**Restack update**

- Rebased onto the updated `delete-propagation` base; stack now rooted on latest `main`. No change to this PR's own commits.
- **Note:** like #687, a fork-point slip during the restack briefly collapsed this branch onto a base commit and GitHub auto-closed the PR. Recovered and reopened with the correct head.

Head is now `9a3d6fab4`.

PR #737, opened 2026-08-18
fix(server): project settings had a gate nobody could pass; route it through a capability

StingD85:
## This widens server authority

**Who may edit project settings goes from NOBODY to project managers and above.** That is the intended effect and it was approved explicitly. It is not a refactor, and this PR does not describe it as one.

## The measurement that motivates it

`ProjectSettingsController:149` required the caller's `ProjectMember.Iso19650Role` to be `"K"` or `"C"`. Its docstring called those "a project role of K (BIM Manager) or C (Coordinator)". **Neither code is assignable.**

There are three incompatible ISO vocabularies in this codebase:

| Source | Vocabulary | K? | C? |
|---|---|:--:|:--:|
| `GET /members/roles` — what the server serves, what the web grid picks from | A·PM·BC·BA·AR·SE·ME·CE·QS·CA·CT·SC·FM·OM·CL·M·V·Z | ✗ | ✗ |
| `AppUser.Iso19650Role` (AppUser.cs:14, default `"Z"`) | A·M·E·S·H·P·C·I·K·Q·F·W·L·Z | ✓ | ✓ |
| `ProjectMember.Iso19650Role` (default `"M"`, validated by nothing) | populated from the **first** list | — | — |

The check was written against the **AppUser** vocabulary and applied to the **ProjectMember** column.

Measured on the local stack, 2026-08-18, 34 ProjectMember rows:

```
K or C: 0
A=11  M=9  BC=8  PM=2  QS=1  S=1  AR=1  EL=1
```

**And K/C have zero rows in `AppUser.Iso19650Role` either.** That kills the last charitable reading — this was not a column mix-up that used to work for somebody and drifted. The gate has matched nobody for its whole life. Plausibly nobody has ever been able to edit project settings.

Same class as the fifteen `ProjectRole == "PM"` sites #540 fixed, and the twelfth found in review.tsx.

## The fix

A third capability on `ProjectRoles`, alongside the two #540 added:

```
CanAdministerProject = tenant Admin/Owner
                    OR ProjectRole in { Manager, Owner, Admin }
                    OR Iso19650Role in { PM, A, BC }
```

Both shapes, kept in step as the existing pairs are: an in-memory form and an EF-translatable `Expression`. One test asserts they agree on every seeded row; another asserts the Expression reaches SQL rather than client-evaluating.

**No ninth authorization mechanism.** This is a third predicate on the class that already owns the second and third — the count goes down over time, not up.

## The model this encodes

Three layers, one job each:

| Layer | Question | Where | Changes |
|---|---|---|---|
| **Capability** | may I DO this? | `ProjectRoles` (#540) | rarely |
| **Information ACL** | may I SEE this? | CDE / discipline / suitability (#665) | per member |
| **ISO 19650 role** | who is ACCOUNTABLE? | `ProjectMember.Iso19650Role` | per appointment |

**Nothing gates on layer 3.** A capability may *consult* the ISO role as one input — which is what the `PM`/`A`/`BC` terms are, and what `CanCurateProject` already does. What must never happen again is a raw letter comparison at a call site.

The reasoning is in the code so it survives: ISO 19650 defines information-management *functions* (appointing party, lead appointed party, appointed party) that move with appointments and can change mid-project on novation. They answer "who is accountable for this deliverable", not "who may perform this operation". Those change on different clocks, and binding them makes every appointment change a permissions migration. The single-letter codes are a house convention, not part of the standard — which is exactly why two incompatible versions exist. Where ISO 19650 *does* prescribe access semantics is the CDE (WIP/Shared/Published/Archive plus suitability), and ISO 19650-5 asks for need-to-know against the **information**; that axis is the ACL, not this class.

## Deliberately no legacy tolerance

`CanCurate` accepts a `ProjectRole` of `"PM"` because the gate it replaced tested that column, so dropping it would have **narrowed** access for such a row. There is no equivalent obligation here: the replaced gate granted access to nobody, so no row can lose anything. Adding terms beyond the three above would widen on a guess.

## One incidental fix

The replaced check looked up a `ProjectMember` row first and returned `Forbid` when it found none — so a **tenant Owner with no member row was refused their own tenant's project settings**. The capability layer resolves the tenant claim before touching the table, which is behaviour every other capability site already had. Tested.

## Tests — RED before GREEN

With the old gate restored and the capability already in place, **exactly the six widening assertions fail** and the other 24 pass. With the fix: **71/71** across all three capability test files (`ProjectAdministerCapabilityTests`, `ProjectRoleCapabilityTests`, `ProjectCapabilitiesEndpointTests` — no regression in the siblings).

`The_replaced_gate_matched_nobody` runs the **old predicate** against a fixture holding a Manager, an Owner, a Coordinator and three ISO office-holders, and asserts it selects none of them. The "widens from NOBODY" claim above is therefore a test result, not a sentence.

Fixture traps avoided, per the sibling files: contexts always built with a tenant (the `Guid.Empty` fallback matches no rows and makes assertions pass vacuously), identifiers minted per fixture, `Sanity_the_fixture_actually_has_rows` proves non-empty first, and SQLite rather than EF InMemory so the translation test means something.

## Before merging — one query

`docs/sql/check_iso19650_role_population.sql`, read-only, safe for Render's psql shell. It **leads with the only count that changes the work**:

```sql
SELECT count(*) FROM "ProjectMembers" WHERE "Iso19650Role" IN ('K','C');
```

- **Zero** → the gate is dead in production too and this PR widens from nobody, as framed.
- **Non-zero** → **STOP.** Those people can edit settings today, and `CanAdministerProject` as specified may not include them, so this could take access away rather than only grant it. Different PR, different justification.

The stray-value distribution is in the same script, clearly labelled as **not** decision-changing.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/ProjectSettingsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ProjectSettingsController.cs
@@ -4,6 +4,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Planscape.Infrastructure.Data;
 using Planscape.API.Authorization;
+using Planscape.API.Services;
 
 namespace Planscape.API.Controllers;
 
@@ -130,8 +131,29 @@ public class ProjectSettingsController : ControllerBase
     }
 
     /// <summary>
-    /// Project admin can overwrite the overrides JSON. Caller must have a
-    /// project role of K (BIM Manager) or C (Coordinator).
+    /// Project admin can overwrite the overrides JSON.
+    ///
+    /// Authorization is the <c>CanAdministerProject</c> capability
+    /// (<see cref="Planscape.Core.Entities.ProjectRoles"/>): tenant Admin/Owner,
+    /// or a ProjectRole of Manager/Owner/Admin, or an ISO 19650 role of
+    /// PM / A / BC.
+    ///
+    /// THIS WIDENS ACCESS, AND THE PREVIOUS RULE IS WORTH RECORDING. This
+    /// endpoint used to require <c>member.Iso19650Role</c> to be "K" or "C" —
+    /// described here as "a project role of K (BIM Manager) or C (Coordinator)".
+    /// Neither code is assignable. <c>GET api/projects/{id}/members/roles</c>,
+    /// the vocabulary this same server serves and the web grid picks from,
+    /// offers A/PM/BC/BA/AR/SE/ME/CE/QS/CA/CT/SC/FM/OM/CL/M/V/Z — no K, no C.
+    /// K and C belong to a different list, the one on
+    /// <c>AppUser.Iso19650Role</c>, so the check was written against one column
+    /// and applied to another. Measured 2026-08-18 against the local stack: zero
+    /// ProjectMember rows carry K or C — and zero AppUser rows do either, so it
+    /// was never a column mix-up that happened to work. Nobody could edit
+    /// project settings.
+    ///
+    /// The gate now consults the ISO role as ONE INPUT to a capability rather
+    /// than comparing letters at the call site, which is the practice that
+    /// produced this bug and the eleven dead <c>ProjectRole == "PM"</c> gates.
     /// </summary>
     [HttpPut]
     public async Task<ActionResult> UpdateSettings(Guid projectId, [FromBody] Dictionary<string, object?> overrides)
@@ -141,12 +163,11 @@ public class ProjectSettingsController : ControllerBase
             .FirstOrDefaultAsync(p => p.Id == projectId && p.TenantId == tenantId);
         if (project == null) return NotFound();
 
-        var userIdClaim = User.FindFirst("user_id")?.Value;
-        if (!Guid.TryParse(userIdClaim, out var userId)) return Forbid();
-
-        var member = await _db.ProjectMembers
-            .FirstOrDefaultAsync(m => m.ProjectId == projectId && m.UserId == userId && m.IsActive);
-        if (member == null || (member.Iso19650Role != "K" && member.Iso19650Role != "C"))
+        // One capability, resolved server-side. It also handles the tenant
+        // Admin/Owner claim, which the replaced check ignored entirely — a
+        // tenant Owner with no ProjectMember row was refused their own
+        // project's settings.
+        if (!await this.CanAdministerProjectAsync(_db, projectId))
             return Forbid();
 
         // Phase 144 — split admin booleans (first-class columns) from soft
--- a/Planscape.Server/src/Planscape.API/Services/ControllerProjectMembershipExtensions.cs
+++ b/Planscape.Server/src/Planscape.API/Services/ControllerProjectMembershipExtensions.cs
@@ -61,6 +61,22 @@ public static class ControllerProjectMembershipExtensions
         => HasCapabilityAsync(controller, db, projectId, ProjectRoles.CanApproveSitePhotosPredicate, ct);
 
     /// <summary>
+    /// True when the caller may ADMINISTER this project — edit project-level
+    /// settings (ISO naming enforcement, the custom deliverable state machine,
+    /// the preferences blob).
+    ///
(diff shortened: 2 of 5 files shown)


StingD85:
## Merge gate — one number decides this, and it is not in the diff

This PR is `OPEN`, `MERGEABLE`, `CLEAN`. **It must not merge until the owner runs one query in Render's psql shell.** The query ships in this PR at `docs/sql/check_iso19650_role_population.sql`; it is read-only, touches no personal data, and takes a second.

```sql
SELECT count(*) AS "rows_that_can_edit_settings_today"
FROM   "ProjectMembers"
WHERE  "Iso19650Role" IN ('K', 'C');
```

### How to read the answer

**`0` → merge.** The gate is as dead in production as it is locally: `Iso19650Role IN ('K','C')` is the condition `ProjectSettingsController` requires, and nobody satisfies it. This PR then widens who may edit project settings **from nobody to project managers and above**, which is exactly what the PR body claims and is a pure grant.

**Non-zero → STOP, and do not merge as framed.** Those people can edit project settings **today**. `CanAdministerProject` as written may not include them, so this change could **remove** access rather than only grant it. A PR that takes access away from live users is a different change needing a different justification, and that is the owner's call, not a reviewer's. Say so here and leave it open.

### Why this cannot be settled from the code

The number is a property of production **data**, not of the diff. Locally it is `0`, which is why the PR is framed as a widening — but a local database that was seeded rather than used is not evidence about production, and the whole point of the gate is that it depends on which humans happen to hold which role letter. Merging on the local number would be asserting something nobody measured.

### Everything else in that file is context, not a decision input

`check_iso19650_role_population.sql` also reports the full `ProjectMember` ISO-role distribution and the `AppUser` one. Those are useful — the local run found two strays (`'S'` and `'EL'`, one row each) and evidence that `AppUser.Iso19650Role` has drifted from its own declared vocabulary — but **none of it changes whether this PR should merge**. Strays are tolerated by design in the follow-up work: new writes get validated without making an existing row unsaveable. Read the distribution only if you are curious, or if query 1 came back non-zero.

Do not let the stray-value output delay the merge decision. One number, two outcomes.

StingD85:
## Addendum — the odds of that query returning non-zero just went up, and here is the mechanism

While validating the ISO vocabulary on write, I found where `'K'` and `'C'` come from. It is not arbitrary.

`Planscape/app/project-settings/members.tsx:41` — the **mobile member editor** — hardcoded its own role list:

```ts
const ISO_ROLES = ['K', 'C', 'TI', 'L', 'AP', 'LAP'];
```

**`K` and `C` are the first two entries**, and that screen `PUT`s the selected value into `ProjectMember.Iso19650Role`. None of the six appear in the vocabulary `GET api/projects/{id}/members/roles` serves (`A/PM/BC/BA/AR/SE/ME/CE/QS/CA/CT/SC/FM/OM/CL/M/V/Z`), and the column accepted any string, so every save from that screen landed a value no gate could interpret — except `ProjectSettingsController`, which was written to match exactly the first two.

So this is a **third** vocabulary for one concept, alongside `ProjectMember`'s and `AppUser`'s.

### What this changes for the merge decision

**The decision rule itself is unchanged.** Zero means merge; non-zero means stop. But the prior probability shifts: `'K'` and `'C'` are not values that could only arrive by accident — they are the **default two options at the front of a picker a project manager taps**. Anyone who edited a member's ISO role on mobile most likely wrote one of them.

Locally that count is `0` because nobody used that screen against this database. That is now clearly a fact about the local data and not about the feature, which is exactly why the production query matters rather than being a formality.

If it comes back non-zero, the reading in my earlier comment holds and gets sharper: those people can edit project settings today **because a mobile picker gave them `K`**, not because anyone assigned them authority. Taking it away may well be right — but it is then a deliberate revocation, not the pure widening this PR is framed as, and it should say so.

### The picker is fixed separately

The mobile screen now fetches the vocabulary from `GET .../members/roles` instead of hardcoding it, so it can no longer produce `K`/`C`/`TI`/`L`/`AP`/`LAP`. That ships in the ISO-validation PR, not here — it is independent of this one and does not change what the query returns for rows already written.

Nothing in this addendum is a reason to merge or not merge before the number is known.

Issue #631, opened 2026-08-06
security: ProjectMemberAcl enforces nothing — all three ACL columns are projected to null


StingD85:
Found while measuring 403 behaviour for #558. This is a gate that reports as present and enforces nothing.

## The defect

`Planscape.API/Authorization/ProjectMemberAcl.cs`, `ResolveAsync` projects all three ACL columns to a literal null:

```csharp
.Select(m => new
{
    m.Id,
    AllowedCdeStates     = (string?)null,
    AllowedDisciplines   = (string?)null,
    AllowedSuitabilities = (string?)null,
})
```

Since `null` means "all" by this class's own documented convention, every member resolves to an unrestricted slice. `ApplyTo` therefore never narrows a query, and the Phase 177 per-folder / per-discipline / per-suitability ACL has never restricted anything.

**Measured live, not inferred:** a member restricted to `WIP,SHARED` was still allowed to request `PUBLISHED`, and the server answered **201**.

## Why it is still here

The projection carries a comment explaining it avoids crashing when the `AddProjectMemberAcls` migration has not been applied. That reasoning is obsolete: schema on every environment comes from `OnModelCreating` plus `PlatformSchemaPatcher`, not from migrations (ADR-0001, and `PLANSCAPE_USE_ENSURE_CREATED=true` in `render.yaml`). The columns exist. A temporary defensive stub became permanent.

I reviewed this code earlier in the same session, read the comment, and accepted it as deliberate. The comment is persuasive and the behaviour is not what it says. That is the general lesson worth recording: a defensive stub with a good comment reads exactly like a design decision.

## What correct looks like

Read the three columns. Keep null-means-all, because that is the documented semantic and existing rows rely on it — a member with no allow-list must not suddenly be restricted to nothing.

Do NOT simply delete the stub without checking what starts being enforced: this turns a currently-inert control into a live one, so establish first how many members have populated allow-lists. If the answer is zero, the change is safe and inert. If it is not zero, those members' access narrows the moment it ships, and that needs to be a deliberate decision rather than a side effect.

## Verification

Reproduce the 201 above as a RED case: a member restricted to `WIP,SHARED` requesting `PUBLISHED` must be refused once fixed. Assert on non-empty data — the `Guid.Empty` tenant fallback makes empty-result tests pass vacuously.

PR #739, opened 2026-08-20
fix(billing): decouple the tenant account ceiling from paid seat totals

### LINKED ISSUE #653: tenant.MaxUsers still caps free viewers — decouple it from TotalSeats
## The gap

PR #626 retires the .NET per-user quota axes, because seat entitlement now lives in the StingTools licence recorded in D1 (#621). That removes the `402 quota_exceeded` gate on invites.

Once it merges, **`tenant.MaxUsers` becomes the only bound on account creation** — and it is derived from `TotalSeats` (`AuthController:463` and `:1098`), enforced at `ProjectMembersController:310` and `AdminController:86` as `userCount >= MaxUsers` over active `AppUser` rows.

## Why that is wrong

The commercial position decided this week is: **author seats are the only paid unit; viewers and coordinators are free and unlimited**, including the client (Appointing Party).

`MaxUsers` derived from `TotalSeats` contradicts that. Studio is `1 + 5 = 6`, so a Studio firm is refused its **7th account including a free viewer**. "Viewers are free" is therefore only half-true, and it fails at exactly the moment a customer tries to bring their client or a site engineer onto the platform — the adoption behaviour the free tier exists to encourage.

This shortfall was flagged honestly in #607 before that PR was superseded; it needs recording so it is not lost with it.

## Why it cannot simply be removed

`MaxUsers` is the last thing bounding account creation once #626 lands. Removing it outright leaves nothing between a tenant and unbounded automated signup.

## Why it cannot be licence-derived

D1 meters **machines** holding a StingTools licence. Viewers and coordinators hold no licence at all, so there is no licence-side number that corresponds to a total-account count.

## Proposal

Make `MaxUsers` an **abuse ceiling decoupled from `TotalSeats`** — a single large flat number that stops runaway automated signup, not a commercial cap that scales with plan.

- Remove the `TotalSeats` derivation at `AuthController:463` / `:1098`.
- Choose a value high enough that no legitimate practice reaches it.
- Keep the enforcement sites as they are; only the source of the number changes.

## Acceptance

A tenant on any plan can add an unlimited number of read-only accounts. Paid author seats remain bounded by the licence count in D1. Runaway signup is still refused.

## Related

- #621 — licence presentation; D1 owns entitlement
- #626 — retires the .NET per-user quota axes (must merge before this matters)

StingD85:
Closes #653.

`Tenant.MaxUsers` bounds **accounts**. It was derived from
`BillingPlanLimits.TotalSeats` — `MaxAuthors + MaxCoordinators`, two **paid role**
caps. A free viewer therefore consumed a paid allowance, and a Studio firm was
refused its 7th account at `1 + 5 = 6`.

We say the opposite in public. `marketing-site/pricing.html:351`, live on the
pricing FAQ:

> They get role-restricted access to the project you've invited them to (typically
> Viewer or Client role). **They don't count against your coordinator cap** as long
> as their firm doesn't also need to author models.

They did count. And since #626 removed the 402 invite gate, `MaxUsers` is the only
bound on account creation left, so it could not simply be deleted.

## Two further defects, both live on `main`

**The derivation did not use the plan the tenant is assigned.** `AuthController:463`
read the limits of the plan the caller *requested* while assigning
`BillingPlan.Trial`:

| `req.Plan` on an anonymous signup | resulting `MaxUsers` |
|---|---|
| omitted | 20 (Network) |
| `"Studio"` | 6 |
| `"Enterprise"` | `int.MaxValue` |

So a signup chose its own abuse ceiling — precisely what the ceiling exists to
stop. The often-repeated "this is latent because both creation paths assign Trial
(1 + 0)" was **not true of this path**.

**The handoff path contradicted its own comment.** `AuthController:1098` hardcoded
Network → **20**, directly beneath:

> Provision the mirror generously so this side never locks out a customer D1
> considers paid; reconciliation is deliberately out of scope.

20 locks out any D1-paid firm larger than that. Worse, this is the path that maps
unknown roles **down** to `UserRole.Viewer`, so it manufactures the very free
accounts a seat-derived cap charged for.

## What changed

- **`BillingPlanLimits.AccountCeiling`** — flat 10,000, independent of the role
  axes, written at both creation paths and now the entity default (was 5). An
  anti-abuse bound, not a commercial one; paid entitlement is metered by the D1
  licence count, which counts *machines* and so can never bound viewers.
- **`AccountCeilingPolicy`** — one place the cap is read. Both enforcement sites
  open-coded `count >= MaxUsers`, and that comparison **cannot express
  "unlimited"**: every non-positive value — this repo's own `-1` sentinel from
  `TierLimits`, a `0` from a partial row, the `-2` from #616 — is `>=` true at zero
  users and denies the tenant its **first** account. It now fails **open** on a
  nonsensical cap, so the #616 class is closed at the comparison, not just at its
  one source.
- **`MaxAuthors`/`MaxCoordinators` documented as what they have become.** After
  #626 retired the Authors/Coordinators quota axes and this PR removes the sum,
  **no enforcement decision anywhere reads them**. Their three consumers are all
  presentation (`PricingController`, the signup response, `TenantAdminController`).
  That is the answer to "what are these two fields still for": the pricing table.
- **`TotalSeats` kept but marked display-only** — "up to 6 users" on the pricing
  page *is* that number. Its saturation guard and `BillingPlanSeatTotalTests` stay,
  so if anyone rewires it to a cap again it degrades to "unlimited" rather than to
  the #616 lockout.
- **Boot report** for tenants still carrying a seat-derived cap, at Warning (prod
  runs at `MinimumLevel=Warning`, so Information would be invisible and its absence
  unreadable).

## The one judgment call

Existing rows keep their old cap. **Nothing in the data distinguishes "derived from
a plan" from "an admin deliberately capped this tenant"** — `20` could be either —
so a blind `UPDATE` would silently overwrite deliberate limits on every boot. This
reports instead, naming the tenants and printing the per-tenant statement to run,
following the #631 precedent. If you would rather it rewrite, say so and I will
make it a one-shot patcher statement.

## Verification

- `dotnet build` — **0 errors**; warning set byte-identical to `main`
  (12×NU1603, 6×CS0618, 2×CS1587), checked with `-t:Rebuild`.
- `dotnet test` — **788 passed, 0 failed, 15 skipped**. The 15 are the Postgres
  `SkippableFact`s with no `PLANSCAPE_TEST_PG` set. `known-failing-tests.txt` is a
  deliberate zero-failure gate and stays met.
- `AccountCeilingTests` pins both acceptance criteria from the issue: no plan or
  sentinel can refuse a first account, and the ceiling does not move with the paid
  role axes. Both enumerate `BillingPlan` rather than listing plans, so a plan added
  later is covered without anyone remembering the file exists.
- No production consumer of `TotalSeats` remains — `grep -rn TotalSeats src` returns
  the definition and doc references only.

The pricing FAQ quoted above is now true; no `marketing-site` files are touched, so
nothing deploys to production from this PR.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/AdminController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/AdminController.cs
@@ -86,8 +86,12 @@ public class AdminController : ControllerBase
         if (tenant == null) return NotFound("Tenant not found");
 
         var userCount = await _db.Users.CountAsync(u => u.TenantId == tenantId && u.IsActive);
-        if (userCount >= tenant.MaxUsers)
-            return BadRequest($"User limit ({tenant.MaxUsers}) reached for {tenant.Tier} tier");
+        // Via AccountCeilingPolicy, not a raw >=: a non-positive cap must read as
+        // unlimited rather than denying the tenant its first user (#616, #653).
+        if (!Planscape.Core.AccountCeilingPolicy.Allows(userCount, tenant))
+            return BadRequest(
+                $"Account limit ({Planscape.Core.AccountCeilingPolicy.Label(tenant.MaxUsers)}) " +
+                "reached for this organisation.");
 
         if (await _db.Users.AnyAsync(u => u.Email == req.Email))
             return Conflict($"Email {req.Email} already exists");
--- a/Planscape.Server/src/Planscape.API/Controllers/AuthController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/AuthController.cs
@@ -460,7 +460,13 @@ public class AuthController : ControllerBase
             Plan          = BillingPlan.Trial,
             Currency      = currency,
             BillingCycle  = BillingCycle.Monthly,
-            MaxUsers      = planLimits.TotalSeats,
+            // Flat anti-abuse ceiling, NOT planLimits.TotalSeats (#653). The old
+            // derivation read the limits of the plan the caller ASKED for while the
+            // tenant is assigned Trial, so an anonymous signup chose its own cap —
+            // omitted plan gave 20, "Enterprise" gave int.MaxValue. It also charged
+            // free viewers against paid role caps, which the pricing FAQ says we
+            // don't do. Paid entitlement is metered by the D1 licence count.
+            MaxUsers      = BillingPlanLimits.AccountCeiling,
             MaxProjects   = planLimits.MaxProjects,
             MimEnabled    = false,
             TrialExpiresAt = DateTime.UtcNow.AddDays(30)
@@ -1079,6 +1085,14 @@ public class AuthController : ControllerBase
                 .FirstOrDefaultAsync(t => t.Slug == slug);
             if (tenant == null)
             {
+                // Network only for MaxProjects/display. The account ceiling is
+                // deliberately not taken from a plan here: this mirror must not make
+                // entitlement decisions (D1 does), and Network's seat total is 20 —
+                // which would have locked out any D1-paid firm larger than that,
+                // the exact failure the comment below promises to avoid. Compounded
+                // by this path defaulting unknown roles DOWN to Viewer, so it is the
+                // path that manufactures the free accounts a seat-derived cap
+                // charged for. See #653.
                 var limits = BillingPlanLimits.For(BillingPlan.Network);
                 tenant = new Tenant
                 {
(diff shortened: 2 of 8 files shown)


StingD85:
#740 is stacked on this branch — it fixes the same defect on the **projects** axis and touches `Tenant.cs` and `AuthController.cs` too, so it based off here rather than `main` to avoid a conflict. Merge this one first; #740 then retargets to `main` cleanly.

Worth knowing while reviewing this: the projects axis had a **live** version of the same problem. `CreateProject` carried two gates reading two sources, and the cloud handoff was discarding the `tier` D1 has always sent — so a paying customer was capped at one project. Details in #740.

PR #754, opened 2026-08-22
fix: publishing a federated model published the container, not the model

StingD85:
A federated site — host container, every building and the site itself a Revit link, **all
visible in `{3D}`** — published as a bare toposolid with no buildings. The viewer reported
**8 ELEMENTS / 0% TAGGED** for an entire project.

Two independent faults. Either alone breaks a federated publish.

## 1. The exporter looked every element up in the wrong document

[RevitGltfExporter.cs:114](StingTools/BIMManager/RevitGltfExporter.cs:114) did
`_doc.GetElement(id)` with `_doc` fixed to the host. `OnLinkBegin` *was* implemented and
pushed the transform, so Revit was traversing the links — but **element ids are
document-local**. An id from inside a link, looked up in the host, is not a near-miss; it
is a lookup in the wrong table:

- `null` → `RenderNodeAction.Skip` → **geometry silently dropped.** The missing buildings.
- id exists in the host → geometry emitted carrying **another element's** name, category,
  colour and `UniqueId` — the key the viewer joins metadata on.

`Clash/ClashExportContext` has always done this correctly, with a document stack and
`LinkNode.GetDocument()`. This exporter never did. It now keeps the same stack, and
`OnLinkEnd` pops under the same guard as the transform stack — they are pushed together
and must unwind together, or every element *after* a link resolves against the wrong
document.

Material resolution had the identical bug, and `_appearanceCache` was keyed on the raw
`ElementId` value — so a link's material 12 and the host's material 12 shared one entry,
and whichever was seen first decided the name and texture for both.

## 2. The element map never looked at links at all

[PublishModelCommand.cs:506](StingTools/BIMManager/PublishModelCommand.cs:506) collected
from `new FilteredElementCollector(doc, activeView.Id)` — host only. On a federated model
that is a few link instances and nothing else: the "8 ELEMENTS", exactly. Fixing the
geometry alone gives properties-free meshes; fixing the map alone, propertied nothing.

It now walks links **recursively**, because Revit's view traversal descends nested links
and a one-level walk reproduces this same bug one level down — geometry present,
properties missing, the variant hardest to notice *because the model looks right*.
Documents are visited once (a cycle is legal in Revit) and depth is capped.

## Then: links became a choice, not a behaviour

Including them unconditionally is the wrong default. A discipline publish wants the host
alone — smaller, faster, and the viewer federates published models anyway. A site or
coordination publish wants one artefact. Both are reasonable, so it asks — **only when the
view actually contains links**, since a dialog whose answer cannot change anything is
noise. `PLANSCAPE_EXPORT_LINKS=1` skips it for scripted runs.

**The obligation that comes with defaulting to OFF is that exclusion must be visible.**
Links vanishing silently is the entire reason this option exists, and a deliberate default
is not a licence to reproduce it. The exporter counts skipped links; the publish reports
them in a dialog that names the fix; the log says so too.

`OnLinkBegin` returns `Skip` *without pushing* — Revit does not call `OnLinkEnd` for a
skipped node, so pushing there would leave everything after the first link on the wrong
document and transform.

## Details worth reviewing

- **Keys.** Host elements keep their bare `UniqueId`, so nothing already published changes
  key. Linked elements are namespaced by link path, because `UniqueId` is unique *within* a
  document and buildings are routinely Saved-As from one another — two links can genuinely
  carry the same id. Both producers compute it through one helper; a mismatch renders an
  element with no properties and **errors nowhere**.
- **Bounds.** All eight corners are transformed, not Min/Max — under rotation the
  transformed Min is not the minimum, and an angled link is the normal case, not an exotic
  one. A link's box folded in untransformed drags the published AABB out to enclose empty
  space and the viewer opens zoomed to nothing.
- **The static flag is reset per publish.** Without that, a run that picks an existing file
  inherits the previous run's answer and builds a map that disagrees with the geometry —
  same symptom, arriving only on the *second* publish of a session.
- **Unloaded links are named in the log**, not skipped silently.

## Honest scope limit, stated in the code

The map collects model elements with geometry per link, **not** "elements visible in the
host's active view" — a view id belongs to the host document and cannot filter another
document's collector. The GLB, produced by Revit's own view traversal, remains the
authority on what is drawn. So the map may list a few elements the geometry does not show.
A tree row with no mesh is a much smaller problem than the mesh with no properties it
replaces.

## Verification

`StingTools.csproj` → **0 errors, 0 warnings**. Deployed to the local Revit add-in slot for
testing against the real federated model that surfaced this — the only test that matters
here, since the fault is in Revit API traversal that no headless test exercises.

### DIFF
--- a/StingTools/BIMManager/PublishModelCommand.cs
+++ b/StingTools/BIMManager/PublishModelCommand.cs
@@ -421,6 +421,14 @@ namespace StingTools.BIMManager
 
         private static string? PromptForModelFileOrExport(Document doc)
         {
+            // Reset per publish. This is static state, so without it a run that picks an
+            // existing file would inherit the ANSWER FROM THE PREVIOUS RUN and build an
+            // element map that disagrees with the geometry — the same
+            // meshes-without-properties symptom, arriving by a different route and only
+            // on the second publish of a session, which is the worst kind to reproduce.
+            // Only ExportActiveView, which actually asks, may set it true.
+            _includeLinksThisRun = false;
+
             var dlg = new TaskDialog("Publish Model")
             {
                 MainInstruction = "How do you want to provide the 3D geometry?",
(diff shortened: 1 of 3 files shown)


StingD85:
Closing as already merged, not abandoned.

Every line of this PR is on `main` — it arrived inside #760, because `claude/model-delete-restore` was branched from this branch rather than from `main`, so the squash carried the federation commits with it. Verified marker by marker (`_docStack`, `CollectLinksRecursive`, `Include linked models`, `SkippedLinkCount`, `LinkScope`): branch and main agree on all of them.

It shows CONFLICTING only because both branches appended to the same spot in CHANGELOG.md.

My branching was the mistake — stacking a second piece of work on an unmerged branch meant #760's title now covers two unrelated changes. The federation fix is described in its own CHANGELOG entry (Phase 240), so the history is still readable.

PR #763, opened 2026-08-22
fix: the element map described the documents, not the model

StingD85:
The federation fix (#754) worked — the GLB went from **13 elements to 1,407**, six links
contributing ~7,000 elements each. Then the upload failed:

```
Upload failed: HTTP 400: {"error":"element_map_too_large","maxMb":5}
```

## The map was 12.28 MB describing 1,407 meshes

Measured on the generated file rather than guessed:

| | |
|---|---|
| entries | **37,110** |
| `Lines` | **29,521** |
| `Legend Components` | **5,706** |
| Walls / Pipes / Furniture / Columns / Windows / Doors / Toposolid | 181 / 133 / 130 / 106 / 87 / 67 / 57 |

**95% of the payload was legend and detail content** living inside the linked files'
non-3D views — annotation the viewer can never select.

**A category filter cannot fix this.** `OST_Lines` *is* a model category; these entries come
from views that are not the 3D view. Only the exporter knows what was actually drawn — so
it now says. It records the key of every element that produced at least one mesh, **under
the same condition that keeps the node**, so the two can never disagree about what is in
the file. The map is narrowed to that set: **37,110 → ~1,407**.

Only when the publish ran the exporter. A user-picked `.glb`/`.ifc` leaves the set null and
the map keeps its full scope — guessing what a file we did not produce contains would drop
real elements. Logged as **PUB-2**.

## Two smaller wins, both measured

- **Compact instead of indented**: 12.28 MB → 9.70 MB. A 21% saving on a machine-read
  sidecar nobody opens by hand, which counts against an upload limit.
- ~**2.40 MB** of the original was pure repetition of the absolute link path in every key —
  an artefact of #754's namespacing. Narrowing the set removes most of it.

## Server: the cap was binding on ordinary models

5 MB → **25 MB**, in one named constant rather than two magic numbers that had to agree.

At the ~265 bytes/element a real map costs, 5 MB is roughly **19,000 elements** — a
federated site passes that easily. So the limit was catching ordinary work, not just this
bug.

**Deliberately not higher.** `DownloadElementMap` reads the whole map into a string to
merge the cost sidecar, on a **512 MB** free-tier instance. Storing it gzipped is the
durable answer — **31× measured on this file, 12.28 MB → 0.40 MB** — and is logged as
**PUB-1** rather than bolted on here, because it needs the serve path *and* the cost merge
to decompress, plus a magic-byte check so plugins still sending plain JSON keep working.

## The error message

`{"error":"element_map_too_large","maxMb":5}` was true and unusable. It never said how
large the map *was*, so "slightly over" and "twenty times over" looked identical, and
nothing pointed at the actual cause. It now carries `actualMb` and a `hint`; `error` and
`maxMb` keep their names for anything already matching on them.

## Verification

- Plugin: **0 errors, 0 warnings**. Deployed, and the fix confirmed present in the shipped
  DLL by scanning **both byte alignments** with a must-find and a must-miss control — a
  single-alignment scan reported this exact class of change as missing earlier in the day.
- Server: 0 errors, **813 passing**.

Not yet re-run against the real model — that is the next thing to do, and the log now
prints the narrowed count so the result is visible without opening the file.

### DIFF
--- a/Planscape.Server/src/Planscape.API/Controllers/ModelsController.cs
+++ b/Planscape.Server/src/Planscape.API/Controllers/ModelsController.cs
@@ -58,13 +58,25 @@ public class ModelsController : ControllerBase
 
     // ── List / metadata ────────────────────────────────────────────────
 
+    /// <summary>
+    /// Models in the project. Live only by default.
+    ///
+    /// <para><paramref name="deleted"/> returns the soft-deleted ones instead, which is
+    /// what makes the 30-day grace in <c>ModelPurgeJob</c> reachable. Without a way to
+    /// SEE a deleted model there is no way to restore it, and the grace period protects
+    /// nobody — it just delays the bytes leaving.</para>
+    /// </summary>
     [HttpGet]
-    public async Task<ActionResult> List(Guid projectId, CancellationToken ct)
+    public async Task<ActionResult> List(Guid projectId, CancellationToken ct, [FromQuery] bool deleted = false)
     {
         if (!await ProjectInTenant(projectId, ct)) return Forbid();
-        var rows = await _db.ProjectModels.AsNoTracking()
-            .Where(m => m.ProjectId == projectId && m.DeletedAt == null)
-            .OrderByDescending(m => m.UploadedAt)
+        var q = _db.ProjectModels.AsNoTracking()
+            .Where(m => m.ProjectId == projectId);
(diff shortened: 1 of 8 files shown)


StingD85:
Superseded by a clean branch off `main`. This branch was cut from the delete/restore work, which was then squash-merged as #760, so it carried already-merged commits and went CONFLICTING. Reopened with only the element-map commit on top of current `main` — same change, no replayed history.

PR #812, opened 2026-09-06
Baseline layer 2: mint types inside families that already exist

StingD85:
## What this adds

Layer 1 (#789) creates host types outright. **Layer 2 mints a TYPE inside a family that is already loaded** — and it cannot conjure the family. With none loaded the entry is guidance, exactly as the family expectations report today. A baseline that claimed to create a door family would be inventing its own deliverable.

Same `STING_PROJECT_BASELINE.json`, same project override, same read-only `Baseline_Audit`, same single confirm. **No new command**, per spec §9.

## The rollback is the load-bearing part

`FamilySymbol.Duplicate` **commits before** the parameter set runs. A failed set would leave a baseline-**named** type carrying the *source* type's sizes — which the next audit reads as existing and refuses to touch, making the failure permanent and invisible.

That is **#798 exactly**, on five floor types. So: either the type is what the baseline describes, or it is not there at all.

## The mint is UNPROVEN, and is written as such

Nothing in this codebase duplicates a `FamilySymbol` — only `TextNoteType` and `DimensionType`. Every type is attempted and **reported individually**. Catalog-driven families will refuse to duplicate; that is expected, not exceptional, and is reported per type without crashing the run. **An existing vendor type is never edited** — the auditor only ever marks *absent* types Missing.

## Vendor families are in scope, and the prefix is what makes that safe

Per §8 D4 / §10.4, layer 2 mints into vendor families too. After a vendor reloads their family with "overwrite existing", the STING types are gone — and the audit/apply cycle is idempotent, so the next audit reports them Missing and Apply re-mints them.

The `"STING "` name prefix is the contract that makes a minted type identifiable as ours afterwards. `Validate()` rejects a type name without it — **case- and space-exact**, because a tool looking for minted types matches the literal string.

## Four decisions worth naming

| Decision | Why |
|---|---|
| A name match with **different** parameters is a **CONFLICT**, never an overwrite | The model's 800×2100 may be the deliberate one. Reported as *"exists at Width 800mm"* — what **is**, not what was wanted |
| **0.5 mm tolerance** on that comparison | Revit stores lengths in feet, so 900 mm round-trips as 899.9999999. Comparing exactly would report a conflict on every single type, and a report that cries wolf gets switched off |
| An **unreadable** parameter is not a difference | Treating an absent read-back as 0 mm would report a fabricated conflict at "0mm" on a type that is very likely correct |
| A parameter the family **does not have** is a per-type FAILURE with the parameter named | Not a silent skip. A type minted at the wrong size looks finished |

`familyTypes` **ships empty**. Sizes and naming need human sign-off (§8 D3, settled in §10 as opt-in packs); a corporate default would push a type list into every model that runs Apply.

## Verified

- `dotnet build StingTools/StingTools.csproj -c Release` (`-t:Rebuild`) → **0 errors, 0 warnings**
- `dotnet test StingTools.Boq.Tests` → **744 green**, up from 716
- `tools/check_command_doc_acquisition.ps1` → **PASS**
- `tools/check_path_discipline.ps1` → **PASS**
- `tools/check_workflow_wiring.ps1` → **PASS** (Tier 4: 0)

**Twenty-five deliberately wrong inputs, each shown to make the matching gate FAIL, then restored** — including the delete-the-thing-it-protects check on every new gate: the STING-prefix contract dropped · a category-less entry allowed through · an entry with no family patterns · a blank pattern counted as real · parameter name/value checks dropped · duplicates keyed on name only · the duplicate check removed · layer-2 validation never called · a missing type offered with no family loaded · the host search ignoring the category · the pattern list no longer ordered · a differing type silently present · an existing type offered for creation · the mm tolerance removed · the tolerance widened until a real difference passes · an unreadable parameter read as 0mm · the finding no longer naming the host family · the report no longer naming it · the group label drifting so the minter finds nothing · layer 2 never audited · a family type adopted into the corporate baseline · an **invalid** one adopted (which also fires the pre-existing `The_Shipped_Baseline_Validates`).

**Two results needed a second pass, and both were findings rather than passes:**

1. Relaxing the prefix comparison to `OrdinalIgnoreCase` moved **no test at all** — so `The_Prefix_Is_Case_And_Space_Exact` was added and re-verified firing.
2. The no-read-back test was guarding the **outer** early return, not the inner one. The correct mutation was written; it fires.

## NOT verified

**Nothing Revit-side has been run by anyone** — across T1–T5 and both baseline layers.

`MintFamilyTypes`, the `FamiliesByCategory` reader and `ReadDeclaredTypeParameters` have **never executed against a real model**. The audit half is Revit-free and fully tested; the mint is the unproven half and says so in its own header. Do not read this PR as evidence that minting a family type works.

https://claude.ai/code/session_01Ln3G9zFMNri66K13obgcW2

### DIFF
--- a/StingTools.Boq.Tests/BaselineFamilyTypeTests.cs
+++ b/StingTools.Boq.Tests/BaselineFamilyTypeTests.cs
@@ -1,0 +1,389 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+using StingTools.Core.Baseline;
+using Xunit;
+
+namespace StingTools.Boq.Tests
+{
+    /// <summary>
+    /// BASELINE LAYER 2 — family types.
+    ///
+    /// Layer 1 creates host types outright. Layer 2 mints a TYPE inside a
+    /// family that is already loaded, and cannot conjure the family: with none
+    /// loaded the entry is guidance, exactly as the family expectations report
+    /// today. A baseline that claimed to create a door family would be
+    /// inventing its own deliverable.
+    ///
+    /// The audit half is Revit-free and is what these pin. The MINT is
+    /// Revit-bound and has never run — nothing in this codebase duplicates a
+    /// FamilySymbol.
+    /// </summary>
+    public class BaselineFamilyTypeAuditTests
+    {
+        private static BaselineFamilyType Door(string typeName = "STING Flush Door 900x2100") =>
+            new BaselineFamilyType
+            {
+                Category = "Doors",
+                FamilyNamePatterns = new List<string> { "M_Single-Flush", "Single-Flush", "Door" },
+                TypeName = typeName,
+                Purpose = "Standard internal single leaf",
+                Parameters = new List<BaselineTypeParameter>
+                {
+                    new BaselineTypeParameter { Name = "Width", ValueMm = 900 },
+                    new BaselineTypeParameter { Name = "Height", ValueMm = 2100 },
+                }
+            };
+
+        private static ProjectBaseline With(params BaselineFamilyType[] types) =>
+            new ProjectBaseline { FamilyTypes = types.ToList() };
+
+        private static ModelInventory Model(string family = "M_Single-Flush",
+                                            string existingType = null,
+                                            Dictionary<string, double> existingParams = null)
+        {
+            var m = new ModelInventory();
+            if (family != null)
+                m.FamiliesByCategory["Doors"] = new List<string> { family };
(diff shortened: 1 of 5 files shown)


StingD85:
Superseded: B1 landed in parallel as #811, and B2 as #813, while this was in review. Closing rather than resolving the conflict — the same work twice is not worth reconciling, and main's version is the one everything else now builds on.

Continuing with **B3 (catalogue packs)** and **B4 (harvest)** on top of #811/#813, per §10.2 and §10.3 of the spec.

PR #816, opened 2026-09-06
B3: catalogue packs nobody has adopted, and the merge that made layers 2 and 3 reachable

StingD85:
## A defect found on the way in

`BaselineRegistry.Load` merged materials, host types, levels and family *expectations* from the project override — but **not `familyTypes`, `familyParameters` or `adoptCatalogues`**.

The corporate file ships all three empty, so the project override was the **only possible source** for layers 2 and 3, and it was being dropped on the floor. **Both layers shipped unreachable**: a project declaring `familyTypes` got a clean audit and no explanation.

Fixed here, because B3's adoption mechanism could not have worked either.

## Why not a fixed corporate catalogue

Shipping ~30 East African types in the universal baseline would be **one person's reading of the market**, applied to every future project and reported by the audit as though it were a standard. A project whose doors are genuinely 850 wide would be told, on every run, that it is missing a type it does not want.

That is the familiar failure: a confident default nobody asked for, that nobody checks.

## The mechanism: named packs, none active by default

`STING_TYPE_CATALOGUES.json`. A project adopts by id in its own override:

```json
{ "adoptCatalogues": ["EA-RESIDENTIAL-V1"] }
```

| Property | Why |
|---|---|
| **Adoption is a stated decision** | per project, and reported in the audit |
| **Additive, and the project wins** | adopt a pack and still override individual types; overrides are counted and **named**, not silent |
| **A new market is a new pack** | not a code change, not a change to the universal baseline |
| **Packs are versioned in the id** | `-V1` → `-V2` is an explicit migration, never a silent redefinition of what a project already adopted |
| **An unknown adopted id is NAMED** | a project that believes it adopted a catalogue and did not is exactly the confident absence this codebase keeps paying for |
| **Pack types obey the layer-2 rules** | by borrowing `ProjectBaseline.Validate` rather than restating them — which is what keeps the two from drifting |

## One pack ships, and it is adopted by nobody

`EA-RESIDENTIAL-V1`, status `provisional`. Its `sourceNote` says the sizes are seeded from common Ugandan residential practice, **not** from a standard and **not** from a completed project; that it is a *shape* rather than a recommendation; and that it should be replaced by harvesting a delivered model (B4).

Validation **requires** a source note on any provisional pack — one that does not say why it is provisional reads exactly like a standard. A reviewed pack does not need one, so the rule is about the *claim*, not paperwork.

**A provisional pack nobody has adopted can be wrong without costing anything. A corporate default cannot.**

Two tests hold the corporate file to that: it adopts nothing, and it declares no `familyTypes` inline either — belt and braces on the same decision.

## Verified

- `dotnet build StingTools/StingTools.csproj -c Release` (`-t:Rebuild`) → **0 errors, 0 warnings**
- `dotnet test StingTools.Boq.Tests` → **776 green**, up from 772 on top of #811/#813
- `tools/check_command_doc_acquisition.ps1` → **PASS**
- `tools/check_path_discipline.ps1` → **PASS**
- `tools/check_workflow_wiring.ps1` → **PASS** (Tier 4: 0)

**Twenty-seven deliberately wrong inputs, each shown to make the matching gate FAIL, then restored** — including the delete-the-thing-it-protects check on every new gate: packs applying unadopted · an unknown id ignored · the pack overwriting a project declaration · the override match ignoring the category · an invented "0 packs" finding · six ways of weakening the summary and the validation · the corporate baseline adopting a pack · or declaring the types inline · the shipped pack claiming to be `reviewed` · its source note losing its disclaimer · its id losing its version suffix · a pack type losing the STING prefix or its size · windows dropped from the pack · a pack type that would break the baseline it joins · adoption accumulating instead of replacing.

**One test failed on its own premise rather than on the code:** the family name `M_Single-Flush` does not contain the pattern `Door`, so the audit correctly reported guidance. The **test** was fixed, not the assertion loosened.

## NOT verified

**Nothing Revit-side has been run by anyone** — across T1–T5 and all three baseline layers.

The adoption resolver and library validation are Revit-free and fully tested. `LoadResolved` and the override merge read files through `StingPaths` and have **never executed against a real model**. In particular, **the claim that layers 2 and 3 are now reachable is untested in Revit** — it is a code-reading finding with a repaired merge behind it, not an observed export.

## Note on #812

B1 and B2 landed in parallel as #811 and #813 while my own versions of them were in review. I closed #812 rather than reconciling duplicate work; this PR builds on main's B1/B2.

https://claude.ai/code/session_01Ln3G9zFMNri66K13obgcW2

### DIFF
--- a/StingTools.Boq.Tests/StingTools.Boq.Tests.csproj
+++ b/StingTools.Boq.Tests/StingTools.Boq.Tests.csproj
@@ -60,6 +60,9 @@
     <Compile Include="..\StingTools\Core\MaterialSchedule\RoofAccessoryTally.cs" Link="MaterialSchedule\RoofAccessoryTally.cs" />
     <Compile Include="..\StingTools\Core\Baseline\ProjectBaselineModel.cs" Link="Baseline\ProjectBaselineModel.cs" />
     <Compile Include="..\StingTools\Core\Baseline\BaselineAudit.cs" Link="Baseline\BaselineAudit.cs" />
+    <!-- B3 - opt-in catalogue packs. Adoption decides what reaches a model,
+         so it is Revit-free and tested. -->
+    <Compile Include="..\StingTools\Core\Baseline\TypeCatalogue.cs" Link="Baseline\TypeCatalogue.cs" />
     <Compile Include="..\StingTools\Core\Baseline\BaselineAugmentReport.cs" Link="Baseline\BaselineAugmentReport.cs" />
     <!-- The XLSX writer is Revit-free (ClosedXML only), so the DELIVERABLE itself
          is testable, not just the model behind it. -->
@@ -99,6 +102,9 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
     <None Include="..\StingTools\Data\STING_MATERIAL_STAGES.json" Link="Data\STING_MATERIAL_STAGES.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="..\StingTools\Data\STING_TYPE_CATALOGUES.json" Link="Data\STING_TYPE_CATALOGUES.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
     <None Include="..\StingTools\Data\STING_PROJECT_BASELINE.json" Link="Data\STING_PROJECT_BASELINE.json">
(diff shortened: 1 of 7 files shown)


StingD85:
Superseded: B3 landed in parallel as #814 and B4 as #815 while this was in review.

Worth noting that #814 independently made the same finding and the same fix I had here — `BaselineRegistry.Load` was not merging `familyTypes`, `familyParameters` or `adoptCatalogues` from the project override, which left layers 2 and 3 unreachable. Two sessions arriving at that separately is reassurance about the finding, not a reason to land it twice.

Closing. Verifying main's B1–B4 against the task's invariants instead.

PR #847, opened 2026-09-08
fix(kut-11): GetFamilyName answers for system elements, unblocking 11 dead PROD rules

StingD85:
Closes **KUT-11**.

## What was wrong

`ParameterHelpers.GetFamilyName` returned `""` for anything that is not a
`FamilyInstance` — so walls, floors, roofs, pipes, ducts, cable trays, conduits and
wall footings had no family name, and every consumer keyed on one silently skipped them.

**The ROADMAP row located the harm in the CSI map. The measurable harm was in PROD.**
`ProdResolver.Resolve` gates its *entire* project + corporate rule lookup on a non-empty
family name:

```csharp
if (!string.IsNullOrEmpty(familyName)) { /* project → corporate → LPS → sleeve */ }
// otherwise: straight to the category default
```

So the **eleven corporate PROD rules shipped on Ducts, Floors and Structural Foundations**
had never once fired — `*Concrete Slab*`→`SLB`, `*Post-Tension*`→`PSB`, `*Composite*`→`CFS`,
`*Fire Damper*`→`FSD`, `*Pile*`→`PIL`, `*Raft*`→`RAF` and five more. Every floor, duct and
wall footing took its generic category default instead. The data was right; the plumbing
dropped it, and nothing said so.

**Fixing the family side alone would not have reached them.** Those patterns match the
*type* name, and `GetFamilySymbolName` was empty for exactly the same elements.

## What changed

| | |
|---|---|
| `GetFamilyName` | Falls back to `ElementType.FamilyName` → the **system** family (`"Basic Wall"`, `"Rectangular Duct"`, `"Wall Foundation"`). For a `FamilySymbol` that property *is* the loadable family's name, so loadable families are bit-identical. |
| `GetElementTypeName` | **Promoted out of `CsiMap.TypeName`**, which held the only copy of this fallback and benefited only the CSI resolver. `CsiMap.TypeName` now delegates. |
| `GetLoadableFamilyName` | The old behaviour, under a name that says what it means. |
| `GetFamilyAwareProdCodeCore` | Reads both, so the eleven rules can fire. |

`GetFamilySymbolName` itself is **untouched** — 55 callers, each of which can opt in on its
own evidence rather than being moved by a side effect.

### Callers that depended on `""`, now saying so

- **`Prod_GenerateRules`** writes `FAMILY_PATTERN` rows into the project overlay. A row
  keyed on `"Basic Wall"` matches *every wall in the model* — a category default wearing a
  family rule's clothes. Pinned to `GetLoadableFamilyName`.
- **Tag-map import** matched on `fam.Equals(family)`. A map exported *before* this change
  carries `"family": ""` for every system element while this side now computes `"Basic Wall"`,
  so those elements would have fallen to `skipped`. An unrecorded family is now treated as
  "no opinion"; type + the 500 mm radius still discriminate.

### A divergence found on the way

`BOQCostManager` was passing its **local** helper's answer — which falls back to the *type*
name — into `CsiMasterFormat.Resolve`'s **family** slot, while `CsiAssign` passed `""` for the
same element. One map, two paths, two answers; a `FamilyRegex` row on a system category would
have matched a *type* name down one of them. That call now uses the shared helper. The two
private helpers stay and now document why they differ: a bill and a paragraph want
`"Strip Footing 600x300"`, a rule wants `"Wall Foundation"`.

### Data

The CSV header asserted in prose that a `FamilyRegex` on a system category "can NEVER fire".
That is no longer true. The header was rewritten, and the audit behind it made executable.

## Tests — `StingTools.Tags.Tests` 307 → 311

- The eleven system-category PROD rules resolve **with** a family name and fall through to
  the category default **without** one — driven off the shipped CSV, so a deleted row fails
  here instead of going quietly dead.
- `"Concrete Slab 200"` floor → `SLB`.
- **No shipped CSI `FamilyRegex` row sits on a system category** — the invariant that makes
  this change provably inert for CSI. There are none today, so nothing moves; if someone adds
  one, this fails rather than the map silently re-classifying a project.
- That guard fed the exact row it exists to catch.

| Sabotage | Failed |
|---|---|
| Deleted the shipped `SLB,Floors,*Concrete Slab*` row | `ConcreteSlabFloor_ResolvesToSlb` |
| Appended `Walls,(?i)concrete,…` to the shipped CSI map | `NoShippedCsiFamilyRegexRowSitsOnASystemCategory` |
| Removed `ProdResolver`'s empty-family gate | `SystemCategoryProdRules_WereUnreachable_AndNowResolve` |

## Verification

```
StingTools.Tags.Tests                                311 passed, 0 failed
StingTools.Boq.Tests                                 1128 passed, 0 failed
```

**No Revit runtime path is exercisable from a terminal, and none was exercised.** What
`ElementType.FamilyName` returns per category rests on the API contract, not on a run — the
tests assert that *once* a system family name and a type name arrive, the shipped rules
resolve.

## Behaviour change worth knowing

A newly-tagged floor, duct or wall footing will now receive a *specific* PROD code where it
used to receive the category default. Existing tags are unaffected: token population is
`SetIfEmpty` unless a caller passes `overwrite`.


StingD85:

| PR | Adds |
|---|---|

   and needs no change.

PR #854, opened 2026-09-08
feat(kut-10): key the structural CSI codes on material, per MasterFormat 2020

StingD85:
Closes **KUT-10**. Reports on **LIFE-2** (which stays open).

**Owner decision (§9):** follow a well-known standard. That standard is **CSI MasterFormat
2020**, and it resolves the judgement rather than deferring it.

## Why a category-keyed default was a judgement call

MasterFormat divides by **work result and material**, not by authoring-tool category:

| | |
|---|---|
| Division 03 | Concrete — 03 30 00 cast-in-place, 03 41 00 precast |
| Division 04 | Masonry — 04 20 00 unit masonry |
| Division 05 | Metals — 05 12 00 structural steel, 05 21 00 steel joists |
| Division 06 | Wood — 06 11 00 framing, 06 17 00 shop-fabricated structural |

A Revit `Structural Framing` element can be any of those.

## What landed

`CsiRule` gains a 9th **`Material`** column, read from `STRUCTURAL_MATERIAL_PARAM` on the
instance then the type — **the declared parameter, never the type name**, because
`"W310x39"` is a naming convention, not a fact about what a member is made of.

`PrimaryMaterial.Resolve` was **deliberately not reused.** It leads with
dominant-by-volume, which is exactly right for *pricing* a compound assembly and wrong for
*classifying* a member: on a clad element the largest volume can be a finish, and Division
05 steelwork would bill as Division 07. Two questions, two resolvers.

### The scoring is where the real work was

Making material simply outrank everything breaks MasterFormat's own hierarchy — **a bored
pile is made of concrete and is still measured under 31 63 00, not 03 30 00.** So
qualifiers score 2 and a matched material 3:

```
category alone                 2   the fallback
category + family/type         4   a name guess
category + material            5   a DECLARED material beats the guess
category + family + material   7   a named WORK RESULT beats material alone
```

### Two shipped-map defects the tests surfaced — both pre-existing

1. `"Masonry - Concrete Blockwork"` matches **both** the masonry and concrete patterns and
   tied on score. Masonry now leads: a concrete block is Division 04.
2. `(?i)pile` was listed **before** `(?i)shoring|underpin|sheet pile`, and both match
   "sheet pile". **A sheet pile already resolved to 31 62 00 Driven Piles instead of
   31 40 00 Shoring.** The reorder is a fix, not a side effect of this change.

`Structural Foundations` deliberately gets only **two** material rows: a concrete row would
restate the bare-category fallback (same section, same NRM2), and a steel row would
outscore sheet-pile shoring for no gain, since a steel "foundation" there is a base plate.
That leaves precast as the only material row able to outrank a division-31 work result, so
only the driven-pile row needed a twin.

### The header

"Common approximations to be confirmed against the project's RIB SpecLink spec" is retired.
The surviving category rows are labelled **fallback, not answer**: a project whose members
carry no structural material is being classified on the weakest available evidence, and the
fix is to set the parameter, not to re-tune the fallback.

## Backward compatibility

The row widens 8 → 9 columns; the parser still accepts **6, 7, 8 or 9**. `ParseCsvLines`
drops a short row *silently*, so demanding nine would take every project's
`_BIM_COORD/csi_map.csv` offline with no error anywhere. A test loads an overlay of each
width and asserts all four still **match**, not merely that they parse.

## Tests — `StingTools.Tags.Tests` +24

Driven off the shipped map, so a deleted or re-weighted row fails here instead of quietly
re-billing a project.

| Sabotage | Failed |
|---|---|
| `MaterialWeight` lowered so a declared material stops beating a name guess | `ADeclaredMaterialBeatsAGuessFromTheTypeName`, `TheScoreHierarchyIs…` |
| Driven-pile work-result twin deleted | `ANamedWorkResultBeatsTheMaterial` |
| Parser tightened to demand 9 columns | 6 material-resolution cases |
| NRM2 blanked on a material row | `EveryShippedMaterialRowCarriesAnNrm2Code` |

One Boq test pinned the absolute score `1` and now asserts `CsiRule.QualifierWeight`. Every
production caller discards the score (`out _`) and reads only the tie count, so the scaling
is inert — the test was pinning the wrong thing.

## LIFE-2 — reported, not closed

**This does not reduce the 36.** It **adds one** division-31 row (the precast driven-pile
twin), so the QS review sheet goes **36 → 37**. Narrowing Foundations to two material rows
is what kept that from being **39**.

Which LIFE-2 rows the material key reaches. The 37 rows break down as **9 in division 31,
17 in division 32, 11 in division 33**. It reaches **none of the 28 division 32/33 rows** —
landscaping, paving, site utilities and drainage, all work results whose NRM2 section
follows what the work *is* rather than what it is made of, so a material key cannot narrow
them. Within division 31 it touches 4 of the 9 (bored piles, driven piles ×2 including the
new twin, shoring); all four already carried a code and still carry the same one. The
remaining 5 — Topography, Toposolid ×2, Pads, erosion control — are earthworks, again keyed
on work result. **So the answer to §10's question is: no, it reduces nothing.**

**No NRM2 code was authored or altered.** `check_qs_nrm2_review.py` had hardcoded
`'36 still unreviewed'` and now derives the count from the sheet it just wrote, so the next
map change cannot fail a check that says nothing about whether the round trip works — that
derivation was broken once and confirmed failing.

## Verification

```
StingTools.Tags.Tests                                331 passed, 0 failed
StingTools.Boq.Tests                                 1133 passed, 0 failed
```

**No Revit runtime path is exercisable from a terminal, and none was exercised.**
`STRUCTURAL_MATERIAL_PARAM` cannot be read outside a Revit host; what is asserted is that
once a material name arrives, the shipped map resolves it to the MasterFormat division it
belongs to.


StingD85:

| PR | Adds |
|---|---|

   and needs no change.

PR #856, opened 2026-09-08
feat(kut-5): Division 02 keyed on phase, not on naming

StingD85:
Closes **KUT-5**.

Phase state now reaches `CsiMasterFormat.Resolve`, and Division 02 Existing Conditions is
**one rule** keyed on it: *any element the works remove is demolition, whatever it is made
of.*

The withdrawn first pass keyed Division 02 on naming — and **nobody names a toposolid
"demolition"**, so those rows could not fire on a real model. Revit expresses the state
through `Phase Created` / `Phase Demolished`, so that is what this reads.
`ElementPhaseState` is Revit-free (hence testable); `CsiMap.PhaseState` only turns two
`ElementId`s into positions in the document's phase sequence.

## The weighting is the load-bearing decision

A matched phase scores **10** — more than every product qualifier combined — because
MasterFormat Division 02 classifies by the **state of the work**, not by what the thing is.
A demolished precast beam is demolition: the contractor is removing it, not casting it.

At any lower weight the rule would fire only on elements no other rule matched — which is
the withdrawn naming rows' failure in a new costume. There is a test asserting the
inequality directly.

## Three traps encoded rather than discovered later

1. **A single-phase project has no existing elements.** That phase *is* the works.
   Classifying it `Existing` would silently mark a whole building out of scope and leave a
   bill empty for a reason nobody would look for.
2. **A non-phase-aware element is `Unknown`, not `New`.** `Unknown` is the empty string,
   which `Score` already treats as no-match — so a grid can never satisfy a phase rule.
3. **`Existing` and `Temporary` are deliberately not classified.** Existing-to-remain is not
   work, and a rule would assert a bill item for something nobody is being paid to touch;
   temporary works are NRM2 preliminaries and a QS's judgement. Both states are now
   *matchable*, so a project can add its own rule.

## NRM2 1 = Demolitions — read, not guessed

Taken from **this repository's own vocabulary** in `BOQCostManager.GuessSectionName`, not
from published NRM2. The in-house scheme has `4` Foundations and `5` In-situ concrete where
published NRM2 has `5` Excavating and filling. Guessing across that gap moves money.

## The parser hazard §7 names

The row widens to 10 columns; **6, 7, 8 and 9 all still load *and still match*** — asserted,
not assumed, because `ParseCsvLines` drops a short row *silently*.

**Column 9 is RESERVED** for KUT-10's `Material` qualifier (PR #854), so the two changes
compose under either merge order. Putting both at column 9 would make one branch's data
silently mean the other's.

## §7 said "re-run the resolution harness". There was no harness.

The map's reachability had never been measured. One is now committed: **51 categories × 5
phase states = 255 probes** against the shipped map, through the real resolver.

| | Probes landing in Division 02 |
|---|---|
| before | **0** of 255 |
| after | **51** of 255 |

### It immediately found a pre-existing gap — baselined, not papered over

**`Generic Models` and `Pipes` have no bare-category fallback.** Every rule on them carries
a family or SYS qualifier, so an unqualified element in either **resolves to nothing and its
CSI section is left blank.**

Not fixed here, deliberately: authoring a catch-all would put a confident CSI number on
exactly the elements nobody has classified — "Generic Models" is Revit's phrase for *we did
not say what this is* — and for a pipe with no SYS token, 22 11 00 and 23 21 00 are both
plausible and only the service decides. The set is **baselined so it cannot grow silently**,
and shrinking it means deleting the entry in the same commit that fixes it.

## The rule is appended, not inserted

`qs_nrm2_review.py` keys its review sheet on a row's **ordinal** among the data rows. Adding
a rule higher up re-points every `ROW` below it and invalidates an in-progress QS review —
the staleness guard catches it (it refuses rather than writing to whatever moved into that
slot), but the whole sheet must be re-exported. I hit this: the first placement moved the
sheet's first row from 109 to 94 and broke 7 of the QS checks.

Position is irrelevant to matching here, since `PhaseWeight` decides on score and row order
only ever breaks a tie. **The 36 review rows and their ordinals are unchanged.**

## Tests — `StingTools.Tags.Tests` +27

| Sabotage | Failed |
|---|---|
| `PhaseWeight` lowered to 1 | 6 × `ADemolishedElementIsDivision02WhateverItIs` |
| Division 02 row deleted | same |
| Parser tightened to demand 10 columns | same |
| Single-phase elements made `Existing` | `ASinglePhaseProjectHasNoExistingElements` |
| Non-phase-aware element made `New` | `NotPhaseAwareIsUnknownNotNew` |

## Verification

```
StingTools.Tags.Tests                                334 passed, 0 failed
StingTools.Boq.Tests                                 1165 passed, 0 failed
tools/check_qs_nrm2_review.py                        ALL CHECKS PASSED (36 rows, ordinals unchanged)
```

**No Revit runtime path is exercisable from a terminal, and none was exercised.**
`CreatedPhaseId` / `DemolishedPhaseId` cannot be read outside a Revit host; what is asserted
is that once a phase state arrives, the shipped map classifies it correctly.


StingD85:

| PR | Adds |
|---|---|

   and needs no change.

PR #860, opened 2026-09-08
fix(life-6): regenerate all three binding outputs together, and gate the drift

StingD85:
Closes **LIFE-6**.

## Re-measured first, as §11 asked

At `c6a4bbe9a` the three outputs of one generator run were **256 / 0 / 526** lines stale
(audit view / gaps / deployable). The row's 245 / 0 / 520 had moved — which is exactly why it
said to re-measure. All three are now regenerated and committed **together**.

### 1. `SHT_*` — the exclusion was the bug

The resolver listed `SHT` among the excluded prefixes, while the shipped deployable bound all
ten `SHT_*` parameters to `Sheets`. Regenerating would have dropped them, and **absence from
that file means *intentionally unbound***.

`Sheets` is a real bindable category (`CategoryEnumMap` → `OST_Sheets`), and the ten
parameters are sheet metadata with nowhere else to live. It is now a `SHEET` domain. **All ten
survive.**

### 2. Materials — the row's premise was wrong

The row read the loader's `Materials` discard as a regression. **It is by design.** Material
binding is done by a different mechanism entirely:
`LoadSharedParamsCommand.IsMaterialRelevantParam` picks parameters **by name prefix** and
`CleanMaterialBindings` binds them. `SharedParamGuids` drops the pseudo-category precisely
because that other path owns it.

All **108** `Materials` rows are recognised by that rule, so **nothing goes dark**.

What the resolver gained instead is a **generation-time gate**: it reads the prefixes **out of
the C# source** — a second copy of that list is the defect class this file exists to remove —
and **fails** rather than write a `Materials` row the C# would not bind. That is the exact
failure the code's own comment records for `BLE_MATERIAL_TXT`.

## The "~22 regressions" did not materialise

```
0 rows lost      218 gained      3,025 -> 3,243 bindings
```

Of 138 changed rows: **110 widened** to universal, 28 re-scoped. Of the **14 that narrow, 13
agree exactly with the hand-authored `CATEGORY_BINDINGS.csv`.** The regeneration is correcting
a stale deployable *toward* the authored sources, not regressing a good one.

The seven `MAT_*`/`COMP_MAT_*` rows the row named are among those 13:
`CATEGORY_BINDINGS.csv`, `FAMILY_PARAMETER_BINDINGS.csv` **and the parameters' own
descriptions** all say `Materials`. Only the stale deployable disagreed.

### One genuine outlier, reported rather than buried

**`MAT_TAG_6_TXT`** narrows from 37 categories to `Materials`, disagreeing with
`CATEGORY_BINDINGS.csv`. Its four siblings `MAT_TAG_2..5_TXT` are all `Materials`-only there,
so the 37-category row looks like the data error rather than the rule. It remains bound (by
the material pass); it is no longer on 37 element categories. Flagged for whoever owns the
parameter rather than silently preserved *or* silently changed.

## §11's question: should the audit view be committed at all?

**Keep it, and gate it.** It is the file a reviewer reads; deleting it trades a drift problem
for a discoverability one.

`--check` closes the drift instead — it regenerates in memory, writes nothing, and exits 1 on
any difference in **any of the three**. Three ages from one generator becomes a failure rather
than a discovery months later. It is CI-gate ready.

## Assertions, on the row's own terms

| | |
|---|---|
| Deployable carries at least as many bindings | **3,243 ≥ 3,025** ✅ |
| `MR_PARAMETERS.txt` non-TEXT parameters ≥ 763 | **763** ✅ |
| All ten `SHT_*` rows present | **10** ✅ |
| Rows lost | **0** ✅ |

| Sabotage | Result |
|---|---|
| Hand-edited one row (`SHT_DISC_TXT,Sheets` → `Walls`) | `--check` exits **1** — it compares content, not just line count |
| `SHT` put back on the excluded list | all ten `SHT_*` rows **vanish** |

## Verification

```
StingTools.Tags.Tests                                307 passed, 0 failed
StingTools.Boq.Tests                                 1177 passed, 0 failed
tools/param_binding_resolver.py --check              all three outputs current
```

`StingTools.Tags.Tests` asserts the COBie field map **against this very file**, so its 307
passes are direct evidence the regenerated spec is still coherent.

**No Revit runtime path is exercisable from a terminal, and none was exercised.** Whether a
binding actually lands in a model is a Revit operation; what is asserted here is that the spec
says what the authored sources say.


StingD85:
**Left open deliberately — not wholly superseded.**

Checked against `main` @ `7a5fcf05d`. The LIFE-6 work this PR was written to do **is** superseded by #877:

- `docs/ROADMAP.md:22` — the LIFE-6 row is struck through and marked **CLOSED 2026-09-08**.
- `.github/workflows/binding-spec-drift.yml` exists and re-runs the resolver on every touch.
- `python tools/param_binding_resolver.py` on `main` produces **no diff** across all three outputs.

So the drift-gating half is done, and merging this for that reason would be redundant.

**But this branch is not empty of value, and merging it would regress `main`.**

It carries a **Materials cross-check** `main` does not have: a `Materials` row in `RESOLVED_BINDINGS.csv` is honoured only if `LoadSharedParamsCommand.IsMaterialRelevantParam` also recognises the parameter, and one it does not sits in the spec looking bound while binding to nothing. That is worth having.

It would also, however, undo three things:

| this branch | `main` |
|---|---|
| `csv.writer(..., lineterminator="\r\n")` | `LF = "\n"`, pinned on purpose — the drift gate regenerates on Linux, so a CRLF file fails CI on a tree nobody edited |
| `import os,re,csv,collections,io,sys,sys` | no duplicated import |
| `return "SHEET","sheet-metadata"` | `return "SHEET","sheet"` — relabelling rewrites 10 rows of a generated file for no reason |

It is also 48 commits behind with 3 conflicts, on exactly the files #877 changed.

**Action taken:** the cross-check has been re-derived against today's `main` as **#901**, without the regressions, and sabotage-verified in four directions. Once #901 merges, this PR has nothing left that `main` lacks and can be closed.

Measured today: the cross-check finds **0 orphans across 108 Materials rows** — it is a guard, not a repair.

StingD85:
**Closing. Superseded by #901, and merging this would still regress `main`.**

Re-verified against `origin/main` @ `25f1448c7`, not carried over from the earlier note.

### The one idea worth having has landed

This PR's genuinely useful contribution was the **Materials cross-check** — fail the resolver when a row resolves to `Materials` but `LoadSharedParamsCommand.IsMaterialRelevantParam` would not bind it, so the spec cannot claim a binding nothing delivers.

That is on `main` now, re-derived as **#901** (`tools/param_binding_resolver.py:155` `material_prefixes()`, `mat_orphans` guard). It reads the prefixes out of the C# source rather than copying them, and it is anchored on `IsMaterialRelevantParam(` — a bare-name `find()` prefix-matched a renamed method, which sabotage caught.

### The three regressions are all still real — I re-checked each

| this PR | `main` |
|---|---|
| `csv.writer(buf, lineterminator="\r\n")` (line 198) | `LF = "\n"` (line 6), used on all three writers |
| `import os,re,csv,collections,io,sys,sys` (line 1) | no duplicated import |
| `if pre == "SHT": return "SHEET","sheet-metadata"` (line 73) | `return "SHEET","sheet"` |

The CRLF one is not cosmetic: `.gitattributes:35-37` pins all three outputs `text eol=lf`, and the drift gate regenerates on **ubuntu-latest** and compares bytes. This PR's writer would fail CI on a tree nobody had edited.

The `sheet-metadata` relabel rewrites 10 rows of a generated file to say the same thing differently.

### The one thing #901 does not have — and why it is not a reason to keep this open

This PR also adds a `--check` mode: regenerate in memory, write nothing, exit 1 on any difference. `main` has no such flag.

It is not a gap. `.github/workflows/binding-spec-drift.yml` already regenerates and fails on drift, and its header sets out that design deliberately. A second way to ask the same question is the "one fact one place" defect this repo keeps closing, not a feature — and the only advantage over the gate is not mutating a dirty working tree locally, which does not justify carrying a conflicting PR.

If a local dry-run is later wanted, it should be added on top of `main` in ~15 lines, not recovered by resolving 3 conflicts on a branch 48 commits behind on exactly the files #877 and #901 changed.

Nothing here is being lost. Closing.

PR #907, opened 2026-09-09
One timber vocabulary for three consumers, and the matcher swap that needed stems (W3, W5c)

StingD85:
W3 and W5c from `HANDOVER_MATERIAL_ALIGNMENT.md`, together because they are the same region of `TypeRenamePlanner`. Branched from `origin/main` (`7a5fcf05d`), independent of #904, #905 and #906.

## W3 — four word-lists, one carbon figure

`MaterialClassPlanner` knew fourteen timber words, `BiogenicCarbon` nine, `TypeRenamePlanner` four, and `UgCarbonFactors` read `Material.MaterialClass` — the controlled field, and therefore the only one right by construction.

Measured over the **1,808 distinct names** in the shipped register plus the delivered-model corpus, the three word-lists gave **different answers for 85** of them, including the two the brief named:

```
MDF CEILING PANEL 12MM   class Wood · biogenic credit YES · not timber to the renamer
SOLID BAMBOO 14MM        class Wood · biogenic credit NO  · not timber to the renamer
```

`SubstanceVocabulary.Wood` is now the one list and all three read it.

## What moved on a carbon number — all eighteen named

**Three names LOSE a credit, and all three should never have had one.** `BiogenicCarbon` matched by **substring** and carried `"ply "` with a trailing space. Across all 1,808 names that needle matched **exactly three**:

```
PVC SINGLE PLY 1.5MM              a PVC roofing membrane
SUPPLY REGISTER 150X150MM 1-WAY   an air diffuser   ← sup-PLY
SUPPLY REGISTER 200X100MM 2-WAY   an air diffuser
```

Two air diffusers and a plastic membrane were earning a biogenic carbon credit. Zero true positives in the entire corpus, so `ply` is **dropped** rather than made whole-word — `plywood` covers what it was reaching for.

**Fifteen names GAIN a credit**, all listed by name in `The_Fifteen_Names_That_GAIN_A_Biogenic_Credit_Are_Listed_By_Name`:

> BAMBOO PLANK · CHEVRON PARQUET 15MM · CORK TILE · CORK TILE 6MM · Cork - Plastic · FLOOR LAMINATE OAK-DARK / -GREY / -LIGHT / -MEDIUM · HDF CORE · HERRINGBONE PARQUET 15MM · MAPLE FLOORING · Oak Flooring · Roca - TENET - 402 City Oak · SOLID BAMBOO 14MM

**Two of those fifteen are questionable, and the test says so rather than burying them in the count.** `Roca - TENET - 402 City Oak` is a sanitaryware colour, not a timber; `Cork - Plastic` is a cork-look plastic. Both are the colour-word collision already documented in `MaterialClassPlanner`'s header, and both are library appearances rather than modelled quantities — but a carbon term moved on them and you should know.

Whole-word matching also costs suffix forms, so the two the data actually contains are listed: `wooden` (`WOODEN SPORTS FLOOR 22MM`) and `barnwood` (`BARNWOOD SIDING 22MM`) both lost their credit without them, and both are real timber.

## The gate is about the NAME, not about the class table

`A_Name_With_Two_Substances_Is_Still_Decided_By_ORDER_Not_By_This_List` pins that `VINYL LVT WOOD OAK-DARK` and `Cork - Plastic` are still **Plastic**. "Timber for one is timber for all" is a lexical question; what an *ordered* class table does with a name carrying two substance words is precedence, and that stays where it is. Only the vocabulary is shared.

## W5c — the matcher swap needed stems, not just a matcher

`TypeRenamePlanner.Match` was the third `IndexOf` site — the #863 shape, in a file that reaches a PROD code. (**Correction to the brief:** it is `TypeRenamePlanner.cs:337` on `main`, not `:446`; the file is 341 lines.)

Swapping it for whole-word matching **alone breaks seven rows** of the 2026-09-09 plan, because the data is plural and the needles were singular:

| | rows | effect |
|---|---:|---|
| walls cored `Concrete Masonry Units` | 5 | Blockwork → RC — a block wall priced and carbon-counted as in-situ reinforced concrete |
| roofs cored `CLAY TILES PREMIUM 15MM` | 2 | `Clay Tile Roof` → no name at all |

Shipped with `concrete masonry units`, `masonry units` and `clay tiles` beside their singulars, plus `cobblestone` / `bluestone` / `flagstone` for Stone and `tiles` / `plastered` / `rendered` / `painted` for the Finish table — measured, not guessed: those are the forms the 1,808 names actually use.

Sharing the wood list also makes the renamer name **seven** core materials it used to refuse. Every one is literally a `Structure, Wood Joist/Rafter Layer`, so this is the table catching up with the model — but it is a behaviour change and `The_Seven_Rows_That_GAIN_A_Name_Are_Wood_Joists_And_Rafters` names them. The other **119** refusals stay refusals, asserted.

## RED then GREEN

| Gate | RED | GREEN |
|---|---:|---:|
| `Timber_For_One_Is_Timber_For_All` (three unmodified consumers) | **85 of 1,808** names — and the failure list **names `MDF CEILING PANEL 12MM` and `SOLID BAMBOO 14MM`**, as required | 0 |
| `SubstanceVocabularyTests` | **5 failed / 3 passed of 8** | **8 of 8** |
| `Not_One_Row_The_Run_Named_Loses_Its_Substance` (whole-word matcher, stems removed) | **7 of 168** — the brief's figure, reproduced exactly | 0 |
| `TypeRenameSubstanceStemTests` | **5 failed / 5 passed of 10** | **10 of 10** |

The W5c RED is the precise claim: the matcher was already whole-word in that run, and only the stems were removed, so the 7 is the cost of the swap and not of something else.

**A probe defect was found and fixed rather than worked around.** The renamer probe first ran on `Walls`, where `CodeFor` has no `Plywood` key — so five plywood names read as "not timber", a gap in the **code table** being reported as a gap in the **vocabulary**. It probes `Floors` now, which has entries for both `Timber` and `Plywood`, and the docstring says why. Reported because it changed the RED count (90 → 85) and an unexplained number is not evidence.

- Tags: 839 passed, 0 failed (baseline 821)
- Boq: 1249 passed, 0 failed — `BiogenicCarbon` is linked there too, so `SubstanceVocabulary` was added to that project's includes
- `dotnet build StingTools/StingTools.csproj -c Debug` — 0 warnings, 0 errors
- `tools/check_workflow_wiring.ps1` — Tier 4 = 0; `tools/check_path_discipline.ps1` — OK

## Fixture

`Fixtures/type_rename_core_materials_20260909.csv` — the 294 rows of `type_rename_plan_20260909_075110.csv` whose `Reason` names a core material, with the proposal that run produced. **A different fixture from the `type_rename_20260909.csv` that #902 adds**, deliberately: that PR is unmerged and stacking on it is what orphans a change. The two should be reconciled when both land.

## NOT verified in Revit

`deploy.bat` was not run and nothing here executed in a Revit session.

- **No embodied-carbon report was regenerated.** The eighteen carbon changes above are the vocabulary's answers over material **names**; what they do to a project's kgCO₂e depends on quantities this has not seen.
- `Baseline_RenameTypes` was not run. The seven newly-named types and the five CMU walls are re-planned from the recorded core materials, not observed.
- `MaterialClassPlanner`'s `Map` is now built at static-init rather than declared as a literal. The order is unchanged and the 1,815-name corpus test passes unmoved, but the change is structural and has only been exercised headlessly.

## Expected conflict

This PR and **#906** both edit `MaterialClassPlanner.cs` — #906 adds a registry consult before the needle loop, this one splices the wood rows into the table. Different regions of the same file; whichever merges second will need a small manual resolution. Flagged rather than avoided by stacking.

https://claude.ai/code/session_01WoCsYLA9TmuDC1F2DDTx26

### DIFF
--- a/StingTools.Boq.Tests/StingTools.Boq.Tests.csproj
+++ b/StingTools.Boq.Tests/StingTools.Boq.Tests.csproj
@@ -28,6 +28,10 @@
     <Compile Include="..\StingTools\BOQ\MeasuredAddition.cs" Link="MeasuredAddition.cs" />
     <Compile Include="..\StingTools\UI\MaterialLookupParser.cs" Link="UI\MaterialLookupParser.cs" />
     <Compile Include="..\StingTools\BOQ\BiogenicCarbon.cs" Link="BiogenicCarbon.cs" />
+    <!-- W3: BiogenicCarbon no longer keeps its own timber word-list; it reads the one
+         shared with MaterialClassPlanner and TypeRenamePlanner. Two copies of that list
+         disagreed on 85 of 1,808 real material names, on a carbon figure. -->
+    <Compile Include="..\StingTools\Core\Materials\SubstanceVocabulary.cs" Link="Core\Materials\SubstanceVocabulary.cs" />
     <!-- CA-1 — Document-free currency converter; pins the no-silent-mis-currency
          invariants headlessly (UGX base, default→UGX, single FX). -->
     <Compile Include="..\StingTools\BOQ\Rates\RateCurrency.cs" Link="Rates\RateCurrency.cs" />
--- a/StingTools.Sustainability.Tests/StingTools.Sustainability.Tests.csproj
+++ b/StingTools.Sustainability.Tests/StingTools.Sustainability.Tests.csproj
@@ -80,6 +80,11 @@
     <!-- WS A3 — pure BOQ waste + biogenic helpers shared with the materials carbon path. -->
     <Compile Include="..\StingTools\BOQ\WasteFactor.cs" Link="Engines\WasteFactor.cs" />
     <Compile Include="..\StingTools\BOQ\BiogenicCarbon.cs" Link="Engines\BiogenicCarbon.cs" />
+    <!-- W3: BiogenicCarbon reads the ONE timber vocabulary now, which needs the
+         whole-word matcher with it. Three copies of that list disagreed on 85 of
+         1,808 real material names, on a carbon figure. -->
+    <Compile Include="..\StingTools\Core\Materials\SubstanceVocabulary.cs" Link="Engines\SubstanceVocabulary.cs" />
+    <Compile Include="..\StingTools\Core\MaterialSchedule\PatternMatch.cs" Link="Engines\PatternMatch.cs" />
     <Compile Include="..\StingTools\Core\Sustainability\IMetricProvider.cs" Link="Engines\IMetricProvider.cs" />
     <Compile Include="..\StingTools\Core\Sustainability\SchemeEvaluator.cs" Link="Engines\SchemeEvaluator.cs" />
     <Compile Include="..\StingTools\Core\Sustainability\EdgeKpiSnapshot.cs" Link="Engines\EdgeKpiSnapshot.cs" />
--- a/StingTools.Tags.Tests/Fixtures/type_rename_core_materials_20260909.csv
+++ b/StingTools.Tags.Tests/Fixtures/type_rename_core_materials_20260909.csv
@@ -1,0 +1,295 @@
+Category,CurrentName,CoreMaterial,RecordedProposal,RecordedOutcome
+Ceilings,16mm GWB on Metal Stud,Metal Stud Layer,,cannot name
+Ceilings,16mm GWB on Wood Furring,"Softwood, Lumber",PLNS_CPB_Timber38-Boarded,rename
+Ceilings,600mm x 1200mm ACT System,Default,,cannot name
+Ceilings,600mm x 600mm ACT System,Default,,cannot name
+Floors,ANTI-SLIP EPOXY 5MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
+Floors,ANTI-STATIC FLOOR 5MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
+Floors,BLUESTONE PAVING 40MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
+Floors,BROADLOOM CARPET 8MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
+Floors,CARPET TILE 6MM 500X500MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
+Floors,CAST IN-SITU TERRAZZO 30MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
+Floors,CERAMIC MOSAIC 6MM,"Concrete, Cast-in-Place gray",PLNS_SLB_RC100,rename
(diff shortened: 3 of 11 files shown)


StingD85:
**Independent review — recommend merge, with one gap named and one figure corrected.**

Re-measured against `origin/main` @ `25f1448c7` (i.e. after #904/#905/#906/#908 landed), not inherited from the branch.

### Confirmed

- **The `"ply "` finding is exactly right.** I pooled the register and the fixtures myself: the substring needle matches **3** names, and all three are false positives — `PVC SINGLE PLY 1.5MM` (a PVC membrane) and two `SUPPLY REGISTER …` air diffusers (sup-**PLY** ). Zero true positives. Dropping it rather than making it whole-word is the right call.
- Matching goes through `PatternMatch.Contains`, i.e. whole-word — the anchoring requirement is met.
- Tags **901 → 919**, 0 failing. Boq 1311 and Sustainability 438 unchanged. Build 0 warnings / 0 errors.

### One figure is wrong

The body and the `SubstanceVocabulary.cs` header both say *"the **1,808** distinct names in the shipped register plus the delivered-model corpus"*.

**1,808 is the corpus WITHOUT the register.** Measured:

```
BLE_MATERIALS + MEP_MATERIALS   :    40
fixtures (material_names_*.csv) : 1,808
union                           : 1,848
```

The count is quoted five times across the header, the PR body and two test comments. The *test* is unaffected — it asserts `set.Count > 1500`, not a literal — so this is prose-vs-data only, but it is the exact defect class this repo keeps producing. Worth correcting the wording to 1,848 (or "1,808 corpus names plus the 40-name register") before merge.

### The gap: nothing pins the whole-word semantics

I sabotaged `IsWood` back to a bare `Contains` — the defect the header says the matcher exists to prevent — and **the whole suite still passed.**

That is not a deficient test suite. I measured it: with `ply` gone, whole-word and substring give **identical answers on all 1,848 names**. There is nothing on today's data to catch.

But it means the anchoring is **preventive, not load-bearing**, and the header slightly overstates by presenting it as the fix. What actually fixes the three false positives is dropping `ply`; the whole-word swap changes nothing measurable today.

It becomes load-bearing the moment anyone adds a short word — `ash`, `elm`, `fir`, `teak` — and at that point nothing would fail. A synthetic assertion would close it cheaply and does not depend on the corpus:

```csharp
// Whole-word, not substring: these contain a wood word inside a longer word.
[InlineData("CORKSCREW HOLDER", false)]   // CORK screw
[InlineData("MAPLEWOOD VENEER", true)]    // "maple" is bounded; "wood" is not
[InlineData("OAKUM CAULKING", false)]     // OAK um
```

(`MAPLEWOOD` is the interesting one — it is true via `maple`, not via `wood`, which is what whole-word means.)

Recommendation: **merge**, ideally after fixing 1,808 → 1,848 and adding a synthetic whole-word pin. Neither blocks; both are cheap.

PR #913, opened 2026-09-09
Type creation stops reporting success after failing, and stops inventing what it was not given (W1-W4)

StingD85:
W1, W2, W3 and W4 from `HANDOVER_TYPE_CREATION_TRUTH.md`, in one PR: three of them are the same file and the fourth is W1's only real regression gate. Branched from `origin/main` (`625212814`) after W0.

## W3 first, because it changed what W2 should do

**Two denominators, and conflating them is how the old comment stayed wrong.**

Across the **files** — 326 BLE rows and 120 MEP rows declare no layer at all, 446 of 1,279. The comment being replaced said the homogeneous fallback *"covers 165+ BLE rows with no layer data"*. It is **326**, out by nearly two-fold, and that comment was the only record anywhere.

Across the rows a **command actually reads** — the four host-type commands select **380** of the 815 BLE rows by `MAT_ELEMENT_TYPE`:

| | Walls | Floors | Ceilings | Roofs | total |
|---|---:|---:|---:|---:|---:|
| rows | 115 | 95 | 81 | 89 | **380** |
| built from declared layers | 64 | 95 | 81 | 89 | **329** |
| homogeneous (row declares none) | 51 | 0 | 0 | 0 | **51** |
| invented thickness | 0 | 0 | 0 | 0 | **0** |
| dropped layer | 0 | 0 | 0 | 0 | **0** |
| R-value in a material column | 0 | 0 | 0 | 0 | **0** |

### The central question is answered in the NEGATIVE

The brief asked me to establish why all 95 `FLR-*` rows carry a layer-1 material and yet 87 floor types ended up as one default layer. **Neither path in this file can produce them**, and I checked four hypotheses rather than one:

1. **Column drift** — the positional constants (`ColLayer1Start = 16`, `LayerStride = 3`) were my first hypothesis. Checked against both 72-column headers: **correct**. Now pinned by a test.
2. **The homogeneous fallback** — all 95 `FLR-*` rows carry a valid layer 1, so it fires for **0 of 95**.
3. **`BuildLayers` producing the material** — it only ever uses a material named in the CSV's layer columns or the row's own `MAT_NAME`. `Concrete, Cast-in-Place gray` is a Revit stock material. It *does* appear in 54 BLE rows — in column 34, `BLE_APP-REVIT-BASE-MATERIAL`, which **nothing in `CompoundTypeCreator` reads**. And if it did, the thicknesses would differ per row; all 87 share 100 mm.
4. **`ParseThickness`** — returns the row's declared thickness (50 / 75 / 40 mm); its only fallback is **10 mm, never 100**.

One identical 100 mm layer across 87 differently-declared rows is the signature of a **duplicated base type whose compound structure was never replaced**. That is consistent with W1's catch path and with any other duplicate-and-rename route. **It is not established**, and I am not claiming it: the ten logs on the live plugin path start 2026-08-17 and hold no type-creation line at all. The only other shipped `FloorType` duplicator, `StructuralTypeFactory`, names types `STING Slab {n}mm` and is not it either.

## W1 — four creators returned `true` after failing

`CreateWallType` / `CreateFloorType` / `CreateCeilingType` / `CreateRoofType` each caught a `SetCompoundStructure` failure, logged, and returned success. The comment said it honestly — *"type was created, just no layers"* — and the function then told its caller the whole operation succeeded. `CreateMEPType`, eighty lines below, has always returned false on every failure path.

They return **`false`** now, and the half-made type is **deleted inside the same transaction**. The brief asked for the choice to be explicit, so it is argued in the code: a type carrying a build-up nobody asked for is a wrong quantity that looks like a right one, and it measures, prices and carbon-counts exactly like a real one. Deleting is safe *here specifically* because the type was duplicated microseconds earlier in the same transaction and this command places no instances — and if that ever stops being true, keep it and report it as unbuilt, but do not go back to reporting it as built.

A delete that itself fails is its own **louder** outcome (`RefusedButLeftBehind`) with a `WARNING` line, never folded into an ordinary refusal.

`TypeCreationTally` holds the decision and the counting rule, Revit-free. "Created N types" no longer means two different things.

## W2 — the four substitutions, carried out of the function

| old behaviour | now |
|---|---|
| `if (layerThickMm <= 0) layerThickMm = 10;` silently | reported, and **fatal** |
| `if (layerThickMm < 1.0) layerThickMm = 1.0;` silently | reported, **not** fatal — Revit's own minimum, 180 rows take it |
| `>500 mm` → `continue` (warned) | reported, and **fatal** |
| material create fails → layer added with the wrong material | reported, and the substitution is **named** |

**The brief asked me to measure before deciding whether an invented thickness should fail the type. It should, and it is free:** zero rows the commands read trip it. That is the whole argument for doing it now rather than when a register row makes it expensive.

### A fifth defect the brief did not list

`CountActualLayers` `continue`d past a blank slot and returned a **count**, which the caller used as an **index bound**. A row populating slots 2, 3, 4 was read as slots 1, 2, 3 — a blank material at slot 1, and slot 4 silently dropped. `WL-024 GYPSUM DRYWALL SYSTEM 25MM` and `WL-037 FIBER CEMENT BOARD 15MM` are in that shape, and **both of the register's only two invented thicknesses came from exactly it**. Slots are read by position now.

### Two things measured that contradict the code documenting them

- The `>500` rule cites *"mm² stored as mm, e.g. 300.0 for a 300 mm² conductor"* as its reason. **300 is below 500**, so that conductor is built as a 300 mm layer. Pinned by a test rather than tidied — changing the threshold would be a quantity change made on no evidence.
- The rule fires on **no row of either file**. It is a guard, not a fix, and the tests say so.

## W4 — the sweep, done properly

`tools/find_lying_catches.py` brace-matches method bodies rather than pattern-matching text, because a regex over a brace language cannot see *"this catch block ends here and the next statement on this path is `return true`"*.

**Its own first version repeated the brief's false negative by a different route.** Requiring the braces between the catch and the return to *balance* made it miss all four confirmed cases, which sit inside an `if`. Closing braces are fine — they only mean we left an enclosing block; an opening brace is not. Validated by running it against the pre-fix file and requiring it to name all four.

It also does **not** count a catch that records the failure where the caller or user will see it. The `foreach { try/catch { errors.Add(…) } } … return Result.Succeeded` shape is a command that succeeded and said what did not; counting those is the trap CLAUDE.md's 683 empty catches already sets. That filter took the raw number from 372 to 152.

| | total | Tier 1 (bool helpers) | Tier 2 (`Execute`) |
|---|---:|---:|---:|
| before | **156** | **25** | 131 |
| after | **152** | **21** | 131 |

**Tier 1** is bool-returning helpers, where the return value is the caller's only signal — all four confirmed cases were there. **Tier 2** is `Execute` returning `Result.Succeeded`; eight sampled at random were **all** optional side-effects after the real work had been done and reported (opening Explorer, copying to the clipboard, refreshing a panel, setting a window owner).

**152 is a triage list, not 152 defects.** The only ones verified by reading are the four now fixed. Wired into CI as a **ratchet that fails when the count rises**, with a baseline file and a note saying to raise it only with a reason.

## RED then GREEN

| Gate | RED | GREEN |
|---|---:|---:|
| `CompoundLayerPlanTests` — old `BuildLayers` semantics restored inside the planner (count-as-index, silent 10 mm, silent drop, ParseFailure folded into homogeneous) | **11 failed / 3 passed of 14** | **14 of 14** |
| `tools/find_lying_catches.py` — run against the pre-fix `FamilyCommands.cs` | **156 total, Tier 1 = 25**, naming `CreateWallType:405`, `CreateFloorType:442`, `CreateCeilingType:477`, `CreateRoofType:512` | **152, Tier 1 = 21**, none in that file |
| the sweep's own validation | first version found **0 of 4** — the same false negative it was written to replace | finds **4 of 4** |
| `TypeCreationTallyTests` | new unit, so no before-state; it asserts instead that the **old** decision printed an *identical* report for a run that built everything and one that refused half | **6 of 6** |

- Tags: 939 passed, 0 failed (baseline 919)
- Boq: 1311 passed, 0 failed
- `dotnet build StingTools/StingTools.csproj -c Debug` — 0 warnings, 0 errors

## Figures I re-measured that differ from the brief

| Brief | Measured |
|---|---|
| homogeneous fallback "covers 165+ BLE rows" → brief corrected to **326** | **326** confirmed — and MEP adds **120**, which the brief did not give |
| line numbers for the four `return true` (408 / 445 / 480 / 515) | the returns are at **405 / 442 / 477 / 512** on `origin/main`; the functions and the shape are exactly as described |
| `>500 mm` drop, with "300.0 for a 300 mm² conductor" as the example | the threshold is 500, so **300 does not trip it**; and **no row of either file** trips it |
| — | **`invent 10 mm` fires on only 2 rows**, and both are caused by the count-as-index defect, not by malformed cells |

## NOT verified in Revit

`deploy.bat` was not run and nothing here executed in a Revit session.

- **`doc.Delete` on a just-duplicated `HostObjAttributes`, inside the same transaction, is unproven.** It is the load-bearing half of the W1 decision. If it throws, the code takes the `RefusedButLeftBehind` path and says so loudly — that branch is also unexercised.
- **`SetCompoundStructure` is now called through the `HostObjAttributes` base** rather than on `WallType`/`FloorType`/`CeilingType`/`RoofType` directly. It compiles; it has not run.
- `cs.OpeningWrapping = OpeningWrappingCondition.None` is now applied on **all four** kinds (it was on three). That is a behaviour change on walls and it has not run.
- The new report text — the tally's `Report()` and the read-notes block — has not been seen by a user.
- **No compound structure was rewritten on any model, and no register row was edited.**

## Unrelated, found while merging W0

`StingTools/Core/Baseline/TypeRenamePlanner.cs` contains a **literal NUL byte** inside a string on `main` (from #902's `(p.Category ?? "") + "\0" + held.Trim()`, written as a raw 0x00 rather than the two-character escape). Git treats the file as binary, so it does not diff or auto-merge — the whole file conflicted during W0 and had to be re-applied by hand. One-character fix; not made here because it is unrelated to this PR.

https://claude.ai/code/session_01WoCsYLA9TmuDC1F2DDTx26


StingD85:
**Change requested — one item.** Taking almost all of this otherwise.

The negative result on W3 is the most valuable thing in the run: establishing that *neither* path in `FamilyCommands.cs` built the 87 retires a hypothesis the brief had written as near-certain, and you established it rather than asserting it. `RefusedButLeftBehind` as an outcome distinct from `RefusedStructureFailed` is exactly right — a type left carrying the base build-up is a worse result than a clean refusal, not the same one. Pinning the dead `>500 mm` rule with a test instead of tidying it away, and splitting the 152 sweep hits into 21 bool-helpers and 131 `Execute` sites with "152 is a triage list, not 152 defects", are both right.

## `OpeningWrapping = None` must stay off walls

`FamilyCommands.cs:427-429`:

```csharp
// Revit applies default end-cap / wrapping conditions that are invalid on
// non-wall types; clearing them is why this ever threw.
cs.OpeningWrapping = OpeningWrappingCondition.None;
```

**The comment says non-wall and the code says all four.** On `main` the setting appears three times — Floor, Ceiling, Roof — and `CreateWallType` deliberately lacks it. The original `// BUG-01 FIX` comment names walls as the exception, and this consolidation reverses that exclusion while deleting the sentence explaining it.

On a wall the setting is not a workaround, it is load-bearing: `OpeningWrapping` decides whether layers wrap into a door or window reveal. That is a graphics answer and a takeoff answer, and this would change it for every wall type the command creates. You disclosed it, which is why this is a request rather than a complaint — but it is a side effect of deduplication, not a decision anyone made, and it is the only thing in this PR that changes model output.

The helper already takes `kind`:

```csharp
// Revit applies default end-cap / wrapping conditions that are invalid on
// non-wall types; clearing them is why this ever threw. Walls are excluded
// deliberately: there the setting decides whether layers wrap into a door or
// window reveal, which is a graphics AND a takeoff answer, and the four bodies
// this replaced left it alone for that reason.
if (!string.Equals(kind, "Wall", StringComparison.OrdinalIgnoreCase))
    cs.OpeningWrapping = OpeningWrappingCondition.None;
```

Please gate it — assert a wall's structure is not given `None` while the other three are — so the next deduplication cannot quietly re-merge them.

## Two corrections to the report, so the record is right

- **The four `return true` are at 408/445/480/515.** Verified on `8139edf20` (the commit the brief was written against) and on current `origin/main`. `405/442/477/512` is off by three — probably read post-edit. Immaterial to the work.
- **The register column is 35, not 34** (1-based). The substance is right and it matters: `BLE_APP-REVIT-BASE-MATERIAL` is populated on 54 rows and nothing in `CompoundTypeCreator` reads it.

One measurement that strengthens your conclusion. The 95 `FLR-*` rows declare **four different** base materials — `Tile, Mosaic, Gray` x22, `Concrete, Cast-in-Place gray` x16, `Concrete Masonry Units` x14, `Concrete, Precast` x8 — yet all 87 in the model carry the same one, and `ProjectBaselineCommands.cs:474` duplicates from a single `source` for every row. So the uniformity does not merely suggest an unreplaced structure: it **proves the register's base-material column played no part**, because had it been read the 87 would be four materials, not one.

## Also merged

#915 fixes the raw NUL byte you found in `TypeRenamePlanner.cs`. That was mine, from #902 — I meant `"\u0000"` and shipped `0x00`. You were right to flag it rather than tidy it, and right that it cost the #907 merge real time: a source file git treats as binary cannot be diffed, which means it cannot be reviewed either. `main` is clean.

## Still unproven, and worth repeating in the PR body

`doc.Delete` on a just-duplicated `HostObjAttributes` inside the same transaction is the load-bearing half of the W1 decision, and neither it nor its failure branch has executed in Revit. `SetCompoundStructure` through the `HostObjAttributes` base rather than the four concrete subtypes is in the same position.

StingD85:
## The register audit ran on the real model. It moves this from "consistent with" to established.

`Materials_RegisterAudit` executed in Revit on `Amanda S.Herring.rvt`, 2026-09-10 20:28.
471 rows. **Your reconstruction was faithful: 28 Flattened / 67 Differs, exactly.**

```
Verdict         Category    n
Matches         Roofs      89
NoStructure     Ceilings   81
Differs         Floors     67
Flattened       Floors     28
Matches         Walls      49
NotInRegister   (17 cats) 157
```

### 1. One material across all 95, not four

Every Flattened row and every Differs row reports the same thing:

```
STANDARD CEMENT SCREED 50MM  register: 1 × 50 mm CEMENT SCREED       model: 1 × 100 mm Concrete, Cast-in-Place gray
POLISHED CEMENT SCREED 50MM  register: 1 × 50 mm + 1 × 1 mm POLISH   model: 1 × 100 mm Concrete, Cast-in-Place gray
LUXURY VINYL TILE 5MM        register: 1 × 5 mm LVT + 1 × 2 mm UNDER model: 1 × 100 mm Concrete, Cast-in-Place gray
```

67 of 67 Differs say `Concrete, Cast-in-Place gray`. 67 of 67 say `1 × 100 mm`.

The register declares **four different** base materials across those 95 `FLR-*` rows — `Tile, Mosaic, Gray` ×22, `Concrete, Cast-in-Place gray` ×16, `Concrete Masonry Units` ×14, `Concrete, Precast` ×8. The model shows one. Had `BLE_APP-REVIT-BASE-MATERIAL` been read, they would be four. **A single source type was duplicated 95 times and its structure was never replaced** — which is the shape your W1 catch path produces.

The Flattened/Differs split turns out to be cosmetic: both are the identical unreplaced base, one measured against a multi-layer register row and one against a single-layer row.

### 2. The pipeline works. Floors specifically fail.

**138 types Match** — 89 roofs and 49 walls carry the build-up their register row declares. This is not a broken importer; it is a per-category failure, which is what a `SetCompoundStructure` that throws for some categories and not others looks like.

A dated window fits: `CreateFloorType` shipped **2026-02-26** (`7a01ade9e`), and the `OpeningWrapping = None` fix — whose comment says *"invalid on non-wall types"* — landed **2026-03-08** (`3ee61d573`). Ten days in which floor structure assignment would throw, be caught, and return true. Whether these 95 passed through that window cannot be read off the model, so I am not claiming it. But it explains why the fix was written for floors: floors are what threw.

### 3. Nothing placed is affected

**Only 13 of 471 types carry elements, and not one of them is Flattened or Differs.**

```
CLAY BRICK VILLAGE LARGE (295×150×130MM)   48   Matches
Generic - 225mm / 2 / 3                     22   NotInRegister
stepsr 7                                     8   NotInRegister
Exterior_CreamWhite_230 2 / BrownWhite_230  11   NotInRegister
coping                                       2   NotInRegister
IntFloor_tile_150                            1   NotInRegister
```

No quantity in this model is currently wrong because of the 95 — the carbon and cost error is latent, not realised, and lands the moment someone draws a floor with one of those types. The one placed register-named type, `CLAY BRICK VILLAGE LARGE` with 48 elements, **Matches**.

### 4. A failure mode nobody had: 81 ceilings with no structure at all

```
GYPSUM BOARD STANDARD 9.5MM   register: 1 × 9.5 mm   model: no compound structure
```

Not wrong — **absent**. All 81 ceiling types in the register set. `CreateCeilingType` is one of the four, so the same mechanism fits, but the outcome is different enough to deserve its own look and it was in no brief.

---

## Pushed to this branch: the OpeningWrapping exclusion

`ec28d2d6d`. The four bodies set `OpeningWrapping = None` on Floor, Ceiling and Roof and never on Wall; the consolidation applied it to all four and deleted the comment that named the exception. `TypeCreationTally.IsWallKind` now carries the exclusion somewhere Revit-free so a test can hold it, and `ApplyStructureOrFail` guards on it.

Honest about the gate's reach: **the predicate is tested; the call site is verified by reading**, because it is Revit-bound. A source-scraping test would discriminate but rot. Build 0/0, Tags 947/947.

With that in, I would merge this.

Still unexercised in Revit and worth keeping in the PR body: `doc.Delete` on a just-duplicated `HostObjAttributes` inside the same transaction, and its failure branch.

PR #929, opened 2026-09-10
Second KUT build prompt: finish the defect class #927 half-closed

StingD85:
Docs-only — one work prompt and its index row. Branched from `main`, **not** stacked on #927.

Adds [`docs/PROMPT_KUT_INTEGRATION_BUILD_2.md`](docs/PROMPT_KUT_INTEGRATION_BUILD_2.md), the second autonomous build pass.

## Why there is a second pass

Reviewing #927 meant re-running its gates and sabotaging its work five ways — swallowing an `HttpRequestException` as `EmptyOk`, mapping 404 to `EmptyOk`, making a schema change read as empty, renaming a key in the Niagara example, and injecting a bogus `commandTag`. **All five were caught.** The work holds up and should merge.

But it applied `AccFetchResult<T>` to **one** read path. The same defect is live in two more, and one is worse than the one that was fixed.

## The two that remain

**A1 — a failed ACC issue read reads as "the issues are gone."** `AccIssueSync.PullIssuesAsync` still returns a bare list: empty on auth failure (`:269`), **partial** on a mid-pagination failure (`:285`). `AccSyncIssueStatusCommand` then reports every tracked clash as `NOT_FOUND` — which reads to a coordinator as ACC having deleted the issues, not as a read that failed — and writes the sidecar anyway (`:87`), presenting a partial reconciliation as a complete one.

**A2 — a malformed Niagara feed reads as "live, nothing commissioned", and eats the cache.** Four steps compound:

1. `NiagaraPointParser.Parse` correctly sets `Error` and returns empty points (`:80`).
2. `NiagaraJsonClient.ParsePoints` logs it and **returns `r.Points` anyway (`:77`), dropping the `Failed` flag**.
3. `FetchPoints` hands back a non-null empty dictionary, so `CommissioningSource` takes the **live** branch and reports `"live (captured …)"`.
4. `Persist` then **overwrites `last_station_points.json`** — destroying the cached fallback that exists for exactly this outage.

A JSON error envelope such as `{"error":"unauthorized"}` is the nastiest variant: it parses as an object, yields zero points and sets **no error at all**. This chain feeds a 5D commissioning valuation that the cost module carries toward payment certification.

## The rest

A3 closes the gap my review found — the Revit-bound command branch that decides `Result.Failed` is verified by reading only. A4 gates `check_kut_workflow_tags.py` in CI (it is well built and self-testing, but nothing runs it, so it rots). A5 resolves the upload-chaining question #927 correctly left open: `ACCPublish` already writes a deterministic `zipPath` (`PlatformLinkCommands.cs:1196`) and registers it (`:1201`), so *"upload the bundle step 6 just made"* is a file choice that is **not** a guess — while still forbidding an automatic upload into an issued CDE container before B1 proves one live round-trip. A6 aligns `AccModelUpload`'s result shape. A7 corrects three stale `CLAUDE.md` claims.

## Discipline carried forward

Every line citation was verified against the branch before writing; **six were tightened** after checking. The prompt reuses `AccFetchResult<T>` rather than inventing a parallel mechanism, names the `StingLogShim` pattern that makes `CommissioningSource` testable at all, and requires asserting **attribution** rather than verdict — the technique that caught #927's own alternative-path escape.

It also warns against the over-correction its framing invites: an empty station, an empty container and a clash-clean model set are real, expected states. Making emptiness itself an error would be worse than the defect, because it would make a clean cycle unreportable.

Section B is unchanged and quarantined. **B1, the APS credential, is overdue** and remains the only thing between the pack and a proven coordination cycle.

`check_docs_index.py` and `check_kut_documents.py` (100 assertions, 5 docs) both green.

### DIFF
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -169,6 +169,10 @@ proven coordination cycle ·
 2026-09-10) — the build that closes what the readiness report found, ordered by what unblocks KUT
 soonest. Eight offline tasks, each with the command that proves it and the sabotage that must break
 it; three third-party-blocked items are quarantined in their own section ·
+[`PROMPT_KUT_INTEGRATION_BUILD_2.md`](PROMPT_KUT_INTEGRATION_BUILD_2.md) (work prompt, written
+2026-09-10) — the second build pass, after PR #927. Finishes the defect class #927 half-closed:
+a failed ACC issue read still reads as "the issues are gone", and a malformed Niagara feed still
+reads as "live, nothing commissioned" while overwriting the cached fallback. Seven offline tasks ·
 [`PHASE6B_CAPABILITY_VERIFICATION.md`](PHASE6B_CAPABILITY_VERIFICATION.md) ·
 [`MEP_PRINT_READY_PUNCHLIST_RUNNER.md`](MEP_PRINT_READY_PUNCHLIST_RUNNER.md) (work prompt)
 
--- a/docs/PROMPT_KUT_INTEGRATION_BUILD_2.md
+++ b/docs/PROMPT_KUT_INTEGRATION_BUILD_2.md
@@ -1,0 +1,500 @@
+# Work prompt 2 — finish the KUT integration work (Fohlio · Niagara · ACC)
+
+**For an agent executing autonomously with no further input.** You cannot ask questions. Everything
+you need is here. Where something is genuinely unknowable, the task says to leave it blank with a
+marker and report it — never to guess.
+
+This is the **second** build pass. The first (PR #927, branch `claude/kut-integration-build`) fixed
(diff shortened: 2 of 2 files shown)


StingD85:
Superseded, not abandoned — the content is already on `main`.

PR #944 (which carried #932's work) shipped `docs/PROMPT_KUT_INTEGRATION_BUILD_2.md` at a **byte-identical blob** (`43fcfa6f0`) plus the same `docs/INDEX.md` entry, so this PR has nothing left to add.

Merging it would be actively harmful rather than merely redundant: this branch predates ~12,000 lines of work that has since landed on `main`, so the merge would reintroduce a stale tree for zero content gain.

Closing and deleting the branch. Nothing is lost.

PR #930, opened 2026-09-10
W2: resolve an element's MAT_CODE through its material — and the rate delta, measured

StingD85:
> **This PR changes what the BOQ rate chain looks up.** The delta is stated below with its numbers, measured on the Herring model. It is **zero**, and the reason is a finding in its own right.

## What

Four places asked a wall for a parameter bound to materials:

```csharp
MatCode = ParameterHelpers.GetString(el, "MAT_CODE") ?? "",
```

`BOQCostManager:1184` and `:2816` (the rate request), `CostStamp:152` (the rate request **and** the shared rate-cache key), `MeasurementStandards:209` (the CESMM4 description). `MAT_CODE` binds to `Materials` and to nothing else — `CATEGORY_BINDINGS.csv:4693`, agreeing with four other data files — so all four returned empty for every element ever costed, and `CsvRateProvider` **"Pass 3 — MATERIAL"** (`RateProviders.cs:323`, confidence 85) has never fired.

All four now go through **one** resolver. The decision is Revit-free — `MaterialCodeResolution` takes a layer list plus a name→code map:

1. **primary layer material** → its code (`PrimaryMaterialSelector` picks the layer — no second definition of "primary")
2. the element's **single material**, only when the layers name none
3. the element's **own** `MAT_CODE`, for the day somebody binds it there
4. **empty — and empty stays empty**

**The subtlety is the whole risk.** When the layers DO name a core and that core has no code, resolution **stops**. Falling through would hand back the FINISH SKIN — a 230 mm rendered masonry wall answering "gypsum" — which is #873 exactly. A guess would ride at confidence 85, above the category match.

---

## The rate delta, measured on the Herring model

From the model's own exported artefacts (`register_audit_20260910_202805.csv`, `type_rename_core_materials_20260909.csv`) against the shipped register and `cost_rates_5d.csv`. No Revit.

| Category | types | instances | gain a `MAT_CODE` | coded instances | Pass 3 matches | old rate UGX | **delta** |
|---|---:|---:|---:|---:|---:|---:|---:|
| Floors   | 98 | 1  | **0**  | 0  | 0 | 444,000 | **0** |
| Roofs    | 96 | 11 | **88** | 0  | 0 | 555,000 | **0** |
| Walls    | 96 | 61 | **49** | 48 | 0 | 315,000 | **0** |
| Ceilings | 4  | 0  | **0**  | 0  | 0 | 166,500 | **0** |
| **Total** | **294** | **73** | **137** | **48** | **0** | | **0 UGX** |

**137 host types gain a `MAT_CODE`. Zero rates change.** The reason is exact, and it is the same defect this work was opened for, one file over:

> **Not one of the register's 1,279 codes is a key in `cost_rates_5d.csv`.**
> The register issues `FLR-028`, `CLG-052`, `WL-128`. The rate table has a column **called** `MAT_CODE` holding 43 three-letter, PROD-shaped abbreviations — `FLR`, `CLG`, `WAL`, `AHU`. One column name, two vocabularies, empty intersection.

Pass 3 *can* now fire and still *does* not. Those are different sentences, and the difference is the deliverable: before W2 the lookup was dead at the key; now it is dead at the table, which is a place a QS can fix by editing a rate card.

**Floors gain nothing, and that is the 87-type finding again** — 90 of the 98 floor types have the core material `Concrete, Cast-in-Place gray`, a Revit stock material absent from the register.

**Pinned, not fixed.** `The_Shipped_Rate_Table_Shares_No_Key_With_The_Register` asserts the empty intersection and fails loudly, **naming the codes**, the moment one crosses over. Changing 43 rate keys IS a cost change and needs its own evidence; this guarantees nobody makes it by accident.

### Two other behaviour changes, stated so they don't pass unremarked

- `CostStamp`'s shared rate cache is keyed `{category}|{disc}|{prod}|{matCode}|{unit}`. A previously-constant empty segment now varies, so the cache **partitions more finely** and can no longer merge two elements differing only by material.
- Every CESMM4 description said `"{category}; as drawn"` with no material named. It will now name the code where one resolves.

## RED and GREEN — by sabotage, both counts

```
A_Layered_Element_Answers_With_Its_CORE
  RED    first layer instead of the selector      WL-SKIN   (4 of 13 fail)
  GREEN                                           WL-CORE

A_Layered_Element_Whose_Core_Has_No_Code_Does_Not_Fall_Back_To_The_Skin
  RED    fall-through restored                    WL-SKIN   (1 of 13 fails)
  GREEN                                           ""

The_Shipped_Rate_Table_Shares_No_Key_With_The_Register
  RED    one row keying FLR-028 added to the rate table — fails, names
         FLR-028, demands the mover count what moved   (2 of 13 fail)
  GREEN  0 of 1,279
```

| | before | after |
|---|---|---|
| `dotnet build` | 0 err / 0 warn | 0 err / 0 warn |
| `StingTools.Boq.Tests` | 1,311 | **1,324** |
| `StingTools.Tags.Tests` | 919 | 919 |

## Left alone deliberately, logged in ROADMAP

`UI/IfcMaterialPsetWriter.cs:69` carries a **second** `ReadPrimaryMaterialName` that skips the compound structure entirely and returns the first material — #873's defect, still live, in a second place, and it is what gets written into `Pset_StingMaterial`. Out of scope for a cost change. Logged as **MC-2**, with MC-1 (the vocabulary split), MC-3 (the concrete floors) and MC-4 (the memo's TTL).

## Not verified in Revit

`deploy.bat` was **not** run. `ElementMatCodeReader` — the compound-structure read, the single-material read, the memoised name→code map and its 30-second stale window — is **entirely unexercised against a live document**. Every defect this area has produced has lived in the Revit-bound half, and this is the Revit-bound half.

https://claude.ai/code/session_01XzLmu1kH7PQKwjgXpYmXNC

### DIFF
--- a/StingTools.Boq.Tests/MatCodeChainEndToEndTests.cs
+++ b/StingTools.Boq.Tests/MatCodeChainEndToEndTests.cs
@@ -1,0 +1,278 @@
+// ══════════════════════════════════════════════════════════════════════════
+//  MatCodeChainEndToEndTests.cs — W4. One row, register to rate.
+//
+//      register row FLR-028   ->  a material carrying MAT_CODE = FLR-028
+//                             ->  a floor type whose PRIMARY layer is that material
+//                             ->  the BOQ resolves MatCode = "FLR-028"
+//                             ->  CsvRateProvider Pass 3 matches
+//                             ->  Provenance "<rate file> MAT_CODE match"
+//
+//  THE ASSERTION IS THE PROVENANCE STRING, not the number. A rate that happens
+//  to be right for another reason is not this chain working — and on the shipped
+//  rate table every one of these categories has a Pass 4 category row, so a test
+//  that checked only the figure would pass on a category average and prove
+//  nothing. That is exactly the shape this codebase produces.
+//
+//  WHAT THIS WOULD HAVE CAUGHT. Every link here existed and was wired four
+//  phases ago; MAT_CODE was bound to Materials only, the BOQ read it off a wall,
+//  and Pass 3 has never once fired. Nothing failed. This test fails.
+//
+//  The rate table used is a SYNTHETIC one keyed on FLR-028, because the shipped
+//  cost_rates_5d.csv shares no key with the register — see
+//  MaterialCodeResolutionTests.The_Shipped_Rate_Table_Shares_No_Key_With_The_
+//  Register, which pins that and is why W2's measured rate delta is zero. The
+//  chain is proven here on the rate card a project would have to write; the
+//  reason it does not fire on shipped data is proven there. Both are true and
+//  they are different facts.
+// ══════════════════════════════════════════════════════════════════════════
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using StingTools.BOQ.Rates;
+using StingTools.Core;
+using StingTools.Core.Materials;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace StingTools.Boq.Tests
+{
+    public class MatCodeChainEndToEndTests
+    {
+        private readonly ITestOutputHelper _out;
+        public MatCodeChainEndToEndTests(ITestOutputHelper output) => _out = output;
+
+        private const string RateFile = "project_rate_card.csv";
+
+        private static string DataDir()
+        {
+            var dir = new DirectoryInfo(AppContext.BaseDirectory);
+            while (dir != null && !File.Exists(Path.Combine(dir.FullName, "StingTools", "Data", "BLE_MATERIALS.csv")))
+                dir = dir.Parent;
+            Assert.True(dir != null, "Could not locate StingTools/Data from " + AppContext.BaseDirectory);
+            return Path.Combine(dir.FullName, "StingTools", "Data");
+        }
+
+        /// <summary>MAT_NAME and MAT_CODE for FLR-028, read from the SHIPPED register.
+        /// Not restated here: a copy that drifted from the file would pass while the
+        /// thing it describes was wrong.</summary>
+        private static (string name, string code) Flr028()
+        {
+            foreach (string f in new[] { "BLE_MATERIALS.csv", "MEP_MATERIALS.csv" })
+            {
+                var lines = File.ReadAllLines(Path.Combine(DataDir(), f));
+                int h = Array.FindIndex(lines, l => l.ToUpperInvariant().Contains("MAT_CODE"));
+                if (h < 0) continue;
+                var hdr = Split(lines[h]);
+                int ci = hdr.FindIndex(c => c.Trim().Equals("MAT_CODE", StringComparison.OrdinalIgnoreCase));
+                int ni = hdr.FindIndex(c => c.Trim().Equals("MAT_NAME", StringComparison.OrdinalIgnoreCase));
+                for (int i = h + 1; i < lines.Length; i++)
+                {
+                    var c = Split(lines[i]);
+                    if (c.Count > Math.Max(ci, ni)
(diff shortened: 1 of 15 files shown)


StingD85:
## W4 added to this branch (commit `10397b3`)

**Why here and not a fourth PR.** W4's chain — register row → material → primary layer → `RateRequest.MatCode` → Pass 3 → provenance — cannot exist without W2's resolver. A branch cut from `origin/main` would have had to carry a copy of `MaterialCodeResolution.cs`, which is stacking with extra steps and would conflict on merge. It is one commit here instead, so this PR still merges as one change with one title.

### The chain

```
register row FLR-028  ->  a material carrying MAT_CODE = FLR-028
                      ->  a floor type whose PRIMARY layer is that material
                      ->  the BOQ resolves MatCode = "FLR-028"
                      ->  CsvRateProvider Pass 3 matches
                      ->  Provenance "project_rate_card.csv MAT_CODE match"  @ 612,500 UGX/m2
```

**The assertion is the provenance string, not the number.** The test's rate card keys `Floors` at 444,000 and `FLR-028` at 612,500, so a broken chain still returns a plausible figure. Proven load-bearing: a sabotage that leaves Pass 3's **rate unchanged** and only mislabels its provenance as a category average fails **exactly one test — this one — on the string**.

**What it would have caught.** Every link existed and was wired four phases ago. Nothing failed. Sabotaged back to the pre-W2 read, `FLR_028_Walks_From_The_Register_To_A_Rate` reports `""` where `FLR-028` was expected, and `The_Skin_Never_Prices_The_Floor` reports `WL-999` — the plasterboard finish pricing the floor slab.

### `CsvRateProvider` is now a compute/present split

That is what made any of this assertable. The five passes moved to `BOQ/Rates/CsvRateLookup.cs` — **Revit-free, one copy, called by the provider** — and `RateResolutionLevel` moved to its own file because `IRateProvider.cs` imports the Revit API for `RateRequest.Element`. Same namespace, **no call site changed**. CLAUDE.md P1 #4, on the provider that prices most of a model.

`The_Pass_Order_Is_Specificity_Order` strips the winning key one at a time and requires the next-most-specific to answer — five levels, five confidences, one loop — so a reordering fails here rather than on a tender. That order has been wrong before (K-16b: category consulted first and returning, pricing a fire door and a cupboard door alike).

### The rate card is synthetic, and says so

The shipped `cost_rates_5d.csv` shares no key with the register, so the chain is proven on the rate card a project would have to write, while the reason it does not fire on shipped data is proven next door in `The_Shipped_Rate_Table_Shares_No_Key_With_The_Register`. Different facts; neither stands in for the other. **The measured rate delta in the PR description is unchanged: 0 UGX.**

### RED and GREEN

```
FLR_028_Walks_From_The_Register_To_A_Rate
  RED    pre-W2 read restored — expected "FLR-028", actual ""     (2 of 7 fail)
         and The_Skin_Never_Prices_The_Floor -> "WL-999"
  GREEN  "project_rate_card.csv MAT_CODE match" @ 612,500 UGX/m2

the provenance assertion itself
  RED    Pass 3 keeps its rate, reports the category sentence     (1 of 7 fails)
  GREEN  "project_rate_card.csv MAT_CODE match"
```

| | W2 alone | with W4 |
|---|---|---|
| `dotnet build` | 0 err / 0 warn | 0 err / 0 warn |
| `StingTools.Boq.Tests` | 1,324 | **1,331** |
| `StingTools.Tags.Tests` | 919 | 919 |

**Still not verified in Revit.** The chain is proven over the Revit-free halves the plugin actually calls; the Revit-bound link between them — `ElementMatCodeReader` reading a compound structure and a material parameter — remains unexercised against a document.

PR #967, opened 2026-09-17
Make the 12 shared-parameter conflicts diagnosable, and gate the class

StingD85:
Work from the FUNC/PROD binding-strategy session. It started as "why do the FUNC/PROD codes not
write?" and ended in the shared-parameter type conflicts that stop `STING_Tag_Universal.rfa`
loading at all.

## What is in it

**The diagnosis, made permanent**
- `tools/check_shared_param_types.py` — every kit/trimmed shared-parameter file must agree with
  `MR_PARAMETERS.txt` on GUID *and* type. Two of the twelve conflicts came from
  `docs/UNIVERSAL_TAG_MASTER_PARAMS.txt` typing a NUMBER and a YESNO as TEXT under the same GUIDs,
  and nothing detected it until a human hit the load dialog. Passes on all 75 kit parameters, and
  baselines the 332 pre-existing type disagreements in `FAMILY_PARAMETER_BINDINGS.csv` so they
  cannot grow (ROADMAP PARAMTYPE-3).
- `docs/UNIVERSAL_TAG_CONFLICT_RESOLUTION_RUNBOOK.md` — the ordered fix, label work before the
  deletions, because deleting a parameter a row still references strips the row.

**Token writes stop lying**
- A failed token write is no longer reported as a missing code, and the category is named alongside
  the parameter (`Core/TagTokenIntegrity.cs`, `Core/TokenBindingCoverage.cs`,
  `Tags/PreTagAuditCommand.cs`), so "the binding question is answered before tagging, not after".
- Three new Revit-free suites in `StingTools.Tags.Tests` cover the integrity, coverage and
  failure-report paths.

**Kit corrections**
- Row 1 of the universal label is `ASS_DISPLAY_TXT`, not `ASS_TAG_1_TXT`; rows 22 and 50 repointed
  onto their `_TXT` display mirrors; in-tag status badges marked abandoned (a visibility formula
  cannot read the tagged element's parameters — status is delivered by Status Register).

**Protocol**
- `docs/TAG_TEST_PROTOCOL.md` brought up to date, with one test corrected that had begun to
  false-pass, and `docs/UNIVERSAL_TAG_DUCT_SMOKE_TEST_RUN.md` added as the executable gate before
  propagating to 206 families.
- `docs/TAG_FAMILY_LIBRARY_STRATEGY.md` — the reuse mechanism already exists; this writes down what
  it is.

## Verification

`tools/check_shared_param_types.py` passes. The three new test suites are Revit-free and run in
`StingTools.Tags.Tests`. The `.rfa` work in the runbook is by hand in Revit and is not done by this
PR — the runbook is the instruction, not the result.


StingD85:
One correction to `tools/check_shared_param_types.py`'s header, measured while chasing the failed
duct propagation (#968):

> FAMILY_PARAMETER_BINDINGS.csv is read by BatchAddFamilyParamsCommand … so a wrong DataType here
> becomes a real family parameter, and a load conflict, on any set-up project.

`BatchAddFamilyParamsCommand` opens `MR_PARAMETERS.txt` and binds the *definition* it finds there
(`Temp/TemplateManagerCommands.cs`, `spfPath = FindDataFile("MR_PARAMETERS.txt")`). The CSV supplies
which parameter and which category; its `DataType` column is never used to type anything. So the 332
disagreements are a documentation hazard, not a load hazard, on that path.

The baseline is still worth keeping — a data file that disagrees with MR misleads every reader of it,
and the gate stops it growing — but the stated reason is stronger than the evidence. Worth softening
the comment before merge; the assertion itself needs no change.

Same file, same spirit: `TAG_PARAM_ALIGNMENT_AUDIT.csv` lists all twelve as `TEXT` and has **no
consumer anywhere in the C# tree**. It reads like the answer and is inert.

